PR #22, opened 2017-03-08
Format css fonts

chazzmoney:
PR for adding the css/fonts format.  Some bug fixes/cleanup in here too; the main changes are  in bfc4398.

Also, I think there is one problem - I hard coded the path to the font to include a ../ from the current directory (essentially assuming that the user has a css folder at the root level).  Maybe this should be an argument passed in via config.files (as per issue 20)?

### DIFF
--- a/example/complete/properties/font.json
+++ b/example/complete/properties/font.json
@@ -2,10 +2,10 @@
   "font": {
     "family": {
       "base": {
-        "value": "Open Sans"
+        "value": "{asset.font.opensans.name.value}"
       },
       "icons": {
-        "value": "MaterialIcons"
+        "value": "{asset.font.icon.name.value}"
       }
     }
   }
--- a/example/complete/properties/font.json
+++ b/example/complete/properties/font.json
@@ -2,7 +2,7 @@
   "font": {
     "family": {
       "base": {
-        "value": "Amazon Ember"
+        "value": "Open Sans"
       }
     }
   }
--- a/example/complete/properties/font.json
+++ b/example/complete/properties/font.json
@@ -3,6 +3,9 @@
     "family": {
       "base": {
         "value": "Open Sans"
+      },
+      "icons": {
+        "value": "MaterialIcons"
       }
     }
   }
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -34,89 +34,9 @@ function cssHeader() {
 }
 
 module.exports = {
-  'css/fonts': function(dictionary) {
-    var fontFaces = {};
-    _.each(dictionary.allProperties, function (prop) {
-      if (prop.attributes.category === 'asset' && prop.attributes.type === 'font') {
-        var parent = prop.path.slice(0);
-        var child = parent.pop();
-        var attrStr = JSON.stringify(parent);
-        if(!fontFaces[attrStr]) {
-          fontFaces[attrStr] = {};
-        }
-        fontFaces[attrStr][child] = prop.value;
-      }
-    });
-
-    var genSrc = function (properties) {
-      var tabs = '  ';
-      var srcIndent = '     ';
-      var asset_path = '../';
-
-      var name = '', weight = '', style = '';
-      var prefix = '@font-face {\n' + tabs + "font-family: '{{NAME}}';\n";
-      var postfix = '}\n';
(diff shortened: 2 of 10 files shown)


chazzmoney:
Once this is in, I think we can include a font change as part of our example (actually, I don't know if changing the font.family.base.value has an impact on the ios/android in example/complete...)

example/complete/properties/font.json
```diff
@@ -2,7 +2,10 @@
   "font": {
     "family": {
       "base": {
-        "value": "Amazon Ember"
+        "value": "Open Sans"
+      },
+      "icons": {
+        "value": "MaterialIcons"
```
dbanksdesign: Change this to {asset.font.icon.name.value}

lib/common/formats.js
```diff
@@ -34,29 +34,106 @@ function cssHeader() {
 }
 
 module.exports = {
-  'css/variables': function(dictionary) {
-    var to_ret;
-    return cssHeader() + ':root {\n' + _.map(dictionary.allProperties, function (prop) {
-      if (prop.attributes.category !== 'asset') {
-        to_ret = '  --' + prop.name + ': ' + prop.value + ';';
+  'css/fonts': function(dictionary) {
```
dbanksdesign: This method is really long, I think functionality like this is better suited for a template. We can achieve the same functionality with this template:

```
<% _.forIn(properties.asset.font, function(font) { %>
  @font-face {
    font-family: <%= font.name.value %>;<% if (font.eot) { %>
    src: url('../<%= font.eot.value %>');
    src: url('../<%= font.eot.value %>?#iefix') format('embedded-opentype'),<% } if(font.ttf) { %>
      url('../<%= font.ttf.value %>') format('truetype'),<% } if(font.otf) { %>
      url('../<%= font.otf.value %>') format('otf'),<% } if(font.woff) { %>
      url('../<%= font.woff.value %>') format('woff'),<% } if(font.woff2) { %>
      url('../<%= font.woff2.value %>') format('woff2');<% } %>
  }
<% }); %>
```

Or maybe use this line `_.forIn(properties.asset.font, function(font) {` instead of looping over properties and creating these objects on the fly, and then follow the same type of logic in the template above to construct the string.

lib/common/templates/static-style-guide/index.html.template
```diff
@@ -61,7 +62,7 @@
         <% if(property.attributes && JSON.stringify(property.attributes)!=="{}") { %>
           <div class="style-guide-property-attributes-control"></div>
         <% } %>
-        <div class="style-guide-property-value"><%= property.value %></div>
+        <div class="style-guide-property-value"><%= property.value %><% if(property.path.indexOf('icon')!=-1 && property.path.indexOf('content')!=-1) { %><i class="icon-font"><%= property.value %></i><% } %></div>
```
dbanksdesign: This test might be more readable: `property.attributes.category === 'content' && property.attributes.type === 'icon'

lib/common/formats.js
```diff
@@ -34,29 +34,106 @@ function cssHeader() {
...
+      return to_ret_prop;
     }).filter(function (strVal) { return !!strVal; }).join('\n') + '\n}\n';
+
+    return cssHeader() + root_vars;
   },
 
   'scss/variables': function(dictionary) {
     var to_ret;
     return sassHeader() + _.map(dictionary.allProperties, function (prop) {
-      if (prop.attributes.category !== 'asset') {
```
dbanksdesign: Removing this check produces invalid css/scss

lib/common/templates/static-style-guide/index.html.template
chazzmoney: As mentioned, I think this forces people into using the CTI structure that we've defined.  I'll change it, and at some point we should have the philosophical conversation about property metadata vs organizational structure vs use structure and how people will end up using this system.

lib/common/formats.js
chazzmoney: Really we should check the property values and if they don't match <declaration-value> syntax, then we should escape and quote it.  That said... for now I'll just quote things with the category of asset.

Ref: https://drafts.csswg.org/css-syntax-3/#typedef-declaration-value

PR #32, opened 2017-04-07
Improved tests

cpenarrieta:
This PR covers the following:
- I found a bug on some tests while using `assert.throws`. The first parameter of `.bind` is the `this` value and the rest are the arguments of the function. The tests were still passing because the arguments where undefined thus causing an exception (but not really the exception we wanted).
- all the `assert.throws` tests now check for the specific error message
- fixed `package.json` to include all test in the coverage and lint
- fixed a bug in `transformGroup.js`. The `this` reference inside the `forEach ` was wrong.

### DIFF
--- a/lib/register/transformGroup.js
+++ b/lib/register/transformGroup.js
@@ -27,10 +27,10 @@ function registerTransformGroup(options) {
   if (!_.isArray(options.transforms))
     throw new Error('transforms must be an array of registered value transforms');
 
-  options.transforms.forEach(function(t) {
+  options.transforms.forEach((function(t) {
     if (_.has(this.transforms, t))
       throw new Error('transforms must be an array of registered value transforms');
-  });
+  }).bind(this));
 
   this.transformGroup[options.name] = options.transforms;
   return this;
--- a/lib/register/transformGroup.js
+++ b/lib/register/transformGroup.js
@@ -28,7 +28,7 @@ function registerTransformGroup(options) {
     throw new Error('transforms must be an array of registered value transforms');
 
   options.transforms.forEach((function(t) {
-    if (_.has(this.transforms, t))
+    if (!_.has(this.transform, t))
       throw new Error('transforms must be an array of registered value transforms');
   }).bind(this));
 
--- a/test/registerTransformGroup.js
+++ b/test/registerTransformGroup.js
@@ -18,47 +18,96 @@ var assert = require('chai').assert,
 describe('registerTransformGroup', function() {
   it('should error if name is not a string', function() {
     assert.throws(
-      StyleDictionary.registerTransformGroup.bind({transforms: ['foo']})
+      StyleDictionary.registerTransformGroup.bind(null, {
+        transforms: ['foo']
+      }),
+      Error,
+      'transform name must be a string'
     );
 
     assert.throws(
-      StyleDictionary.registerTransformGroup.bind({name: 1, transforms: ['foo']})
+      StyleDictionary.registerTransformGroup.bind(null, {
+        name: 1,
+        transforms: ['foo']
+      }),
+      Error,
+      'transform name must be a string'
     );
 
     assert.throws(
-      StyleDictionary.registerTransformGroup.bind({name: [], transforms: ['foo']})
+      StyleDictionary.registerTransformGroup.bind(null, {
+        name: [],
+        transforms: ['foo']
+      }),
+      Error,
+      'transform name must be a string'
     );
 
     assert.throws(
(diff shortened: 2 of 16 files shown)


chazzmoney:
Looks pretty good. Danny built the tests so I'll let him see if anything looks out of place, but I'm a thumbs up on this otherwise.

Quick question:  in 'style-dictionary/lib/utils/resolveObject.js' you added line 132. (current_context = [];)  Since it throws there, it looks like it will error out all the way up the chain.  Is the intent here to prevent future issues if said error was caught?  Or am I missing something in how this piece is executing?

Seems like it might be better placed before line 42?  Danny?

cpenarrieta:
@chazzmoney the reason I added `current_context = [];` is to avoid errors in test cases.
In `test\utils\resolveObject.js` once `circular.json` is tested `current_context` have the value `['a']`. Then when `circular_2.json` is executed `current_context` = `['a', 'a', 'b', 'b']` (the first `a` is from the previous execution). This caused a `maximum call stack exceeded` error.
I'm not happy with that solution, I'm open to suggestions.

chazzmoney:
I'm with you.  So what do you think of removing that line from it current position and inserting it immediately before line 42?

cpenarrieta:
@chazzmoney I agree. I'll fix it

chazzmoney:
Looks good to me.  I'm sure Danny will have some good comments.  He might be out this week (we work in different groups so I'm not 100% sure on his schedule).  He usually gets to things within a few days, so I'm sure he will get you a response soon.  If he doesn't get one done by Friday I'll just do a merge and we can fix anything he finds after.

cpenarrieta:
np. I'll wait for the code review. Thanks.

chazzmoney:
No worries.  I know that Danny would be onboard with a merge as is and then fix anything else later.  I also think he'll probably get you a review soon and we won't need to worry about it.

Some context for that thinking:

It isn't like we write perfect code.  Right now, it's just us trying to make something useful for everyone.  That means more regular updates and less of a focus on having strong releases.  We still want to be stable... and this is the building out of the project as opposed to a fully functional product.  Lots of small changes are much better than a single larger "perfect" change.

Also, since I failed to say it earlier, thanks for your contribution!

lib/register/transformGroup.js
```diff
@@ -27,10 +27,10 @@ function registerTransformGroup(options) {
   if (!_.isArray(options.transforms))
     throw new Error('transforms must be an array of registered value transforms');
 
-  options.transforms.forEach(function(t) {
+  options.transforms.forEach((function(t) {
     if (_.has(this.transforms, t))
```
dbanksdesign: Not your fault, but looking at this I discovered this code is actually broken. this.transforms is undefined, the real thing we need is this.transform. And the conditional is backwards, we want to make sure that each transform matches a transform already defined. If we revert this code you added, this error will never throw so it was never catching what it was supposed to!

So lets change this line to
`if (!_.has(this.transform, t))`

We will need to update line 117 in test/registerTransformGroup.js to use an actual prebuilt transform like 'color/rgb' or 'name/cti' as this test will now fail because 'foo' is not a registered transform.

test/registerTransformGroup.js
```diff
@@ -18,47 +18,96 @@ var assert = require('chai').assert,
...
+      }),
+      Error,
+      'transforms must be an array of registered value transforms'
     );
   });
 
   it('should error if transforms arent registered', function() {
     assert.throws(
-      StyleDictionary.registerTransformGroup.bind({name: 'foo', transforms: ['bar']})
+      StyleDictionary.registerTransformGroup.bind({ transforms: { bar: 'cat' } },
```
dbanksdesign: Lets change the `.bind({ transforms: { bar: 'cat' } }` to `.bind(StyleDictionary)` because the code was falsely testing the wrong property ('transforms' and not 'transform').

cpenarrieta:
@dbanksdesign Thanks for the review, now it makes more sense. I updated the code please take a look.

cpenarrieta:
@chazzmoney  @dbanksdesign unrelated to this project. 
Do you know who should I contact that can check this PR https://github.com/amzn/amzn.github.io/pull/1 ?

chazzmoney:
Unfortunately there is a lot of bureaucracy in relationship to open source stuff at Amazon.  We are trying to cut through it some with our project.  I'm not sure who could look at / approve that, but I'll try to see if we can find someone.  Thanks for all your support!

dbanksdesign:
Merged!

PR #38, opened 2017-06-06
Adding clean action

chazzmoney:
This is the pull request too add the clean action system to style-dictionary.

I'm not doing comments because we handled it all in the issue.

### DIFF
--- a/bin/style-dictionary
+++ b/bin/style-dictionary
@@ -27,6 +27,13 @@ program
   .action(styleDictionaryBuild);
 
 program
+  .command('clean')
+  .description('Removes files specificed in the config of the style dictionary package of the current directory.')
+  .option('-c, --config <path>', 'set config path. defaults to ./config.json')
+  .option('-p, --platform [platform]', 'only clean specific platform(s). Must be defined in the config', collect, [])
+  .action(styleDictionaryClean);
+
+program
   .command('init <type>')
   .description('Generates a starter style dictionary')
   .action(function(type){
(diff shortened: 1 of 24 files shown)


chazzmoney:
Actually, now I do want comments.  I had to modify the order of things, was getting an ENOENT because clean actions was deleting a directory and the clean files was trying to delete a file in that directory (which no longer existed).

I modified the order of things to be more correct (undo files, undo actions, undo dirs).

However, I also think that maybe we should not throw an error for trying to delete a missing file?  What do you think?

dbanksdesign:
Can you be more specific about the clean action trying to delete a file in a folder that no longer existed? The clean actions are defined by the actions themselves, so was it a problem with a specific clean action? If so, we should fix that action. Having the cleanDirs stuff seems a bit overkill.

chazzmoney:
Sure.  Lets look at example/complete.  Specifically, the 'android-asset' platform.  And pretend we are running the old code (before cleandirs).

Two things happen on build and need to be undone:
-copy_assets
-build of 'assets/data/properties.json'

If I run the undo on copy assets, it deletes the folder.  Then it tries to delete assets/data/properties.json, which no longer exists because its containing folder was deleted.  Error.

If the file delete process happens first, then assets/data/properties.json is deleted.  The directory is not because the other assets files were copied in.  The copy_assets then deletes the directory it copied everything to.  This is correct, but only in the case where there isn't a mutli-level directory structure being created by the file builder.  It is correct on accident.  In other words, for it to be correct, the person building the config.json would have to know that they were creating (or avoiding creating) a potential error case.

dbanksdesign:
Hmm I see what you are saying. I think that case is particularly weird because looking at it, the copy_assets action should wipe out that directory because the files are generated first then actions happen.

chazzmoney:
To clarify, the cleanFile is trying to delete a file that is not there since the copy_assets undo removes the parent directory for the file.  Thus I put file clean, then action clean, then dir clean, in that order.

PR #47, opened 2017-08-15
add commonJS format

akdetrick:
Consumers may expect a javascript distribution to be pre-transpiled. This format enables generation of a distribution with CommonJS module exports.

### DIFF
--- a/docs/default_formats.md
+++ b/docs/default_formats.md
@@ -41,6 +41,11 @@
       <td></td>
     </tr>
     <tr>
+      <td>javascript/commonJS</td>
+      <td></td>
+      <td></td>
+    </tr>
+    <tr>
       <td>json</td>
       <td></td>
       <td></td>
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -95,6 +95,14 @@ module.exports = {
       }).join('\n');
   },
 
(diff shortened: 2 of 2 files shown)


akdetrick:
I should mention that I like the idea of limiting the built-in formats and transforms given the flexibility for custom formats. The javascript landscape however, might make this a special case.

Feel free to close this PR if this new format doesn't belong.

dbanksdesign:
Sorry for the late reply, looks good! If you could add some quick unit tests, like this: https://github.com/amzn/style-dictionary/blob/master/test/formats/javascriptModule.js it should be good to go!

dbanksdesign:
Ping @akdetrick

chazzmoney:
Ping @akdetrick

akdetrick:
👀 I'll take another look at this to add tests and resolve conflicts

akdetrick:
Looks like the `javascript/module` format added recently covers this use case in a more portable way; closing this PR.

danlevy1:
I'm actually looking for the solution that was created in this PR. Essentially, I'm looking for the `javascript/es6` format, but using CommonJS instead of ES Modules. Does anyone see a problem with this proposed solution? If not, I'll just open a new PR and add unit tests.

PR #48, opened 2017-12-14
Collision detection

### LINKED ISSUE #39: Throw an error or warning when merge collisions happen
According to tenet 7, "Fail loudly", when property collisions happen we should error out. We don't want downstream inconsistencies to happen because someone mistakenly overrode a property.

dbanksdesign:
- Fix #39 
- Alert the user when a property collision happens (a property is
defined twice in same package, does not include imports)
- Usage: set 'log' to 'warn' or 'error' in the config of a style
dictionary
- If set to 'warn' will just output to the console, but not throw
- If set to 'error' will throw

### DIFF
--- a/lib/extend.js
+++ b/lib/extend.js
@@ -34,7 +34,7 @@ function extend(opts) {
 
   // Creating a new object and copying over the options
   // Also keeping an options object just in case
-  to_ret = deepExtend({options: options}, this, options);
+  to_ret = deepExtend([{options: options}, this, options]);
 
   // Update properties with includes from dependencies
   if (options.include) {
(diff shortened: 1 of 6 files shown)


lib/extend.js
```diff
@@ -54,8 +54,18 @@ function extend(opts) {
     if (!_.isArray(options.source))
       throw new Error('source must be an array');
 
-    var props = combineJSON(options.source, {deep: true});
-    to_ret.properties = deepExtend({}, to_ret.properties, props);
+    var props = combineJSON(options.source, true, function Collision(opts) {
+      if (options.log) {
+        var str = `Collision detected at: ${opts.path.join('.')}! Original value: ${opts.target[opts.key]}, New value: ${opts.copy[opts.key]}`;
+        if (options.log === 'warn') {
```
NicestRudeGuy, line 60: I couldnt find the log option inside the doc. I was gonna write JS script but got saved.

EiffelFly: @NicestRudeGuy Hi, How to you config this? Could you share a example? Thanks

EiffelFly: I think the type declaration about this option is wrong

NicestRudeGuy: @EiffelFly  if you mean to stop the StyleDictinary from executing when there is collision you can just add 
`"log": "error"` to its config file

PR #61, opened 2018-02-22
Fixes #56

### LINKED ISSUE #56: Add support for non-string values
Currently, when style dictionary tries to transform a value it does this:
```js
if (transform.type === 'value' && property.value.indexOf('{') < 0)
```
[lib/transform/property.js](https://github.com/amzn/style-dictionary/tree/master/lib/transform/property.js)

However if you put a number as a value, which should make sense for unit-less numbers like line-height or milliseconds. I think we just need to add a check in there to make sure `property.value` is a string before testing for '{'

dbanksdesign:
Adds the ability to have non string, or more specifically, types that don't have a `indexOf` method, as values. The type of the value will be retained when referencing it in the style dictionary.

Also updated the logic for when to run transforms and what to include in `allProperties` to not include any leaf node that does not have a `value` attribute. This will now allow you to have any arbitrary data/content in the style dictionary JSON files.


chazzmoney:
Did you mean to commit the package-lock and scapp_icons-regular.ttf files?  Otherwise it looks really good.

dbanksdesign:
The package-lock file, probably not? I mean no dependencies are changing with this commit, should I just remove revert it?

For the ttf file, yes, that probably should not have been in there at all.

dbanksdesign:
For posterity, the reason why the package-lock.json got updated even though no dependencies changed is I installed a new node version (9.6) and when doing npm i, pulled new dependency versions I guess because of the new node and npm versions?

Issue #66, opened 2018-03-04
Transforms aren't performed on non-string values


dbanksdesign:
The problem is with this line of logic:
https://github.com/amzn/style-dictionary/blob/master/lib/transform/property.js#L35

If the value is not a string, transforms will never run. The unit tests were not testing if transforms ran on non-string values, only that references to them were properly copied over.

chazzmoney:
Closing after merged PR.

PR #67, opened 2018-03-04
Fix for #66

dbanksdesign:
Fix for #66 Transforms are not being performed on non-string values. Added unit tests to make sure transforms run properly.


chazzmoney:
Looks good to me.

Issue #54, opened 2018-02-09
Add format for less variables


dbanksdesign:
Pretty straightforward, like 'scss/variables' or 'css/variables' formats. Make sure to add unit tests.

davixyz:
I might be able to tackle this. Let me give it a go

PR #76, opened 2018-03-24
Add LESS format and create tests for less

davixyz:
*Issue #, if available:*
https://github.com/amzn/style-dictionary/issues/54

*Description of changes:*
- Add LESS format based on SCSS output
- Use of the same function for file headers since all the format languages support multi-line comments
- Updated docs to reflect new changes
- Created tests for the format and also updated general tests to add new format

*Notes:*
Should I add myself as contributor if this is merged?

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -14,25 +14,38 @@
 var _ = require('lodash');
 var module_def = 'module.exports = ';
 
-function jsHeader() {
+function fileHeader() {
   return '/**\n' +
          ' * Do not edit directly\n' +
          ' * Generated on ' + new Date() + '\n' +
          ' */\n\n';
 }
 
-function sassHeader() {
-  return '//\n' +
-         '// Do not edit directly\n' +
-         '// Generated on ' + new Date() + '\n' +
-         '//\n\n'
+function variablesWithPrefix(prefix, properties) {
+  return _.map(properties, function(prop) {
+      var to_ret_prop = prefix + prop.name + ': ' + (prop.attributes.category==='asset' ? '"'+prop.value+'"' : prop.value) + ';';
+
+      if (prop.attributes.comment)
+        to_ret_prop = to_ret_prop.concat(' // ' + prop.attributes.comment);
+      return to_ret_prop;
+    })
+    .filter(function(strVal) { return !!strVal })
+    .join('\n');
 }
 
-function cssHeader() {
-  return '/* Do not edit directly */\n' +
-         '/* Generated on ' + new Date() + ' */\n\n';
+function iconsWithPrefix(prefix, properties, config) {
+  return fileHeader() + _.chain(properties)
+    .filter(function(prop) {
+      return prop.attributes.category === 'content' && prop.attributes.type === 'icon';
+    })
+    .map(function(prop) {
+      var varName = prefix + prop.name + ': ' + prop.value + ';';
+      var className = '.' + config.prefix + '-icon.' + prop.attributes.item + ':before ';
+      var declaration = '{ content: ' + prefix + prop.name + '; }';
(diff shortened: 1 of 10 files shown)


lib/common/formats.js
```diff
@@ -110,6 +97,53 @@ module.exports = {
...
+   * @example
+   * ```less
+   * @color-background-base: #f0f0f0;
+   * @color-background-alt: #eeeeee;
+   * ```
+   */
+  'less/variables': function(dictionary) {
+    var to_ret;
+    return fileHeader() + _.map(dictionary.allProperties, function (prop) {
+      to_ret = '@' + prop.name + ': ' + (prop.attributes.category==='asset' ? '"'+prop.value+'"' : prop.value) + ';';
```
davixyz: I noticed there's a little repetition here and with the sass (similar to CSS as well); We could create a method that receives all these props and the prefix and should automatically create the array with the correct value and use the same method across these transforms; Thoughts?

lib/common/formats.js
```diff
@@ -110,6 +97,53 @@ module.exports = {
...
+   * .icon.email:before { content:@content-icon-email; }
+   * ```
+   */
+  'less/icons': function(dictionary, config) {
+    return fileHeader() + _.chain(dictionary.allProperties)
+      .filter(function(prop) {
+        return prop.attributes.category === 'content' && prop.attributes.type === 'icon';
+      })
+      .map(function(prop) {
+        var varName = '@' + prop.name + ': ' + prop.value + ';';
```
davixyz: we could probably do the same for this function since it's shared between SCSS & LESS

test/formats/lessIcons.js
```diff
@@ -0,0 +1,63 @@
...
+  describe('less/variables', function() {
+    var variableResult = '@' + propertyName;
+    it('should have a valid less variable', function() {
+      var expectedResult = variableResult + ': ' + propertyValue + ';';
+      // Check for substring of formatter function
+      assert.include(formatter(dictionary, config), expectedResult, 'string contains substring');
+    });
+
+    it('should have a valid class created', function() {
+      var expectedResult = '.' + config.prefix + '-icon.' + itemClass + ':before { content: ' + variableResult + '; }';
```
davixyz: kinda didn't like this approach to test which is based on substrings. Any good way to test this? I didn't find any examples for scss variables testing

dbanksdesign: Agreed, this testing is a bit brittle. I'm open to ideas for how to improve this. We could have the less/sass compilers as dev dependencies and try compiling the output file?

lib/common/formats.js
davixyz: Yeah, I'll update it

test/formats/lessIcons.js
davixyz: sounds like a good idea; Let me give it a shot

dbanksdesign:
Can you undo the package-lock.json changes? It seems there is some weirdness with some versions of npm where it will update the package-lock.json file when doing npm install. I need to go through and check the dependencies and update them if necessary. I'd like to do that as a separate commit and to clean up the package/yarn files.

davixyz:
Changes are done, added less as a dependency; I can add SASS dependency and tests on another PR.

dbanksdesign:
Thanks for putting that together so quickly! It looks good to me, just doing an internal review and hopefully this can get merged in today or tomorrow. Thanks!

dbanksdesign:
Sorry for the delay on this, still waiting for an internal review.

davixyz:
@dbanksdesign No worries, totally understand. Meanwhile I can make another PR for the SCSS Testing

dbanksdesign:
Approved and merged in. Thanks Carlos!

PR #82, opened 2018-04-17
Adding more testing infrastructure

davixyz:
*Issue #, if available:*

*Description of changes:*
- Fixed most of the ESlint warnings, left the ones in the ES6 export testing file because I also couldn't figure out a way to test ES6 in this normal JS.

- Added Testing for SCSS by adding the `node-sass` library and testing against correctness of syntax (valid output)

- Added Testing of normal Javascript object using the `vm` node API with a try/catch block. This can also be achieved with an eval, but the run in context doesn't pollute the global scope I think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "istanbul": "^0.4.5",
     "jsdoc-to-markdown": "^3.0.3",
     "less": "^3.0.1",
-    "mocha": "^3.1.2"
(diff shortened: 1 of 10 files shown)


dbanksdesign:
Testing the ES6 export might be difficult... Other than that this looks awesome. Doing an internal review now.

Issue #81, opened 2018-04-17
Adding husky to test in git commit


davixyz:
While adding some more testing I noticed that we don't run the tests on commit. It would good to add a pre-commit hook to run the tests so engineers don't have to rely on the CI to catch any integration issues. This also goes well with the 3rd point in the contribution guidelines
https://github.com/amzn/style-dictionary/blob/master/CONTRIBUTING.md#code-rules

I can quickly add this on a PR:
https://github.com/typicode/husky

dbanksdesign:
Love it!

dbanksdesign:
Code merged, closing issue

PR #83, opened 2018-04-17
Add husky pre-commit hook

davixyz:
*Issue #, if available:*
https://github.com/amzn/style-dictionary/issues/81

*Description of changes:*
Adding husky to test pre-commit.
Tried to add the latest husky version but it seems that is incomatible with node V4 and V5... So I guess we'll have to settle with using V0.14.3. Tradeoff is that if you had githooks in the project before sometimes they are imcompatible, so you need to do
`rm -rf .git/hooks/`.

You can see the issues I had here:
https://travis-ci.org/amzn/style-dictionary/builds/367857409

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "version": "node ./scripts/version.js && npm run generate-docs",
     "generate-docs": "node ./scripts/generateDocs.js",
     "serve-docs": "docsify serve docs -p 3000 -P 12345",
-    "install-cli": "npm install -g $(npm pack)"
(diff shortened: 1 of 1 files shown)


davixyz:
@dbanksdesign any thoughts on this?

dbanksdesign:
Sorry, it has been a busy week. We could also just bump up the node version of style dictionary to >= 6, but I don't know if that would be prohibitive to people. This looks good to me though, I'll send it to internal review Monday morning! Thanks again for all your help!

Issue #19, opened 2017-03-03
YML, JSON5, and HJSON support


dbanksdesign:
These formats are a bit better suited than raw JSON because they have nicer formatting and allow comments.

dbanksdesign:
To start, document the ways people can use YML, JSON5, and HJSON already.

dbanksdesign:
Made a yaml example: https://github.com/amzn/style-dictionary/tree/yaml/example/yaml

custa1200:
I also would love to see JSON5 support

chazzmoney:
JSON5 support was added in #165.

HJSON and YML are not yet supported

PR #84, opened 2018-04-25
Adding Yaml Example

dbanksdesign:
*Issue #, if available:* #19 

*Description of changes:* Adding an example showing how to use Yaml instead of JSON. 


chazzmoney:
Hmmm.  This leads me into a question whether supporting a different properties format should be more pluggable?  We added in #89 as supported functionality instead of an example. Do we need consistency for properties?

PR #88, opened 2018-05-11
Update docs to include the word "alias"

kylegach:
*Issue #, if available:*
We recently embarked on extracting design tokens from our component library, and compared style-dictionary vs. theo to do so. We loved what we saw with this project, but ultimately chose theo because we (mistakenly) believed there was no concept of referencing another property's value within a property (what theo calls an "alias").

In hindsight, I see now that your docs make that functionality quite clear, but I think literally using the word "alias", so that it shows up in a quick Find command, would be helpful.

*Description of changes:*
Simple reword of the repo README and the README used as the overview in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ This tells the style dictionary build system how and what to build. The default 
 }
 ```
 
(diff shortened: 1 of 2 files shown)


dbanksdesign:
Thanks for this!

Issue #85, opened 2018-04-25
Add Custom Parser Support


dbanksdesign:
### Acceptance Criteria
User should be able to supply a custom parser to the style dictionary to be used when reading property files. It should be done in a generic way so that any parser (yaml, hjson, and any other) could conceivably be used. A parser should be a function that takes a file (as a path string) as input, and outputs a JS object. Any function that follows this could be used as a parser. 

### How does this benefit the user?
Some do not like writing JSON, it can be verbose and not as simple as something like yaml. This will allow users to use any data formatting syntax to describe their style dictionaries. This fits in with the spirit of style dictionary in being flexible to accommodate any project/platform/framework. At its core, a style dictionary is a data structure for defining styles; the filetype the data structure is written in is irrelevant.

jreichenberg:
Easy half-step: allow a property file to be a node modules that exports a js obj... `utils/combineJSON` could simply replace `JSON.parse()` with `require(files[i])`. If it's actually a json file require will parse it for you. Not the complete solution but non-invasive...

elliotdickison:
I might give this a shot. My initial take is that a `registerParser` API would be consistent with the other extension APIs and would allow the user to specify multiple parser functions. The `registerParser` API would take both a parser function and the file extensions to which it applies. This would enable you to house properties in multiple file types. Is that over-engineering it? A couple use-cases I can think of:
* users who want to keep most properties in static files (e.g. yaml), but need to have a few properties in dynamic files (e.g. JS) - or vice versa
* users who are doing a slow migration from one file type to another (this might dovetail with some of the plans for an import system)

If we go forward with this I would extract the current JS/JSON parser into a parser function (basically just `path => require(path)`) and pre-apply it using the `registerParser` API (like what's being done with built-in transforms/formats). Thoughts?

chazzmoney:
We should decide if we want to do this, or if we want to do something more direct like #19. Either way, probably target for a 3.0 release?

dbanksdesign:
I agree about delaying parsing until we call exportPlatform (which is called internally from buildPlatform), but yea we should do that in the next major release to not break anything.

chazzmoney:
Agreed on the delaying of parsing. 👍🏽

chrisvxd:
Thought I'd share my custom build script to support YAML until it's officially supported: https://gist.github.com/chrisvxd/81137cb29b20a38c438c7918c776af4c

dbanksdesign:
[...]

PR #89, opened 2018-05-17
Allow property files to be node modules that export objects

jreichenberg:
*Issue #, if available:* https://github.com/amzn/style-dictionary/issues/85

*Description of changes:*
Allow property files to be node modules that export js objects, in addition to JSON files.

### DIFF
--- a/lib/utils/combineJSON.js
+++ b/lib/utils/combineJSON.js
@@ -11,16 +11,17 @@
  * and limitations under the License.
  */
 
-var fs = require('fs'),
-  glob = require('glob'),
+var glob = require('glob'),
   deepExtend = require('./deepExtend'),
-  extend = require('lodash/extend');
+  extend = require('lodash/extend'),
+  path = require('path'),
+  resolveCwd = require('resolve-cwd');
 
 /**
  * Takes an array of json files and merges
  * them together. Optionally does a deep extend.
  * @private
- * @param {String[]} arr - Array of paths to json files
+ * @param {String[]} arr - Array of paths to json (or node modules that export objects) files
  * @param {Boolean} [deep=false] - If it should perform a deep merge
  * @param {Function} collision - A function to be called when a name collision happens that isn't a normal deep merge of objects
  * @returns {Object}
@@ -35,12 +36,14 @@ function combineJSON(arr, deep, collision) {
   }
 
   for (i = 0; i < files.length; i++) {
-    var file_content = fs.readFileSync(files[i], 'utf8');
+    var resolvedPath = resolveCwd(path.isAbsolute(files[i]) ? files[i] : '.' + path.sep + files[i]);
+    var file_content;
 
     try {
-      file_content = JSON.parse(file_content);
+      file_content = require(resolvedPath);
     } catch (e) {
-      throw new SyntaxError(files[i] + ' - failed to parse JSON: ' + e.message);
+      e.message = 'Failed to load or parse JSON or JS Object: ' + e.message;
+      throw e;
     }
 
     if (deep) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,45 @@
       "integrity": "sha1-kbR5JYinc4wl813W9jdSovh3YTU=",
       "dev": true
     },
+    "acorn": {
+      "version": "5.5.3",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.5.3.tgz",
+      "integrity": "sha512-jd5MkIUlbbmb07nXH0DT3y7rDVtkzDi4XZOUVWAer8ajmF/DTSSbl5oNFyDOl/OXA33Bl79+ypHhl2pN20VeOQ==",
+      "dev": true
+    },
+    "acorn-jsx": {
+      "version": "3.0.1",
+      "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
+      "integrity": "sha1-r9+UiPsezvyDSPb7IvRk4ypYs2s=",
+      "dev": true,
+      "requires": {
+        "acorn": "3.3.0"
+      },
+      "dependencies": {
+        "acorn": {
+          "version": "3.3.0",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.3.0.tgz",
+          "integrity": "sha1-ReN/s56No/JbruP/U2niu18iAXo=",
+          "dev": true
+        }
+      }
+    },
+    "ajv": {
+      "version": "4.11.8",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.8.tgz",
+      "integrity": "sha1-gv+wKynmYq5TvcIK8VlHcGc5xTY=",
+      "dev": true,
+      "requires": {
+        "co": "4.6.0",
+        "json-stable-stringify": "1.0.1"
+      }
+    },
+    "ajv-keywords": {
+      "version": "1.5.1",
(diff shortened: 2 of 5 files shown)


jreichenberg:
... can now do things like this in a property file...

```
const myProps = {
  foo: 1
  bar: 2
}
myProps.sameAsFoo = myProps.foo;
myProps.somethingInYaml = loadYaml(...);
module.exports = myProps;
```

dbanksdesign:
Taking a look, but at first glance this looks pretty great! Thanks!

lib/utils/combineJSON.js
```diff
@@ -35,12 +36,13 @@ function combineJSON(arr, deep, collision) {
   }
 
   for (i = 0; i < files.length; i++) {
-    var file_content = fs.readFileSync(files[i], 'utf8');
+    var resolvedPath = resolveCwd(path.isAbsolute(files[i]) ? files[i] : '.' + path.sep + files[i]);
```
dbanksdesign: Is this module needed? Is there an issue with relative paths in `require()` on line 43?

jreichenberg: Yeah, the node resole alg. is basically (check me on this...)

> if `./some/path.js` or `../some/path.js` start at the module you're in
> if `some/path.js` resolve a node_module of name `some` and then the file path `path.js`

There's no part of it really that's "start at cwd() and give me file some/path relative to that".

jreichenberg: Could leave that up to the end user to make process.cwd() part of the glob pattern, too, I suppose.

dbanksdesign: Ah ok, I see now. It's a difference in how fs.readFile and require resolve the file paths, and because we are changing how we get the file contents from fs to require this is to ensue it is backwards compatible. Thank you for explanation and care to make sure this doesn't break backwards compatibility.

lib/utils/combineJSON.js
```diff
@@ -35,12 +36,13 @@ function combineJSON(arr, deep, collision) {
...
+    var file_content;
 
     try {
-      file_content = JSON.parse(file_content);
+      file_content = require(resolvedPath);
     } catch (e) {
-      throw new SyntaxError(files[i] + ' - failed to parse JSON: ' + e.message);
+      throw new SyntaxError(files[i] + ' - failed to load or parse JSON or JS Object: ' + e.message);
```
jreichenberg: I noticed one other thing here.. by wrapping the `e` in a SyntaxError it loses some of the details about the actual error. The message is there, but accurate file and line number go missing. Any objections to just throwing the actual `e` here and maybe just maybe prepending the 'failed to load or parse JSON or JS Object' boilerplate to the actual e's message? The difference:

Throw the `e`:

```
TypeError: Cannot read property 'dimension' of undefined
    at tokens (/Users/e828613/sc/manhattan/mds-tokens/src/tokens/decisions/btn/btn-primary.js:4:40)
    at Object.<anonymous> (/Users/e828613/sc/manhattan/mds-tokens/src/tokens/decisions/btn/btn-primary.js:20:4)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at combineJSON (/Users/e828613/sc/manhattan/mds-tokens/node_modules/style-dictionary/lib/utils/combineJSON.js:43:22)
```

Existing (with a different but similar underlying error) doesn't have the detail in the stack...

```
SyntaxError: src/tokens/decisions/btn/btn-primary.js - failed to load or parse JSON or JS Object: Cannot read property 'none' of undefined
    at combineJSON (/Users/e828613/sc/manhattan/mds-tokens/node_modules/style-dictionary/lib/utils/combineJSON.js:45:13)
    at Object.extend (/Users/e828613/sc/manhattan/mds-tokens/node_modules/style-dictionary/lib/extend.js:113:17)
    at Object.<anonymous> (/Users/e828613/sc/manhattan/mds-tokens/src/dictionary-config/style-dictionary.js:8:57)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:188:16)
```

lib/utils/combineJSON.js
```diff
@@ -35,12 +36,14 @@ function combineJSON(arr, deep, collision) {
...
+    var resolvedPath = resolveCwd(path.isAbsolute(files[i]) ? files[i] : '.' + path.sep + files[i]);
+    var file_content;
 
     try {
-      file_content = JSON.parse(file_content);
+      file_content = require(resolvedPath);
     } catch (e) {
-      throw new SyntaxError(files[i] + ' - failed to parse JSON: ' + e.message);
+      e.message = 'Failed to load or parse JSON or JS Object: ' + e.message;
+      throw e;
```
jreichenberg: See my outdated comment https://github.com/amzn/style-dictionary/pull/89#pullrequestreview-121910762 for an explanation. I think JSON error handling worked fine, but for all the kinds of errors that can be made in node modules the extra error detail is helpful.

lib/utils/combineJSON.js
dbanksdesign: Very good point, yes we should throw the e, but make sure the message for JSON files shows the filename. We redid the errors a few weeks ago to be more descriptive so it is easier to debug the issue, and adding the filename helps a lot (it previously just had a generic JSON error with no filename). So this fits perfectly in that theme of useful error messages.

lib/utils/combineJSON.js
jreichenberg: JSON error example with new... file name is in the existing e.message:

```
/Users/e828613/sc/manhattan/mds-tokens/node_modules/style-dictionary/lib/utils/combineJSON.js:46
      throw e;
      ^

SyntaxError: Failed to load or parse JSON or JS Object: /Users/e828613/sc/manhattan/mds-tokens/src/tokens/options/color.json: Unexpected token , in JSON at position 1816
    at JSON.parse (<anonymous>)
    at Object.Module._extensions..json (module.js:671:27)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at combineJSON (/Users/e828613/sc/manhattan/mds-tokens/node_modules/style-dictionary/lib/utils/combineJSON.js:43:22)
    at Object.extend (/Users/e828613/sc/manhattan/mds-tokens/node_modules/style-dictionary/lib/extend.js:113:17)
    at Object.<anonymous> (/Users/e828613/sc/manhattan/mds-tokens/src/dictionary-config/style-dictionary.js:1:117)
```

dbanksdesign: Perfect, thanks!

lib/utils/combineJSON.js
metalsheep: the traditional method is to use fs.readFileSync(__dirname + './my-file')).  The __dirname global is baked into node.js - https://nodejs.org/docs/latest/api/modules.html#modules_dirname

Dashue:
@jreichenberg @dbanksdesign Was using json files and accidentally came across this pr. 
Switched over to js files and wow what a difference in maintenance and development.

I thinkg the use-case with .js files deserves a better callout on the front page documentation. Would be invaluable to fellow developers like me coming along to use it.

dbanksdesign:
@Dashue definitely! The doc site definitely needs some love, we will keep this in mind for the next iteration.

Issue #86, opened 2018-04-30
Filter support across the board


davixyz:
# Summary

While doing some custom implementation where I need to filter very granularly certain properties for specific platform I noticed that the filtering system seems to be applied on some templates only.
In iOS Templates we have:
https://github.com/amzn/style-dictionary/blob/c2c31cb381be05d8324838361188879cd8685f3c/lib/common/templates/ios/colors.m.template#L23

Which basically takes whatever is coming from the configuration object as filter.

In Android, it's slightly different, color for instance:
https://github.com/amzn/style-dictionary/blob/c2c31cb381be05d8324838361188879cd8685f3c/lib/common/templates/android/colors.template#L17

It will only filter colors... but if we want to add a custom filter, we can't.

# Proposals

1) Standardize filters within templates/formats to use whatever is in config and extend a certain configuration.

##### Example in Android:
https://github.com/amzn/style-dictionary/blob/c2c31cb381be05d8324838361188879cd8685f3c/lib/common/templates/android/colors.template#L17

##### Changes
```js
var defaultFilter = {
  attributes: {
    category: "color"
  }
};
var props = _.filter(allProperties, Object.extend({}, this.filter, defaultFilter));
```

2) Add this filtering pre-processing to any transform; So we don't need to pre-filter at template/format level but the library does it automatically for everything.

dbanksdesign:
I like this. The filtering on templates/formats has been a little inconsistent, it would be great to standardize it. 

Could you go into proposal 2 a bit more?

davixyz:
I need to dig deeper on the library to understand the moving parts, but in general it would be that we need to add the filters before the format/template output starts to happen.

I can see that we have some opportunity to do it somewhere here:
https://github.com/amzn/style-dictionary/blob/98fcd5ce6ef125432af253f7f7d6e0392150a537/lib/buildFiles.js#L32

Since at that time we have reference to each file and thus their own filters

Or maybe before running the format?
https://github.com/amzn/style-dictionary/blob/98fcd5ce6ef125432af253f7f7d6e0392150a537/lib/buildFile.js#L43

But one problem that I see is that in formatters or templates we access differently the properties, some use `.allProperties`
https://github.com/amzn/style-dictionary/blob/98fcd5ce6ef125432af253f7f7d6e0392150a537/lib/common/formats.js#L68

And others `.properties`
[...]

PR #101, opened 2018-06-29
Filter API

elliotdickison:
*Issue #, if available:*

#86 

*Description of changes:*

New filter API

These changes introduce a new filter API that addresses #86. The key differences are:

* Filters are applied automatically in the `buildFile` function to both the `properties` and `allProperties` members of a dictionary. Filters are no longer applied ad-hoc in formatters.
* A filter can be a function in addition to an object. If a function is provided the property is passed wholesale to the function and the result (true or false) is used to determine whether the property should be included or filtered out. If an object is provided the behavior is the same as it used to be (a partial deep comparison is performed using lodash).

### DIFF
--- a/docs/package_structure.md
+++ b/docs/package_structure.md
@@ -38,6 +38,7 @@ The default way is to use a config.json file in the root of your package. Here i
       "transforms": ["attribute/cti", "name/cti/snake", "color/hex", "size/remToSp", "size/remToDp"],
       "buildPath": "build/android/src/main/res/values/",
       "files": [{
+        "filter": { "category": "color" },
         "destination": "style_dictionary_colors.xml",
         "template": "android/colors"
       }]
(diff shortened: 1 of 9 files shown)


chazzmoney:
Everything looks good.

In its current form, this is a breaking change to the signature of the buildFile function signature. We could move filter to the end of the buildFile parameter list to prevent this requiring a major version bump. If we agree to make a breaking change here, then we should consider if there are other changes we want to make to the signature.

Along those same lines, if we keep the breaking change here then we should decide if there are other architecture changes we want to make before release of a new major version.

Also, this brings up the function signature for cleanFile. It currently maintains signature parity with the buildFile function. It only actually uses the destination and platform parameters. If we are doing a major version change, we should decide if we want to continue to maintain the function signature parity or if we should reduce it down to only the parameters used.

elliotdickison:
I'll go ahead and move the new param to the end of the line. I do like your suggestions for breaking changes if a major version bump happens, but don't want to force that bump with this PR.

chazzmoney:
LGTM! 👍🏾

Since we just merged something a couple days ago maybe we will wait until next week to do the merge to continue building momentum?

Issue #108, opened 2018-07-25
"javascript/es6" format vs. transformGroup "web"


didoo:
Looking at the documentation (which is amazing, btw!) it's not clear to me the relation between the format "javascript/es6" and the transformGroup "web". 

I have created a simple example config:
```
{
  "source": ["properties/test.json"],
  "platforms": {
    "test": {
      "transformGroup": "web",
      "buildPath": "test/js/",
      "files": [{
        "destination": "tokens.es6.js",
        "format": "javascript/es6"
      }]
    }
  }
}
```

that reads from a very simple properties file:
```
{
  "background": {
    "base": { "value": "#ffffff" },
    "alt": { "value": "#fcfcfcfc" }
  }
}
```
The problem I have is that using the _"transformGroup": "web"_ the generated file is not a valid JS/ES6 file:
```
export const background-base = '#ffffff';
export const background-alt = '#fcfcfcfc';
```
The only way I can have it working is by replacing the _transformGroup_ declaration with a custom transformation:
```
"transforms": ["attribute/cti","name/cti/constant","size/px"]
```
which results in a valid JS/ES6 file:
```
export const BACKGROUND_BASE = '#ffffff';
export const BACKGROUND_ALT = '#fcfcfcfc';
```

Looking at [the example in the documentation](https://github.com/amzn/style-dictionary/blob/f6cfa273c844899118dfea709babd3bb0a431e2a/docs/formats.md#javascriptes6) I see no transformation or _transformGroup_ is declared, but the output is actually in PascalCase, but I can obtain it only applying the custom transforms:
```
"transforms": [ ... ,"name/cti/pascal", ... ]
```

So I am not sure if somehow I am missing something, or actually is an expected behaviour (in that case probably better to update the documentation and be more explicit).

dbanksdesign:
You aren't missing anything, I think this is just some mis-matched code. The "web" transform group applies the transforms: attribute/cti, name/cti/kebab, size/px, and color/css. As you pointed out, the "name/cti/kebab" transform creates the name "background-base", which is not a valid variable name in Javascript. 

[...]

PR #110, opened 2018-07-26
es6 format fix

dbanksdesign:
*Issue #, if available:* #108

*Description of changes:* Adding a "js" transformGroup that won't break js files and fixing the javascript/es6 format example code. 

### DIFF
--- a/docs/formats.md
+++ b/docs/formats.md
@@ -42,7 +42,7 @@ Formats are better if there is little to no boilerplate code like a flat SCSS va
 
 Creates a CSS file with variable definitions based on the style dictionary
 
-**Example**  
+**Example**
 ```css
 :root {
   --color-background-base: #f0f0f0;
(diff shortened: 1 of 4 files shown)


didoo:
Wow, @dbanksdesign you were super fast! Only one potential suggestion: what would you think of having two different "js" transformGroup available, so people can choose their preferred "case" convention? 
```
js/pascal → [..., 'name/cti/pascal', ...]
js/constant → [..., 'name/cti/constant', ...] (or maybe js/uppercase, or something more semantic ?)
```

dbanksdesign:
I think its a delicate balance with how many transformGroups we include in the framework. Especially because they just define groups of transforms, we don't want to get too granular with transformGroups. A user could use transforms: ['attribute/cti', 'name/cti/constant', 'size/rem', 'color/hex'] or some other version if they want a different naming style.

Issue #111, opened 2018-07-31
resolveObject is mutating the properties object


dbanksdesign:
This is the offending line:
https://github.com/amzn/style-dictionary/blob/master/lib/utils/resolveObject.js#L29

`_.clone()` is only a *shallow* copy, so when resolveObject is called it mutates the original properties object. This can have dire consequences because that means the properties object could change from platform to platform. 

Fix: change to `_.cloneDeep`

PR #112, opened 2018-07-31
Fix: #111, resolveObject no longer mutates the original properties ob…

dbanksdesign:
*Issue #, if available:* #111 

*Description of changes:* Make resolveObject no longer mutate the original object by using _.cloneDeep instead of _.clone

Unit test added

### DIFF
--- a/lib/utils/resolveObject.js
+++ b/lib/utils/resolveObject.js
@@ -26,7 +26,7 @@ var updated_object, regex, options;
 function resolveObject(object, opts) {
   options = _.extend({}, defaults, opts);
 
-  updated_object = _.clone(object);  // This object will be edited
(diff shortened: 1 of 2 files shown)


chazzmoney:
The cloneDeep change passes all tests, but definitely could change functionality. I'm ok with pushing, but maybe we should consider what the right versioning is. It isn't a breaking change necessarily... but maybe it could be?

Otherwise, LGTM!

dbanksdesign:
@chazzmoney, correct and this would go in a minor release which is why the PR is on the develop branch. When we are ready for a next release we will merge develop into master.

chazzmoney:
Doing a minor release anyway, LGTM

PR #114, opened 2018-08-08
JavaScript config files

elliotdickison:
This change enables the user to use a JavaScript module for their
config file. The goal of this change is to enable more flexible
configuration similar to Webpack or Babel. For example filter
functions, which previously could only be provided using the Node
API, can now be included directly in a JS config file.

JavaScript config files could also eventually remove the need for
the register APIs because you could simply pass formatters,
transformers, etc. as functions directly in the config instead of
registering them with an alias and then referencing that alias in
the config. This could potentially provide a simpler API very
similar to the way webpack handles plugins:
https://webpack.js.org/configuration/plugins/. Just a thought
though. For now the default config file `./config.json` and the
ability to continue to use JSON config files has not changed.

### DIFF
--- a/lib/extend.js
+++ b/lib/extend.js
@@ -13,7 +13,7 @@
 
 var combineJSON = require('./utils/combineJSON'),
     deepExtend = require('./utils/deepExtend'),
-    fs = require('fs-extra'),
+    resolveCwd = require('resolve-cwd'),
     _ = require('lodash'),
     chalk = require('chalk');
 
@@ -81,13 +81,13 @@ var combineJSON = require('./utils/combineJSON'),
 function extend(opts) {
   var options, to_ret;
 
-  // Overloaded method, can accept a string as a path that points to a JSON file
-  // or a plain object. Potentially refactor.
+  // Overloaded method, can accept a string as a path that points to a JS or
+  // JSON file or a plain object. Potentially refactor.
   if (_.isString(opts)) {
-    options = fs.readJsonSync(opts);
+    options = require(resolveCwd(opts));
   } else {
     options = opts;
-  }
+  }``
 
   // Creating a new object and copying over the options
   // Also keeping an options object just in case
--- a/test/configs/test.js
+++ b/test/configs/test.js
@@ -1,0 +1,1 @@
+module.exports = require("./test.json")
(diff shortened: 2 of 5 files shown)


shaunbent:
❤️ Love this, create contribution @elliotdickison.

lib/extend.js
```diff
@@ -81,10 +81,10 @@ var combineJSON = require('./utils/combineJSON'),
 function extend(opts) {
   var options, to_ret;
 
-  // Overloaded method, can accept a string as a path that points to a JSON file
-  // or a plain object. Potentially refactor.
+  // Overloaded method, can accept a string as a path that points to a JS or
+  // JSON file or a plain object. Potentially refactor.
   if (_.isString(opts)) {
-    options = fs.readJsonSync(opts);
+    options = require(path.resolve(process.cwd(), opts));
```
dbanksdesign: Just a conversation starter, the recent addition to do something similar for style dictionary files (#89) used the resolve-cwd module. Open question to you and @jreichenberg: should we use resolve-cwd here as well? Should we remove the resolve-cwd module and use this way in combineJSON.js? Does it matter?

test/configs/test.js
```diff
@@ -0,0 +1,87 @@
+module.exports = {
```
dbanksdesign: Would it be silly to just require the test.json file here so we don't have to keep the 2 files in sync? I mean, we probably won't be changing these much as the tests would break if we did. I dunno just thinking out loud.

lib/extend.js
elliotdickison: I missed that dependency. Not sure why it's necessary, but I don't really care one way or the other happy to add it in for consistency. @jreichenberg what's the value add of that dependency?

jreichenberg: @elliotdickison See the discussion in 89: https://github.com/amzn/style-dictionary/pull/89#discussion_r189397621

Both the native node require() statement and readJsonSync will parse any JSON you throw at it. The difference is that native require() will resolve packages in the process (yay!) but needs some help resolving a path relative to the cwd (boo!). So kind of depends on the desired capability.  I know that doesn't help much...

elliotdickison:
Switched to using `resolveCwd` to resolve the config file path, gutted `configs/test.js` to just re-export `configs/test.json` as javascript.

PR #121, opened 2018-09-05
fix #113, better comments handling in properties declarations

### LINKED ISSUE #113: Properties to generate comments are not consistent between platforms?
Can it be that the "comment" attribute is expected in different places for different platforms? In that case, maybe we could or unify it or support both (to avoid breaking someone's setup).

on Web/Scss:
```
if (prop.attributes.comment)
    to_ret_prop = to_ret_prop.concat(' // ' + prop.attributes.comment);
```
https://github.com/amzn/style-dictionary/blob/f385fc00540de7297b282b0851e45195d1b02eb2/lib/common/formats.js#L28-L30

on Android:
```
<% if (prop.comment) { %><!-- <%= prop.comment %> --><% } %>
```
https://github.com/amzn/style-dictionary/blob/c2c31cb381be05d8324838361188879cd8685f3c/lib/common/templates/android/dimens.template#L26

didoo:
*Issue #, if available:* #113 

*Description of changes:*
I have updated the templates and formats, so that all the comments are expected as `prop.comment`, updated the tests and finally updated also the documentation. 

Notice: I don't know if that's enough or I have missed something (for example the `xxx.h.template` and `xxx.m.template` have been left untouched). Please let me know if there something else to do/fix.

### DIFF
--- a/docs/package_structure.md
+++ b/docs/package_structure.md
@@ -38,6 +38,7 @@ The default way is to use a config.json file in the root of your package. Here i
       "transforms": ["attribute/cti", "name/cti/snake", "color/hex", "size/remToSp", "size/remToDp"],
       "buildPath": "build/android/src/main/res/values/",
       "files": [{
+        "filter": { "category": "color" },
         "destination": "style_dictionary_colors.xml",
         "template": "android/colors"
       }]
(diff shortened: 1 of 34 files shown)


test/formats/es6Constants.js
```diff
@@ -11,42 +11,42 @@
  * and limitations under the License.
  */
 
-var assert  = require('chai').assert,
-    formats = require('../../lib/common/formats');
+// var assert  = require('chai').assert,
```
dbanksdesign: Did you mean to comment all these lines out?

docs/package_structure.md
```diff
@@ -110,6 +110,27 @@ You can organize and name your style properties however you want, there are no r
 
 Also, the CTI structure provides a good mechanism to target transforms for specific kinds of properties. All of the transforms provided by the framework use the CTI structure to know if it should be applied. For instance, the 'color/hex' transform only applies to properties of the category 'color'.
 
+You can also add a _comment_ to a style property:
```
dbanksdesign: Love that you added this to the documentation.

lib/common/formats.js
```diff
@@ -25,7 +25,7 @@ function variablesWithPrefix(prefix, properties) {
   return _.map(properties, function(prop) {
       var to_ret_prop = prefix + prop.name + ': ' + (prop.attributes.category==='asset' ? '"'+prop.value+'"' : prop.value) + ';';
 
-      if (prop.attributes.comment)
+      if (prop.comment)
```
dbanksdesign: Should we add this to the javascript/es6 format on line 267 below?

didoo: yes we should. sorry I missed it*. done.

* re. this I'll open an issue for discussion, about templates vs formats.

test/formats/es6Constants.js
didoo: Strangely, when I pushed the first time, probably a pre-hook or something triggering an eslint check was complaining about unused variables, but now not anymore. So I have reverted the lines (that actually are not used, so in theory should be commented).

Issue #122, opened 2018-09-05
CSS TransformGroup


dbanksdesign:
In the docs, https://amzn.github.io/style-dictionary/#/formats, it gives an example of a "css" transformGroup even though the framework does not have one built in. We should either change that to "scss" or add a "css" transformGroup. 

Another issue this causes is that when someone uses a non-existent transformGroup, the framework does not throw an error, which it does if someone tries to use a non-existent transform or format. We should throw an error in this case.

didoo:
> when someone uses a non-existent transformGroup, the framework does not throw an error
I've learned it the hard way :)

PR #123, opened 2018-09-06
Fixing 122

dbanksdesign:
*Issue #, if available:* #122

*Description of changes:* Adding a check to throw an error if a user tries to use a transformGroup that doesn't exist. Also adding a 'css' transformGroup so the documentation makes sense. 


chazzmoney:
LGTM

chazzmoney:
LGTM

Issue #124, opened 2018-09-06
Escape strings in ES6 format?


didoo:
Recently I have added a web font stack in a token, like this:
```
{
    ...
    "font-family": {
        "value": "Calibri, 'Helvetica Neue', Helvetica, Arial, sans-serif"
    }
    ...
}
```
which contains this kind of `'` quote. But in the ES format function what we do is to just concatenate the strings:
https://github.com/amzn/style-dictionary/blob/1bdd60b332aadb487df3d176292951a58627150a/lib/common/formats.js#L264-L269

This compiles to:
```
export const [...]_FONT_FAMILY = 'Calibri, 'Helvetica Neue', Helvetica, Arial, sans-serif'
```
which is not valid. Probably we should escape, somehow, the value `prop.value` (in theory also the `prop.name`, but let's be conservative). 

Two questions:
1) do you think we should do it? or are there are ways to avoid this problem? any other ideas?
2) in case we do it, what else should we consider as something to escape? other characters that we should consider special?

chazzmoney:
Fixed in #127. Available in 2.4.0.

PR #127, opened 2018-09-06
Escape quotes es6

chazzmoney:
Suggested fix for #124 

Make sure we escape things as needed to be able to receive the correct values in ES6.

I was going to add tests in tests/formats/es6Constants.js but then I realized after reading that file that I'm not sure I'm smart enough...


chazzmoney:
For tests.... maybe we:

1. gen the file
2. pull the text of the file
3. create string containing tests
4. append 3 to 2
5. use new Fn(text) to create a function with the js and tests in it
6. execute that fn

Maybe this will work?

dbanksdesign:
I think this looks good. For testing, we could probably use babel to transpile es6 > es5. I found this article, it may work:
http://krasimirtsonev.com/blog/article/using-mocha-with-es6-spec-files

chazzmoney:
maybe we just skip the tests for now? 😳

Issue #107, opened 2018-07-25
Make the header text returned by fileHeader() optional


didoo:
When I process the design tokens, even if the actual values didn't change, the generated file contains a different header, like in this case:

<img width="453" alt="screenshot_1636" src="https://user-images.githubusercontent.com/686239/43195769-ed72ba54-8ffd-11e8-95ef-edfb7073e3f5.png">

What do you think of having this header optional? As far as I can see in the [formats.js](https://github.com/amzn/style-dictionary/blob/f385fc00540de7297b282b0851e45195d1b02eb2/lib/common/formats.js) file we have access to the config parameter.

Collecting feedbacks before actually attempting to submit a PR :)

dbanksdesign:
Makes sense to me. I have no objections. My suggestion would be to have it configurable but make the default on so as to not change the default behavior.

chazzmoney:
I also like making this as an option. It will enable people to more intuitively see style changes in a code repo. Basically, the file only changes when the style changes - simple.

@didoo did you want to give it a shot still?

didoo:
@chazzmoney @dbanksdesign is it still ok if I pick this one? (is required if we want to do snapshot testing, otherwise they would always fail :) ) 
Or instead we could simply remove it at all (less complexity in the codebase) and people that actually need it can rely on custom formats/templates.
wdyt? I don't have strong opinions, slightly prefer removing it, but I'm ok with both options.

didoo:
I have done a spike to understand the impact of this change. You can see the code here:
https://github.com/didoo/style-dictionary/commit/2de87aaad5c5f37495c199b71bd47b050f75ba86
https://github.com/didoo/style-dictionary/commit/ccb9f2165c6a2e81a48c750d7943503063be1319

This is in the hypothesis to use this format for the config:
```
{
  "source": [
    "properties/**/*.json"
  ],
  "platforms": {
    "scss-variables": {
      "transformGroup": "scss",
      "buildPath": "build/scss/",
      "files": [{
        "destination": "_variables.scss",
        "format": "scss/variables"
      }],
      "options": {
        "hideDoNotEdit": true,
        "hideTimestamp": true
      }
    },
    ...
```
with "options" as a new block where we can imagine to store the options for the plarform declaration.

As you see, it's quite dirty if we want to 1) make optional the header/timestamp and 2) make it visible by default for backward compatibility.
[...]

PR #132, opened 2018-09-13
Introduce option to control the generation of the "Do not edit" header

didoo:
*Issue #107*

*Description of changes:*
As discussed in #107, I have introduced a way to control the presence of a "Do not edit (plus timestamp)" header in the generated files. 

For backward compatibility, by default the header is generated, and the user has to explicitly set `showHeading=false` to disable it.

I have also updated the documentation accordingly. Let me know if there is something else to update that I may have missed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,8 @@ This tells the style dictionary build system how and what to build. The default 
 | platform.file.destination | String (optional) | Location to build the file, will be appended to the buildPath. |
 | platform.file.format | String (optional) | Format used to generate the file. Can be a built-in one or you can create your own. Must declare a format or a template. |
 | platform.file.template | String (optional) | Template used to generate the file. Can be a built-in one or you can create your own. [More on formats and templates](https://amzn.github.io/style-dictionary/formats_and_templates) |
+| platform.options | Object (optional) | A set of extra options associated with the file. |
+| platform.options.showHeader | Boolean | If the generated file should have a "Do not edit + Timestamp" header (where the format supports it). By default is "true". |
 
 ### Properties
 ```json
--- a/README.md
+++ b/README.md
@@ -142,8 +142,8 @@ This tells the style dictionary build system how and what to build. The default 
 | platform.file.destination | String (optional) | Location to build the file, will be appended to the buildPath. |
 | platform.file.format | String (optional) | Format used to generate the file. Can be a built-in one or you can create your own. Must declare a format or a template. |
 | platform.file.template | String (optional) | Template used to generate the file. Can be a built-in one or you can create your own. [More on formats and templates](https://amzn.github.io/style-dictionary/formats_and_templates) |
-| platform.options | Object (optional) | A set of extra options associated with the file. |
-| platform.options.showHeader | Boolean | If the generated file should have a "Do not edit + Timestamp" header (where the format supports it). By default is "true". |
+| platform.file.options | Object (optional) | A set of extra options associated with the file. |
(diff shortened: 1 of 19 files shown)


lib/common/formats.js
```diff
@@ -34,7 +40,7 @@ function variablesWithPrefix(prefix, properties) {
 }
 
 function iconsWithPrefix(prefix, properties, config) {
-  return fileHeader() + _.chain(properties)
+  return fileHeader(config.options) + _.chain(properties)
```
dbanksdesign: `config` here is referencing the platform object and not the file object, like `this` refers to inside the formatter function (like on line 74 below). Should we change the 3rd argument to just be 'options' and then the formatter passes only the options object? I'm assuming we are only going to do file level options for this PR and do the cascading options in a later one.

lib/common/formats.js
```diff
@@ -12,13 +12,19 @@
...
 
-function fileHeader() {
-  return '/**\n' +
-         ' * Do not edit directly\n' +
-         ' * Generated on ' + new Date() + '\n' +
-         ' */\n\n';
+function fileHeader(options) {
+  var to_ret = '';
+  // for backward compatibility we need to have the user explicitly hide them
+  var showHeader = (options) ? options.showHeader : true;
```
dbanksdesign: What are your thoughts about renaming this? I'm thinking if we eventually want options to cascade (root > platform > file), we might want this option name to be more explicit it is about files. Open to suggestions, maybe showFileComment ?

lib/common/formats.js
```diff
@@ -12,13 +12,19 @@
  */
 
 var _ = require('lodash');
-var module_def = 'module.exports = ';
```
dbanksdesign: Nice, makes sense to remove this as it's only being used once.

lib/common/formats.js
dbanksdesign: Now that I look at it, our iconsWithPrefix function is a bit wonky and confusing. Maybe for now, what if we removed the `fileHeader()` in iconsWithPrefix and moved it to the formatter functions themselves so it matches variablesWithPrefix ?

README.md
```diff
@@ -142,6 +142,8 @@ This tells the style dictionary build system how and what to build. The default
 | platform.file.destination | String (optional) | Location to build the file, will be appended to the buildPath. |
 | platform.file.format | String (optional) | Format used to generate the file. Can be a built-in one or you can create your own. Must declare a format or a template. |
 | platform.file.template | String (optional) | Template used to generate the file. Can be a built-in one or you can create your own. [More on formats and templates](https://amzn.github.io/style-dictionary/formats_and_templates) |
+| platform.options | Object (optional) | A set of extra options associated with the file. |
```
dbanksdesign: I think this should be `platform.file.options` right?

didoo: You're right. My bad, it was a leftover of the other "variant".

lib/common/formats.js
didoo: yes, great idea!

lib/common/formats.js
didoo: I went for `showFileHeader`. Do you think it's clear enough?

chazzmoney:
Switching to develop branch

gearsandcode:
@didoo - can you edit your initial comment here to reflect the name of the option as showHeader: false vs. showHeading=false please.

didoo:
@gearsandcode is it ok now?

Issue #138, opened 2018-10-01
Add 'json/flat' format


dbanksdesign:
From the discussion on #58 , we want to add different types of JSON formatting for different purposes. This format should output a JSON string that only has 1 level where the keys are the property name and the value is the property value. For example, this style dictionary JSON:
```json
{
  "size": {
    "font": {
      "small": { "value": "13px" },
      "medium": { "value": "15px" },
      "large": { "value": "17px" },
      "base": { "value": "{size.font.medium.value}" }
    }
  }
}
``` 

Would output this in the 'json/flat' formatter:
```json
{
  "SizeFontSmall": "13px",
  "SizeFontMedium": "15px",
  "SizeFontLarge": "17px",
  "SizeFontBase": "15px"
}
```

You can have a look at how the other [formatters](https://github.com/amzn/style-dictionary/blob/master/lib/common/formats.js) are written to see how to write a formatter and where it should go. 

Make sure to include proper jsdoc comments and unit test!

ericbreno:
Looking into it, interesting!

chazzmoney:
Merged, closing

kerryj89:
When using `json/flat` it keeps the hyphens and I'm assuming other symbols.  Is there a way to easily transform it to not include them so that we can simply import the json and do `tokens.sizeFontSmall` and not `tokens['size-font-small']`?

jorenbroekema:
Yes, by adding a name transform that converts token names to camelCase

E.g.
```js
{
  "source": [
    "tokens.json"
  ],
  "platforms": {
    "json": {
      "transforms": [
        "name/camel"
      ],
      "buildPath": "build/json/",
      "files": [
        {
          "destination": "vars.json",
          "format": "json/flat"
        }
      ]
    }
  }
}
```

[...]

Issue #118, opened 2018-08-22
resolveObject - circular definition error if more than 2 references in a value?


didoo:
I don't know if this is an error or an expected behavior. 

Let's consider these declarations:

```
{
  "prop1" : { "value": "test1 value" },
  "prop2" : { "value": "test2 value" },
  "prop3" : { "value": "test3 value" }
}
```

When I reference two property values (as aliases) in another property the compilation works as expected. 
This:
```
"prop123" : { "value": "{prop1.value}, {prop2.value} and some extra stuff" }
```
generates in output this (eg. CSS variables):
```
--prop-123: test1 value, test2 value and some extra stuff;
```

But when I reference three property values (as aliases) the compilation throws an error.
This:
```
"prop123" : { "value": "{prop1.value}, {prop2.value} and {prop3.value}" }
```
generates this error:
<img width="1206" alt="screenshot_1682" src="https://user-images.githubusercontent.com/686239/44462970-7eb3bd00-a616-11e8-8f88-38bae6fd5f0c.png">

Oddly enough, this happens also if I do something similar, using two aliases but referencing the same prop:
```
"prop22" : { "value": "{prop2value}, {prop2.value}" }
```
throws the same kind of error.

I looked in the code for the [resolveObject](https://github.com/amzn/style-dictionary/blob/master/lib/utils/resolveObject.js) method, but I am not sure where the culprit is.

Is this an issue, or actually is some kind of safeguard in place to avoid too many recurrences?

didoo:
Mmm, I have also noticed this thing: the error message string says:
```
Error: Variable reference for 'value' resolves to circular definition cycle of 'est1 value test2 value and {prop3.value':
```

If you notice the beginning and end of the string are missing the initial and final characters (_'est1'_ and  '_{prop3.value'_) so maybe is in the way the regexes are executed? in the resolveObject file I see this:

```
 regex = new RegExp(
    '\\' +
    options.opening_character +
    '([^' +
    options.closing_character +
    ']+)' +
    '\\' +
    options.closing_character, 'g'
  );
```
so maybe is considering the whole multiple aliases as a single declaration (but then, why it works with two?)

dbanksdesign:
I think I know what is going on here, we recently made a change to have better error messages like collect all circular definitions and all reference not found so that you can see all the issues rather than seeing one, fixing it, re-building, repeating. I'm pretty sure the offending line is this one:
[...]

PR #145, opened 2018-10-02
Adding 'json/flat' formatter

ericbreno:
*Issue:* #138 

*Description of changes:* Added the 'json/flat' formatter to file `formats.js`. Simple recursive strategy to build up keys with it's path and evaluate values as strings directly or as a reference to another value in `dictionary` input object.

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -347,5 +347,69 @@ module.exports = {
       })
       .value();
     return JSON.stringify(to_ret, null, 2);
+  },
+
+  /**
+   * Creates a flatten JSON from input, capitalizing and then concatenating 
+   * each key with it's pre-key recursively.
+   *
+   * @memberof Formats
+   * @kind member
+   * @example
+   * ```json
+    {
+      "size": {
+        "font": {
+           "small": { "value": "13px" },
+           "medium": { "value": "15px" },
+           "large": { "value": "17px" },
+           "base": { "value": "{size.font.medium.value}" }
+        }
+      }
+    }```
+   * 
+   * Creates:
+   * ```json
+    {
+      "SizeFontSmall": "13px",
+      "SizeFontMedium": "15px",
+      "SizeFontLarge": "17px",
+      "SizeFontBase": "15px"
+    }
+   * ```
+   */
+  'json/flat': function(dictionary) {
+    var evaluateProp = function(property) {
+      // property value is in format '{*}'
+      var isReference = /^\{.*\}/g.test(property);
+      if (isReference) {
+        // removing brackets, taking 'content' out of '{content}'
+        property = property.substr(1, property.length - 2);
+        return _.get(dictionary, property);
+      }
+      return property;
+    };
+
+    // recursive, initial prefix is none (empty string)
+    // iterate over keys capitalizing and concatenating them 
+    // until "value" key is found, then evaluate it's value
+    var deepFlatten = function(object, prefix) {
+      return _.toPairs(object)
+        .reduce(function(output, pair) {
+          var key = pair[0],
+            value = pair[1],
+            isValueProperty = key === 'value',
+            newKey = isValueProperty ? prefix : prefix + _.capitalize(key);
+          if (_.isString(value)) {
+            output[newKey] = evaluateProp(value);
+          } else {
+            Object.assign(output, deepFlatten(value, newKey));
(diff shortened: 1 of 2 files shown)


lib/common/formats.js
```diff
@@ -347,5 +347,69 @@ module.exports = {
...
+      "SizeFontMedium": "15px",
+      "SizeFontLarge": "17px",
+      "SizeFontBase": "15px"
+    }
+   * ```
+   */
+  'json/flat': function(dictionary) {
+    var evaluateProp = function(property) {
+      // property value is in format '{*}'
+      var isReference = /^\{.*\}/g.test(property);
```
didoo: there is an open discussion if continue supporting multiple references in the same token value or not (see #118). in theory, the value could be something like `{abc.def} {ghi.jkl}` in that case the code may not work as expected.
@dbanksdesign suggestions here? what do we do?

lib/common/formats.js
```diff
@@ -347,5 +347,69 @@ module.exports = {
...
+        property = property.substr(1, property.length - 2);
+        return _.get(dictionary, property);
+      }
+      return property;
+    };
+
+    // recursive, initial prefix is none (empty string)
+    // iterate over keys capitalizing and concatenating them 
+    // until "value" key is found, then evaluate it's value
+    var deepFlatten = function(object, prefix) {
```
didoo: probably I am missing something, but I think that at this point in the pipeline you already have the `prop.name` and the `prop.value` (and potentially also the `property.path`), no? can you compare with other formats and see if they may be of help?

ericbreno: Didn't get it. Where I'd have `prop.name` or `prop.path`? There's another pre-processor for JSON that can do this?
In this code I'm iterating recursively (like a DFS) and also saving the actual path until now to build the final path/name for each property

lib/common/formats.js
ericbreno: If it's needed to keep support, I just need to know how it should work. Waiting.

lib/common/formats.js
```diff
@@ -347,5 +347,69 @@ module.exports = {
...
+
+  /**
+   * Creates a flatten JSON from input, capitalizing and then concatenating 
+   * each key with it's pre-key recursively.
+   *
+   * @memberof Formats
+   * @kind member
+   * @example
+   * ```json
+    {
```
dbanksdesign: I think you need to have asterisks at the beginning of these lines because they need to be part of the comment block. Same for the other JSON examples in the comment below

lib/common/formats.js
dbanksdesign: The style dictionary system already handles the references at this point before getting to the formatter, so you should not be trying to resolve the reference. I think the confusion here is in the example I gave in the issue is referencing the input JSON to a style dictionary, which gets processed by the style dictionary before being passed to a formatter. Take a look at https://amzn.github.io/style-dictionary/#/build_process to see what is happening.  Apologies for any confusion.

lib/common/formats.js
dbanksdesign: @didoo is correct, see my comment above about Style Dictionary already processing the input JSON. The formatter function takes 2 arguments, a `dictionary` and a `config`. Both are objects, the important one for this is the `dictionary` which has 2 attributes: `properties` and `allProperties`. `allProperties` is already a flattened list of the design tokens (properties) which have already been processed. If you take a look at the 'javascript/es6' format you can see this in action.

test/formats/jsonFlat.js
```diff
@@ -0,0 +1,44 @@
...
+
+var assert = require('chai').assert,
+  formats = require('../../lib/common/formats');
+
+var formatter = formats['json/flat'];
+
+describe('formats', function () {
+  describe('json/flat should', function () {
+    it('evaluate complex object with property referencing another value', function () {
+      var dictionary = {
```
dbanksdesign: Take a look at test/formats/scssVariables.js for a better example how to test this formatter.

chazzmoney:
@ericbreno Are you still pursuing this? Let us know if you need additional pointers.

chazzmoney:
Closing as there has been no movement. Feel free to re-open if you have new changes.

didoo:
@chazzmoney I am picking up this one (we need a json/flat format among the pre-defined formats)

Issue #141, opened 2018-10-01
Add test: lib/utils/flattenProperties.js


dbanksdesign:
Note: Work off of the 'jest' branch. That is our new branch to work off of for the testing upgrades. When everything is ready we will merge that branch back into develop and mainline.

chrkhl:
working on this ;-)

dbanksdesign:
Closed with #146

PR #146, opened 2018-10-02
Adding tests for lib/utils/flattenProperties.js

bpsagar:
*Issue:* #141

*Description of changes:* Added tests for `lib/utils/flattenProperties.js`.

### DIFF
--- a/__tests__/utils/flattenProperties.test.js
+++ b/__tests__/utils/flattenProperties.test.js
@@ -37,13 +37,18 @@ describe('utils', () => {
         }
       };
       var expected_ret = [
-        { 'value': '#000000' },
-        { 'value': '#FFFFFF' }
+        properties.black,
+        properties.white
       ];
-      var sortedExpectedValues = expected_ret.map(v => v.value).sort();
+      var sortFn = function (a, b) {
+        if (a.value > b.value) return 1;
+        if (b.value > a.value) return -1;
+        return 0;
+      }
+      var sortedExpectedRet = expected_ret.sort(sortFn);
       var ret = flattenProperties(properties);
-      var sortedReturnedValues = ret.map(v => v.value).sort();
-      expect(sortedReturnedValues).toEqual(sortedExpectedValues);
+      var sortedRet = ret.sort(sortFn);
+      expect(sortedRet).toEqual(sortedExpectedRet);
     });
 
     it('should return nested leaf node values as an array', () => {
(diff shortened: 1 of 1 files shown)


dbanksdesign:
Looks like the tests written are passing in the CI (just the date/snapshot tests are the ones failing which have been fixed on the jest branch)

__tests__/utils/flattenProperties.test.js
```diff
@@ -0,0 +1,70 @@
...
+      };
+      var expected_ret = [
+        { 'value': '#000000' },
+        { 'value': '#FFFFFF' }
+      ];
+      var sortedExpectedValues = expected_ret.map(v => v.value).sort();
+      var ret = flattenProperties(properties);
+      var sortedReturnedValues = ret.map(v => v.value).sort();
+      expect(sortedReturnedValues).toEqual(sortedExpectedValues);
+    });
```
chazzmoney: For this last test, instead of testing the values of the value key, I think we should test the objects against each other. This will enable us to add in details to the objects later and make sure they land correctly. Maybe something like:

```
it('should return nested leaf node values as an array', () => {
      var properties = {
        'color': {
          'black': {
            'value': '#000000'
          },
          'white': {
            'value': '#FFFFFF'
          }
        }
      };
      var expected_ret = [
        properties.color.black,
        properties.color.white
      ];
      var sortFn = function(a,b) {
            if (a.value > b.value)
                  return 1;
            else if (b.value > a.value)
                  return -1;
            else
                  return 0;
      };
      var sortedExpectedValues = expected_ret.sort(sortFn);
      var ret = flattenProperties(properties);
      var sortedReturnedValues = ret.sort(sortFn);
      expect(JSON.stringify(sortedReturnedValues)).toEqual(JSON.stringify(sortedReturnedValues));
    });
  });
```

bpsagar: Yes it makes sense! Shall I add this and update my pull request?

__tests__/utils/flattenProperties.test.js
```diff
@@ -0,0 +1,80 @@
...
+          'white': {
+            'value': '#FFFFFF'
+          }
+        }
+      };
+      var expected_ret = [
+        properties.color.black,
+        properties.color.white
+      ];
+      var sortFn = function (a, b) {
```
didoo: a couple of small things: I see the sortFn declared twice; could it be declared above (just after `describe('utils'` or grouping the last two tests in another `describe`) and shared between tests? or alternatively, since I see in some tests we are using lodash, can we use some existing method of lodash?

chazzmoney:
Chatted with Danny and he has a good suggestion for the sort function. After that change we should be good to go!

__tests__/utils/flattenProperties.test.js
dbanksdesign: You could also just use lodash's [sortBy](https://lodash.com/docs/4.17.10#sortBy) function instead of declaring a new function.

PR #148, opened 2018-10-02
test(all): Fix for all tests to match the date and remove of mockdate

davixyz:
*Issue #, if available:*

*Description of changes:*

Tried to run the project, failed some tests went ahead and fixed it. I see that there were a lot of changes around of this, is this the right solution? seems to pass for now.

I checked in the package.json and package-lock.json since I removed mockdate but I could remove these from the PR if needed.

inspiration https://github.com/facebook/jest/issues/2234

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #140, opened 2018-10-01
Add unit tests for es6 format


dbanksdesign:
We are working to have better testing coverage and migrating our tests to Jest from Mocha. With some work, we should be able to test the es6 format, which we have been unable to do because Mocha does not understand es6. 

https://github.com/facebook/jest#additional-configuration

DarthOstrich:
I've been writing a lot of tests using Jest recently, so I'd like to hop on board here. I'm assuming you want the other issues completed, or are you just looking for a complete overhaul of every single test?

dbanksdesign:
Sorry @DarthOstrich looks like @davixyz beat you to the punch. One major thing we are doing right now is migrating our tests from Mocha to Jest, which is mostly done, but in the process we want to go through our tests, make sure they are all functioning properly, add tests that should be there, etc. I made a project to track this: https://github.com/amzn/style-dictionary/projects/4

If you want to hop in to our new Jest tests and see if there is anything that could be improved, that would be super helpful. For this project we have a 'jest' branch on the repository we are working off of. Right now on the branch, the previous tests have been migrated, but we are still looking to see where our tests could improve, either by adding more tests where there aren't any yet, or general housekeeping in the testing. We are also keen to add code coverage with Istanbul/NYC as well. If any of that interests you, take a look at the 'jest' branch and let me know what you think. If you have any questions or anything let me know!

didoo:
@DarthOstrich see the conversation here: https://github.com/amzn/style-dictionary/pull/133
there are a few important points that are still open for discussions, and feedback/ideas/contributions are more than welcome :)

PR #149, opened 2018-10-02
test(javascript/es6): Add test for es6

davixyz:
*Issue #, if available:*
https://github.com/amzn/style-dictionary/issues/140

*Description of changes:*
Add babel so jest can understand es6
https://jestjs.io/docs/en/getting-started.html#using-babel

Add test for es6

Note, this will probably keep failing until https://github.com/amzn/style-dictionary/pull/148 is merged


__tests__/formats/all.test.js
```diff
@@ -39,11 +39,7 @@ describe('formats', () => {
   const globalDate = global.Date;
 
   beforeAll(() => {
-    global.Date = class extends Date {
```
davixyz: @dbanksdesign @chazzmoney @didoo I found a better way to do the constant date trick, in the previous way it was messing up when I added the ES6 support for JEST where it was kinda transforming this code and giving me a very cryptic error; But now it's fixed

davixyz:
Adding ES6 support like this is trivial with babel-preset-env but if we want to go full on with Babel 7.0 we need to add way more dependencies like it explains here:
https://jestjs.io/docs/en/getting-started

Do we want to explore more in babel 7.0 or less dependencies?

dbanksdesign:
TBH my babel/jest knowledge is pretty limited, what benefit would we get with using full on babel? I think I'm good with this PR and we can maybe do full on babel in the future if we need to. Thanks for doing this!

__tests__/formats/all.test.js
dbanksdesign: As long as it works as expected! Nice work!

davixyz:
@dbanksdesign If we use full babel for the whole library then we could essentially re-write it completely in ES6 and transpile it to ES5 on build. That's of course a big effort (similar to this jest effort). Now this addition of babel-preset-env is actually something we would need to do anyways when re-writing the whole library to ES6 so it's good, in this case Jest is smart enough to kinda understand the es6 and use the preset-env to parse the ES6 on runtime

chazzmoney:
LGTM!

Issue #144, opened 2018-10-01
Add test: lib/transform/object.js


dbanksdesign:
Note: Work off of the 'jest' branch. That is our new branch to work off of for the testing upgrades. When everything is ready we will merge that branch back into develop and mainline.

chrkhl:
Hey,
i created a pull request #151 for this issue.

dbanksdesign:
Thank you @chrkhl !

PR #151, opened 2018-10-02
add tests for transform object

chrkhl:
Issue #144

Adds tests for transform object.

### DIFF
--- a/__tests__/transform/object.test.js
+++ b/__tests__/transform/object.test.js
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+ * the License. A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+var transformObject = require('../../lib/transform/object');
+
+var options = {
+  transforms: [
+    {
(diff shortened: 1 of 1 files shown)


__tests__/transform/object.test.js
```diff
@@ -0,0 +1,132 @@
...
+              "value": "16",
+              "comment": "the base size of the font"
+            },
+            "large": {
+              "value": "20",
+              "comment": "the large size of the font"
+            }
+          }
+        },
+        "color": "#FFFF00"
```
didoo: I think this should point to an object with a "value" key, like the other "tokens", or if you want to test that it outputs a property as is, maybe just use a "test" key with a "this is a test" string as value

__tests__/transform/object.test.js
```diff
@@ -0,0 +1,132 @@
...
+      };
+
+      const options = {
```
didoo: is this a duplicate of the declaration above?

__tests__/transform/object.test.js
```diff
@@ -0,0 +1,132 @@
...
+  ]
+};
+
+describe('transform', () => {
+  describe('object', () => {
+    it('does not crash when called without parameters', () => {
+      expect(transformObject()).toEqual({});
+    })
+
+    it('returns expected object', () => {
```
didoo: what do you think of splitting this test in smaller tests to check the different cases that I see in the code for `lib/transform/object` ("is plain object and has value", "is just plain object", "else")? no strong opinions on this, just an idea

dbanksdesign: I think smaller tests are better, they will be easier to debug if a test breaks. Really each test should test 1 thing.

__tests__/transform/object.test.js
dbanksdesign: @didoo  is correct.

chrkhl:
Hey @didoo @dbanksdesign ,
thanks a lot for your feedback. I adopted your proposed improvements.

chazzmoney:
LGTM!

PR #156, opened 2018-10-04
handle no command and invalid commands with friendly console output

chazzmoney:
Description of changes:

- Automatically show help info if someone runs style-dictionary with no commands
- Added friendly output if someone uses an invalid command.

### DIFF
--- a/bin/style-dictionary
+++ b/bin/style-dictionary
@@ -51,6 +51,12 @@ program
     styleDictionaryBuild();
   });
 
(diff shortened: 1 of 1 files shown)


chazzmoney:
I tested that the commands ran as expected. I did not test the output of those commands to see that nothing changed. I thought that might be overkill since the commands were running... Do you want me to double check that the files that are output are the same?

PR #152, opened 2018-10-03
Remove templates

chazzmoney:
First draft at solving #126 

This is intended to be a starting point. I'm sure things are not quite right here and we need ot make some updates before this is ready to roll. Would love feedback @didoo @davixyz @dbanksdesign @elliotdickison and anyone else.

Description of changes:
- Removed built-in templates and moved them to built-in formats.
- Added lots of warning messages anytime someone use a template in their config. Might need to do this better because it probably will show up multiple times. template: 'whatever' still works, but it pops up a warning.
- Added a warning message if someone uses registerTemplate, though it too continues to function correctly.
- Updated documentation to focus on formats and provide reasoning for removing templates.
- Updated tests.formats.all to have the expected 
- Fixed light logic issue with fonts.css template (if properties or properties.asset is undefined it produces a blank file now; crashed in this case previously)
- Switched eslint to ECMAScript 6 to support multi-line string literals (for my warning text)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #158, opened 2018-10-05
Fix errors and improve error messaging

chazzmoney:
Description of changes:
- fixed a bug with deepExtend
- fixed circular reference checking and error reporting
- fixed the resolveObject bug with multiple references (issue #118, via @didoo)
- added a test for #118 
- added some additional circular reference checking tests
- tried to create a standard method to deal with error handling, specifically when you need to allow a section to generate multiple errors and then report them all at once. This is challenging with our well defined module system because the errors need to be generated deep and then thrown shallow. I'm using a key based system to capture groups of errors together.

This will be fun to merge with the jest branch... 😬


chazzmoney:
@didoo @davixyz @dbanksdesign Would love feedback on this, especially the mechanism I'm using to try to catch groups of errors at once.

lib/utils/handleErrors.js
```diff
@@ -0,0 +1,101 @@
...
+
+  done: function (thread) {
+    if(unsavedErrors[thread] && unsavedErrors[thread].length) {
+      var errArr = getFinalThreadErrors(thread);
+      errors = errors.concat(errArr);
+    }
+  },
+
+  count: function (thread) {
+    return thread ? (unsavedErrors[thread] ? unsavedErrors[thread].length : 0) : errors.length;
```
didoo: is this the only way? also, looking at the code, looks like `thread` is always passed to the function so is always true

lib/utils/handleErrors.js
```diff
@@ -0,0 +1,101 @@
...
+  },
+
+  fetch: function (thread) {
```
didoo: can it be that it's used only in the tests, not in the actual code?

lib/utils/handleErrors.js
```diff
@@ -0,0 +1,101 @@
...
+    return thread ? unsavedErrors[thread] : errors;
+  },
+
+  clear: function (thread) {
```
didoo: can it be that is not used at all?

lib/utils/resolveObject.js
```diff
@@ -100,9 +96,28 @@ function compile_value(value, stack) {
...
-          if(stack.indexOf(reference)!==-1) {
-            stack.push(reference);
-            throw new Error('Variable reference for \'' + stack[0] + '\' resolves to circular definition cycle of \'' + reference + '\':  ' + stack.join(', ') );
+
+          // Compare to found circular references
+          var foundCircArr = Object.keys(foundCirc);
+          if (foundCircArr.indexOf(reference)!==-1) {
+            // If the current reference is a member of a circular reference, do nothing
+            // originally this was an error message, but that seems misleading because the circular reference itself is what needs to be fixed
+            // if(stack.filter(function(ref) {return foundCircArr.indexOf(ref)==-1}).length > 0) {
```
didoo: extra commented code to be removed, or is part of the comment above?

lib/utils/resolveObject.js
```diff
@@ -142,16 +158,10 @@ function selfRef(string, obj) {
   // If the reference doesn't change then it means
   // we didn't find it in the object
   if (ref === obj) {
-    throw new Error('Reference doesn\'t exist: ' + context + ' tries to reference ' + string + ', which is not defined');
+    handleErrors.add(ERROR_TYPE, 'Reference doesn\'t exist: ' + context + ' tries to reference ' + string + ', which is not defined');
```
didoo: do you have any idea if in this kind of errors it's possible to know also in which **file** the missing reference (or the collision) is happening? often doing refactoring of the properties it happened to me to see the error in the console, but I had to figure out which file was causing it. looking at the code, is not so easy/straightforward but maybe there is a simple way and I can't see it

test/utils/resolveObject.js
```diff
@@ -13,8 +13,8 @@
 
 var assert        = require('chai').assert,
     helpers       = require('../helpers'),
-    resolveObject = require('../../lib/utils/resolveObject');
-
+    resolveObject = require('../../lib/utils/resolveObject'),
+    handleErrors  = require('../../lib/utils/handleErrors');
```
didoo: regarding these tests, how do we do to "convert" them in the Jest branch?

bin/style-dictionary
```diff
@@ -57,19 +58,30 @@ program.on('command:*', function () {
   process.exit(1);
 });
 
+function displayErrors() {
```
didoo: why do you need a function to call another function? better function name, so more readable?

lib/utils/handleErrors.js
```diff
@@ -0,0 +1,101 @@
...
+        consoleOutputErrArr(errArr, chalkType);
+      }
+    }
+    else {
+      consoleOutputErrArr(errors, chalkType);
+      errors = [];
+    }
+  },
+
+  add: function (arg1, arg2) {
```
didoo: also, I just noticed that it's always called with both the arguments, so this means that you always have a thread, so you don't need to have this double handling of `errors` and `unsavedErrors` (but maybe I am missing something)

didoo:
@chazzmoney I've reviewed the PR code, and added some comments. Overall I understand the intent, and I'm ok with it, but at times I was struggling to follow the code of `handleErrors.js`. It's quite complex; probably I would try to see if there is a way to simplify it even more.

lib/exportPlatform.js
```diff
@@ -41,9 +44,16 @@ function exportPlatform(platform) {
   // values like "1px solid {color.border.base}" we want to
   // transform the original value (color.border.base) before
   // replacing that value in the string.
-  return resolveObject(
+  to_ret = resolveObject(
     transformObject(this.properties, platformConfig)
   );
+
+  if(handleErrors.count(ERROR_TYPE)) {
```
dbanksdesign: Do we need to check for the type of error? If there are other errors wouldn't we want to show them anyways?

lib/extend.js
```diff
@@ -111,18 +112,21 @@ function extend(opts) {
...
 
-    var props = combineJSON(options.source, true, function Collision(opts) {
-      if (options.log) {
-        var str = `Collision detected at: ${opts.path.join('.')}! Original value: ${opts.target[opts.key]}, New value: ${opts.copy[opts.key]}`;
-        if (options.log === 'warn') {
-          console.warn(chalk.keyword('orange')(str));
-        } else if (options.log === 'error') {
-          throw new Error(str);
-        }
+    var ERROR_TYPE = 'Property Value Collisions';
```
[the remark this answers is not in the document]
dbanksdesign: Maybe instead of these ERROR_TYPE strings we could actually subclass Error to have actually different Error types.  
https://gist.github.com/justmoon/15511f92e5216fa2624b

lib/extend.js
```diff
@@ -111,18 +112,21 @@ function extend(opts) {
...
+    var props = combineJSON(options.source, true, function Collision(prop) {
+      var str = `Collision detected at: ${prop.path.join('.')}! Original value: ${prop.target[prop.key]}, New value: ${prop.copy[prop.key]}`;
+      handleErrors.add(ERROR_TYPE, str);
+    });
+
+    if(handleErrors.count(ERROR_TYPE)) {
+      handleErrors.flush(ERROR_TYPE, options.log === 'error' ? chalk.bold.red : chalk.bold.orange);
```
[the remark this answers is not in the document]
dbanksdesign: I like @didoo's comment, and maybe we should put this code in the handleErrors class itself so we don't have to copy and paste that code every time.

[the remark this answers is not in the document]
dbanksdesign: Also we could add the count of errors in `throw new Error('Collisions detected')`

lib/utils/handleErrors.js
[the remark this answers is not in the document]
dbanksdesign: I feel like using an object rather than args might clean up the code below.

lib/utils/resolveObject.js
dbanksdesign: At this point, no, because we have already merged all the token JSON files together.. I mean we could find it, but we would have to go through the files again.

test/utils/resolveObject.js
```diff
@@ -119,34 +119,60 @@ describe('resolveObject', function() {
...
+    handleErrors.clear(ERR_TYPE);
+
+    resolveObject(helpers.fileToJSON(__dirname + '/../json_files/circular.json'));
+    assert.equal(handleErrors.count(ERR_TYPE), 1);
+    assert.equal(JSON.stringify(handleErrors.fetch(ERR_TYPE)), JSON.stringify([
+       'Circular definition cycle:  a, b, c, d, a'
+    ]));
+    handleErrors.clear(ERR_TYPE);
+
+    resolveObject(helpers.fileToJSON(__dirname + '/../json_files/circular_2.json'));
```
dbanksdesign: We should split these up into different tests

chazzmoney:
Thanks @didoo and @dbanksdesign! Will take a look at all your feedback and create another cut at this overall.

lib/utils/resolveObject.js
chazzmoney: I thought about this as well. I went through the code and I could actually show where I found the conflict. However, without significant additional work I would NOT know where the first occurrence was. I thought there were three options:

1. Do the significant additional work to report the positions of the conflicts, and make this PR even bigger/more confusing...
2. Only report the position of the collision. User will have to do a file search for the original source of the colliding properties.
3. Report nothing. User will have to do a file search for both the original and colliding properties.

Since the difference in experience between #2 and #3 seemed minimal, I decided to not go with #2. As I alluded to in my first paragraph, #1 seemed like too much to add to this PR.

I would still like to tackle this, but maybe after we get it merged? Should we add an issue to track it?

didoo: @chazzmoney I didn't mean to do it in this PR, I was just thinking out loud, to see if it's something that we may consider in the future. So whatever you choose to do, is ok for me.

lib/utils/handleErrors.js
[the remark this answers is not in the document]
chazzmoney: Converted to a custom error using https://gist.github.com/justmoon/15511f92e5216fa2624b

lib/utils/handleErrors.js
```diff
@@ -0,0 +1,101 @@
...
+  },
+
+  clear: function (thread) {
+    if (thread) {
+      unsavedErrors[thread] && delete unsavedErrors[thread];
+    }
+  }
+};
+
+module.exports = ErrorHandler;
```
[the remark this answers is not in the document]
chazzmoney: using ErrorHandler

lib/utils/resolveObject.js
chazzmoney: Eh, I thought it was explanatory since we have an if block with no actual code inside. Will delete the comment anyway.

lib/utils/resolveObject.js
chazzmoney: Added as issue #166

lib/extend.js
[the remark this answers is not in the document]
chazzmoney: I was trying to decide what to do about this actually. This is the ONLY place we use options.log.

I thought maybe we should delete it. I doubt anyone is using it as I don't think it is documented anywhere. We do have tests for it, which is why I left it in.

I'm also happy to go the route of options.log = error causing this interaction if thats what we deem as what should happen across the board. We might consider renaming the option as it makes me think of "log_level" which doesn't cause the execution to stop but only changes the level of output.

I think this is important to get settled before we call this work complete. Please let me know what you guys think.

lib/extend.js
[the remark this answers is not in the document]
chazzmoney: Doing what Danny suggested for subclassing.

Also, I'm leaving the declaration in mid-code here because it doesn't apply to the whole file - just the calls immediately below it. Instead, I'm moving the declaration of the error group in exportPlatform to be only next to the areas it is used. It is actually the fact that it is defined the same in exportPlatform and in resolveObject that it works as expected. Let me know if you think that I should make this more clear somehow / pass this information in?

bin/style-dictionary
chazzmoney: I had been passing arguments before. I agree it makes more sense to call flush directly.

lib/exportPlatform.js
chazzmoney: The intent of this code is to provide a grouping mechanism for errors. If there are other errors, they will be thrown immediately in the code executed much more deeply. If we make it out to this point, it suggests that we didn't have other errors so we should take our collected errors and display them all at once and then throw.

I could see how this looks like how we should handle errors in general. If thats the case I should rework ALL other times we spit out errors. Probably needs to be resolved in relationship to the options.log = error conversation as well.

test/utils/resolveObject.js
chazzmoney: Lets do this when we integrate with jest branch

chazzmoney:
@didoo @dbanksdesign Let me know what you think of this revision. There are some notes I put in to conversations that are now marked as outdated, but would love feedback on.

bin/style-dictionary
```diff
@@ -64,12 +65,19 @@ function styleDictionaryBuild(options) {
...
+      options.platform.forEach(function(platform) {
+        styleDictionary.buildPlatform( platform );
+      });
+    } else {
+      styleDictionary.buildAllPlatforms();
+    }
+    ErrorHandler.flush();
+  }
+  catch (e) {
+    ErrorHandler.flush();
```
dbanksdesign: Can all this stuff and the below in this file be removed? Should the style dictionary node module already flush out all the errors? If not it probably should so using the node module and the CLI should behave the same.

lib/exportPlatform.js
```diff
@@ -41,9 +42,17 @@ function exportPlatform(platform) {
   // values like "1px solid {color.border.base}" we want to
   // transform the original value (color.border.base) before
   // replacing that value in the string.
-  return resolveObject(
+  to_ret = resolveObject(
     transformObject(this.properties, platformConfig)
   );
+
+  var ERROR_GROUP = 'Property Reference Errors';
```
dbanksdesign: Can we just have these error group types on the ErrorHandler object itself, like ErrorHandler.referenceErrors or something so we don't have to keep defining ERROR_GROUP in different files and making sure their strings match up?

lib/extend.js
```diff
@@ -111,18 +113,27 @@ function extend(opts) {
...
+    var ERROR_GROUP = 'Property Value Collisions';
+    var props = combineJSON(options.source, true, function Collision(prop) {
+      var err = new CustomError(
+        `Collision detected at: ${prop.path.join('.')}! Original value: ${prop.target[prop.key]}, New value: ${prop.copy[prop.key]}`,
+        {group: ERROR_GROUP}
+      );
+      ErrorHandler.add(err);
+    });
+
+    if(ErrorHandler.count(ERROR_GROUP) > 0) {
```
dbanksdesign: Here you are using `if(ErrorHandler.count(ERROR_GROUP) > 0) {` and above you are using `if(ErrorHandler.count(ERROR_GROUP)) {`

lib/utils/errorHandler.js
```diff
@@ -0,0 +1,99 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+var chalk = require('chalk');
+
+var errors = [];
+var unsavedErrors = {};
```
dbanksdesign: I still don't quite get the difference between errors and unsavedErrors

lib/utils/errorHandler.js
```diff
@@ -0,0 +1,99 @@
...
+  if(!chalkType) {
+    chalkType = chalk.bold.red;
+  }
+  errArr.forEach(function(err) {
+    console.error(chalkType(err.message || err));
+  });
+};
+
+var ErrorHandler = {
+  show: function (errorGroup, chalkType) {
```
dbanksdesign: This method isn't being used.

lib/utils/errorHandler.js
```diff
@@ -0,0 +1,99 @@
...
+        unsavedErrors[errorGroup] = [];
+      }
+      unsavedErrors[errorGroup].push(error);
+    }
+    else {
+      errors.push(error);
+    }
+  },
+
+  done: function (errorGroup) {
```
dbanksdesign: This method isn't being used.

lib/utils/resolveObject.js
```diff
@@ -62,36 +66,29 @@ function traverseObj(obj) {
         traverseObj( obj[key] );
       } else {
         if (typeof obj[key] === 'string' && obj[key].indexOf('{') > -1) {
-          try {
-            obj[key] = compile_value( obj[key], [key] );
-          }
-          catch (e) {
-            errors.push(e);
-          }
+          obj[key] = compile_value( obj[key], [current_context.join('.')] );
```
dbanksdesign: Why is the second argument changing from `[key]` to `[current_context.join('.')]`

lib/utils/resolveObject.js
```diff
@@ -62,36 +66,29 @@ function traverseObj(obj) {
         traverseObj( obj[key] );
       } else {
         if (typeof obj[key] === 'string' && obj[key].indexOf('{') > -1) {
-          try {
-            obj[key] = compile_value( obj[key], [key] );
-          }
-          catch (e) {
-            errors.push(e);
```
dbanksdesign: Because we are removing this, we should remove the instantiation of the errors variable above.

lib/utils/resolveObject.js
```diff
@@ -62,36 +66,29 @@ function traverseObj(obj) {
...
-  stack.push(value.slice(1, -1));
 
   // Replace the reference inline, but don't replace the whole string because
   // references can be part of the value such as "1px solid {color.border.light}"
   value.replace(regex, function(match, variable) {
+    stack.push(variable);
+
     // Find what the value is referencing
-    ref = selfRef(variable.trim(), updated_object);
+    ref = selfRef(variable.trim(), updated_object, stack);
```
dbanksdesign: Why are we passing stack here, the method does not seem to use this argument.

lib/utils/resolveObject.js
```diff
@@ -100,9 +97,24 @@ function compile_value(value, stack) {
...
-          if(stack.indexOf(reference)!==-1) {
-            stack.push(reference);
-            throw new Error('Variable reference for \'' + stack[0] + '\' resolves to circular definition cycle of \'' + reference + '\':  ' + stack.join(', ') );
+
+          // Compare to found circular references
+          var foundCircArr = Object.keys(foundCirc);
+          if (foundCircArr.indexOf(reference)!==-1) {
+            // If the current reference is a member of a circular reference, do nothing
+          }
+          else if(stack.indexOf(reference)!==-1) {
```
dbanksdesign: It's weird to have an empty body to part of a conditional, this could be rewritten as

```js
if (stack.indexOf(reference) !== -1) {
} else if (foundCircArr.indexOf(reference) === -1) {
}
```

lib/utils/resolveObject.js
```diff
@@ -100,9 +97,24 @@ function compile_value(value, stack) {
         // Recursive so we can compute multi-layer variables like a = b, b = c, so a = c
         if ( to_ret.indexOf('{') > -1 ) {
           var reference = to_ret.slice(1, -1);
-          if(stack.indexOf(reference)!==-1) {
-            stack.push(reference);
-            throw new Error('Variable reference for \'' + stack[0] + '\' resolves to circular definition cycle of \'' + reference + '\':  ' + stack.join(', ') );
+
+          // Compare to found circular references
+          var foundCircArr = Object.keys(foundCirc);
+          if (foundCircArr.indexOf(reference)!==-1) {
```
dbanksdesign: We don't need to create an array of object keys, we could just do `foundCircArr.hasOwnProperty(reference)`

lib/utils/resolveObject.js
```diff
@@ -100,9 +97,24 @@ function compile_value(value, stack) {
...
+            // If the current reference is a member of a circular reference, do nothing
+          }
+          else if(stack.indexOf(reference)!==-1) {
+            // If the current stack already contains the current reference, we found a new circular reference
+            // chop down to just the circular part, save it to our circular reference info, and spit out an error
+            var stackIndexReference = stack.indexOf(reference);
+            var circStack = stack.slice(stackIndexReference);
+            circStack.forEach(function(key) { foundCirc[key] = true; });
+            circStack.push(reference);
```
dbanksdesign: TBH I don't really know what is going on here.

lib/utils/resolveObject.js
```diff
@@ -124,7 +137,7 @@ function compile_value(value, stack) {
 }
 
 
-function selfRef(string, obj) {
+function selfRef(string, obj, stack) {
```
dbanksdesign: `stack` isn't being used in this method it looks like.

lib/extend.js
chazzmoney: Good catch, thanks!

lib/utils/errorHandler.js
chazzmoney: Hmmm.... yes. Will delete I guess.

lib/utils/resolveObject.js
chazzmoney: Yeah... I was thinking about removing current_context as a private external variable so that it was passed in using stack instead. What do you think is right here?

lib/utils/resolveObject.js
chazzmoney: Yeah. We may want to do something in the future. I think we had something there before but I removed it. I like having the comment in there to be clear that we are intentionally not doing something there.

Alternatively, I could refactor the logic. Let me know if you think that would be better.

lib/utils/resolveObject.js
chazzmoney: Good call. Was using it for console output but obviously thats all gone now.

lib/utils/resolveObject.js
chazzmoney: Adding comments to the code to explain. Essentially creating documentation of the circular reference cycle that we found. Then using that to look for circular references in the future and to output that info to the error system.

lib/utils/resolveObject.js
chazzmoney: Yeah... I was thinking about removing current_context as a private external variable so that it was passed in using stack instead. What do you think is right here?

lib/utils/errorHandler.js
chazzmoney: From the grouping POV, unsavedErrors are still being grouped and errors have already been grouped.

There is a bit more to it than that, e.g. handling errors that don't need grouping but you want to wait to output. But its just a generalization of the above case where group size is 1.

chazzmoney:
@didoo Let me know if this is better at all.

lib/utils/resolveObject.js
chazzmoney: This is part of the bugfix. [key] was something like ['base'] instead of ['font.size.base']. In the compile_value code, this array is the traced stack of replaced values. if something later references 'font.size.base', it would not be recognized as a circular reference and fail. This behavior is exhibited in #118.

chazzmoney:
Ok @didoo and @dbanksdesign I think we are almost there. I spent some time thinking and decided that this isn't an error handler, it is an error grouper.

Based on the idea that grouping is useful only for specific situations, I reworked the naming schemes and usage. It should make more sense now, especially in terms of specificity of use.

Let me know your thoughts; would like to get this merged prior to #152 

Thanks!

lib/utils/groupErrors.js
```diff
@@ -0,0 +1,83 @@
...
+};
+
+var GroupErrors = {
+  GROUP: {
+    PropertyReferenceErrors: 'Property Reference Errors',
+    PropertyValueCollisions: 'Property Value Collisions',
+  },
+
+  flush: function (errorGroup, chalkType) {
+    if(errorGroup) {
```
didoo: I don't understand why we need this test. In the code, I see no case in which it is called without the errorGroup argument. Also, it would also be strange to see a declaration like this: 

`GroupErrors.flush(,chalk.bold.red);`

(or am I missing something?)

also, see the comment above on how to possibly refactor this function

lib/utils/groupErrors.js
```diff
@@ -0,0 +1,83 @@
...
+  count: function (errorGroup) {
+    return groupedErrors[errorGroup] ? groupedErrors[errorGroup].length : 0;
+  },
+
+  fetchMessages: function (errorGroup) {
+    return (groupedErrors[errorGroup] || []).map(function (err) { return err.message});
+  },
+
+  clear: function (errorGroup) {
+    if (errorGroup) {
```
didoo: maybe here is better:
```
if (errorGroup && groupedErrors[errorGroup]) {
    delete groupedErrors[errorGroup];
}
```
(you used the proposed syntax above, so it's also more consistent)

lib/utils/groupErrors.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+      groupedErrors[errorGroup].push(error);
+    }
+  },
+
+  count: function (errorGroup) {
+    return groupedErrors[errorGroup] ? groupedErrors[errorGroup].length : 0;
+  },
+
+  fetchMessages: function (errorGroup) {
```
didoo: this function is used only in the tests. is there a way to declare it only in the context of the tests, and not here?

lib/utils/groupErrors.js
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+var chalk = require('chalk');
+
+var groupedErrors = {};
+
+var getFinalGroupErrors = function(errorGroup) {
```
didoo: this function is used only once. and is doing two things in one: is "getting the list of all the errors as a 'kind of' array" and "preparing the formatting of the output messages" (see when you add a double newline at the end). in the same way, `consoleOutputErrArr` is similar. I think it would be better to move these two pieces of code where they are actually used. I understand it may increase the complexity there, but is not that moving the code in another place under a function solves the problem :)
I have tried to refactor the code, removing the extra functions and simplifying the logic. see how it looks in this way:
https://gist.github.com/didoo/2b681bee9e5f1a6a4327d6278e35fa09
(if you want to use it, test it anyway, I've just refactored inside a text editor, not actually tested, ok?)

chazzmoney: Yeah, before the refactor they were used in multiple places. I thought about redoing as a single func but thought that there were two reasons to keep it. Clarity of use and that if people wanted the other functionality back it would be nice. It seems like you and Danny both are telling me to stop engineering ahead and just focus on immediate task. I'll remove the functions.

chazzmoney: Yeah, before the refactor they were used in multiple places. I thought about redoing as a single func but thought that there were two reasons to keep it. Clarity of use and that if people wanted the other functionality back it would be nice. It seems like you and Danny both are telling me to stop engineering ahead and just focus on immediate task. I'll remove the functions.

chazzmoney: Yeah, before the refactor they were used in multiple places. I thought about redoing as a single function but thought that there were two reasons to keep it. Clarity of use and that if people wanted the other functionality back it would be nice. It seems like you and Danny both are telling me to stop engineering ahead and just focus on immediate task. I'll remove the functions.

chazzmoney: Yeah, before the refactor they were used in multiple places. I thought about redoing as a single function but thought that there were two reasons to keep it. Clarity of use and that if people wanted the other functionality back it would be nice. It seems like you and Danny both are telling me to stop engineering ahead and just focus on immediate task. I'll remove the functions.

chazzmoney: Yeah, before the refactor they were used in multiple places. I thought about redoing as a single function but thought that there were two reasons to keep it. Clarity of use and that if people wanted the other functionality back it would be nice. It seems like you and Danny both are telling me to stop engineering ahead and just focus on immediate task. I'll remove the functions.

chazzmoney:
@didoo Let me know what you think of latest changes. I need to make a couple more minor updates, but I think you'll like it better.

chazzmoney:
@didoo Sorry for the delay, lots of isssues with github. (looks like travis-ci isn't running yet either).

This is the final version of the updates that I had mentioned.

__tests__/__json_files/not_circular.json
```diff
@@ -0,0 +1,8 @@
+{
+  "prop1" : { "value": "test1 value" },
+  "prop2" : { "value": "test2 value" },
+  "prop3" : { "value": "{prop1.value}" },
+  "prop4" : { "value": "{prop3.value}" },
+  "prop12" : { "value": "{prop1.value}, {prop2.value} and some extra stuff" },
+  "prop124" : { "value": "{prop1.value}, {prop2.value} and {prop4.value}" }
```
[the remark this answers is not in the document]
chazzmoney: Thanks for the examples!

didoo:
@chazzmoney I think is much better and clearer now, no? to me looks good :)

lib/utils/resolveObject.js
chazzmoney: My earlier finding that this was generating errors is no longer replicable because it was a interaction with the old compile_value code.

In its current form testing and code are both correct as written in the PR. I'll try to explain. Lets look at __tests__/json_files/circular_2.json

With [key], the compile_value function is called in the following order with the following arguments:
```
  {j} [ 'c' ]
  {a.b.c} [ 'c', 'j' ]
  {a.b.c} [ 'j' ]
```

With [current_context.join('.')], the compile_value function is called in the following order with the following arguments:
```
  {j} [ 'a.b.c' ]
  {a.b.c} [ 'j' ]
```

What does this mean?

In the first situation, 'c' is found to reference 'j' which is found to resolve to 'a.b.c', which is then found to resolve to 'j', which is a circular reference. It does catch the circular reference, but it was wrong in that it found c -> j -> a.b.c -> j. 'c' isn't actually a property.

In the second situation, 'a.b.c' is found to reference 'j' which is found to resolve to 'a.b.c', which is a circular reference. This is what we should expect.

IMO, the tests failing with [key] is actually a good thing because it ensures that we capture circular references in correct order. I don't see a way to directly write tests to find and catch that the first value in the stack is being incorrectly written. Let me know if you have any suggestions.

chazzmoney: I am pushing on more change to the tests to put each circular reference test in its own it('should ...') wrapper. Turns out Jest stops executing any block inside an it() after it fails an expect, which is maybe not what we want.

Issue #160, opened 2018-10-05
Add JSON5 property file support


chazzmoney:
Pull in properties from files written in JSON5.

vvscode:
I'm taking it

chazzmoney:
The actual import of the file data is done for both JS and JSON here:

https://github.com/amzn/style-dictionary/blob/master/lib/utils/combineJSON.js#L43

You may need to parse through things a bit to see how it all ties in. Let us know if you have questions.

chazzmoney:
Oh, and you should work against the develop branch.

vvscode:
https://github.com/amzn/style-dictionary/pull/165

PR #165, opened 2018-10-11
Add JSON5 property file support

vvscode:
https://github.com/amzn/style-dictionary/issues/160

### DIFF
--- a/lib/extend.js
+++ b/lib/extend.js
@@ -11,6 +11,8 @@
  * and limitations under the License.
  */
 
+require('json5/lib/register');
+
 var combineJSON = require('./utils/combineJSON'),
     deepExtend = require('./utils/deepExtend'),
     resolveCwd = require('resolve-cwd'),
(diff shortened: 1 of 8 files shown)


chazzmoney:
LGTM - would like to wait for @dbanksdesign before merging though.

lib/extend.js
```diff
@@ -11,6 +11,8 @@
  * and limitations under the License.
  */
 
+require('json5/lib/register');
```
dbanksdesign: Dang. That is all we need to do?!?

vvscode: https://www.npmjs.com/package/json5#nodejs-require-json5-files

> Node.js require() JSON5 files
> When using Node.js, you can require() JSON5 files by adding the following statement.

`require('json5/lib/register')`
> Then you can load a JSON5 file with a Node.js require() statement. For example:

`const config = require('./config.json5')`

Issue #139, opened 2018-10-01
Add 'json/nested' format


dbanksdesign:
From the discussion on #58 , we want to add different types of JSON formatting for different purposes. This format should output a JSON string that maintains the structure of the style dictionary JSON, but only includes the values and nothing else. For example, this style dictionary JSON:
```json
{
  "size": {
    "font": {
      "small": { "value": "13px" },
      "medium": { "value": "15px" },
      "large": { "value": "17px" },
      "base": { "value": "{size.font.medium.value}" }
    }
  }
}
```

Would output this in the 'json/flat' formatter:
```json
{
  "size": {
    "font": {
      "small": "13px",
      "medium": "15px",
      "large": "17px",
      "base": "15px"
    }
  }
}
```
You can have a look at how the other formatters are written to see how to write a formatter and where it should go.

Make sure to include proper jsdoc comments and unit test!

davixyz:
Added this format but i'm not sure, to which branch do I issue the PR? I issued it to develop for now but I can change it if needed

PR #167, opened 2018-10-14
feat(json-nested): Add JSON nested transform

davixyz:
*Issue #, if available:*

re #139

*Description of changes:*
Added JSON nested transform, Added test for it and Documentation update

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -52,6 +52,20 @@ function iconsWithPrefix(prefix, properties, config) {
     })
     .value().join('\n');
 }
+
+function minifyDictionary(obj) {
+  var toRet = {};
+  if (obj.hasOwnProperty('value')) {
+    return obj.value;
+  } else {
+    for(var name in obj) {
+      if(obj.hasOwnProperty(name)) {
+        toRet[name] = minifyDictionary(obj[name]);
+      }
+    }
+  }
+  return toRet;
+}
 /**
  * @namespace Formats
  */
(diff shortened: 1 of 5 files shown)


lib/common/formats.js
```diff
@@ -52,6 +52,24 @@ function iconsWithPrefix(prefix, properties, config) {
     })
     .value().join('\n');
 }
+
+function filterValueProperty(properties, path, obj ) {
+  path = path || [];
```
dbanksdesign: I believe there is a way to do this without passing keeping a path around...

I think @chazzmoney wrote this in our demo repo, let me know if this makes sense:
```js
const minifyDictionary = function(obj) {
  let toRet = {};
  if (_.has(obj, 'value')) {
    return obj.value;
  } else {
    for(var name in obj) {
      toRet[name] = minifyDictionary(obj[name]);
    }
  }
  return toRet;
};
```

davixyz: oh nice, let me take a look into this function

chazzmoney: I wrote that? Seems like something I'd do. You might need a if(obj.hasOwnProperty(name)) wrapper on that "toRet[name] = " assignment statement.

lib/common/formats.js
```diff
@@ -52,6 +52,20 @@ function iconsWithPrefix(prefix, properties, config) {
     })
     .value().join('\n');
 }
+
+function minifyDictionary(obj) {
```
davixyz: @chazzmoney definitely a more cleaner approach

lib/common/formats.js
```diff
@@ -52,6 +52,20 @@ function iconsWithPrefix(prefix, properties, config) {
...
+  var toRet = {};
+  if (_.has(obj, 'value')) {
```
dbanksdesign: Can this just be `obj.hasOwnProperty('value')` ? Just because we are using that below and trying to use lodash as little as possible in favor plain JS. :)

chazzmoney: Derp. I should have caught that in my earlier comment. Sorry @davixyz

davixyz: Oh dang! you are right it's better

chazzmoney:
LGTM!

Issue #177, opened 2018-10-21
Navigation for the website should be responsive


didoo:
As reported in #175 the website navigation is not working on small viewports. We need to figure out what is happening (there are some overrides of the _vue.css_, I think, which work only for large viewports).

dbanksdesign:
I opened a PR to fix this, if someone wants to take a look I can merge it today to get this fixed!

didoo:
@dbanksdesign apart from a couple of comments, code looks ok to me.

PR #179, opened 2018-10-23
Fix: #177

dbanksdesign:
*Issue #, if available:* #177

*Description of changes:* Fixing the mobile nav bugs in our doc site I caused by messing with the CSS in docsify. The navigation in mobile should work now!

### DIFF
--- a/docs/assets/styles.css
+++ b/docs/assets/styles.css
@@ -53,11 +53,11 @@ aside.sidebar {
   border: none;
   background-color: var(--theme-color-secondary);
   color: #fff;
-  width: 20%;
+  /* width: 20%; */
 }
 
 body.close .sidebar {
-  transform: translateX(-100%);
+  /* transform: translateX(-100%); */
 }
 
 .sidebar ul li a {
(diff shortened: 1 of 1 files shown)


docs/assets/styles.css
```diff
@@ -386,6 +385,20 @@ section.cover .cover-main {
     max-width: 100px;
     max-height: 100px;
   }
+
+  body .sidebar-toggle { background-color: #fff; }
+  body.close .sidebar-toggle {
+    background-color: transparent;
+    padding: 1rem;
```
didoo: are you sure you need this line? I see the padding declared also in `body .sidebar-toggle {}` and nothing overrides it (or maybe I am missing something).

docs/assets/styles.css
```diff
@@ -386,6 +385,20 @@ section.cover .cover-main {
     max-width: 100px;
     max-height: 100px;
   }
+
+  body .sidebar-toggle { background-color: #fff; }
+  body.close .sidebar-toggle {
+    background-color: transparent;
```
didoo: maybe you can move this declaration directly under `body .sidebar-toggle {}`? (currently has background transparent and you are overriding it.

docs/assets/styles.css
dbanksdesign: I need it because these styles are overriding another 3rd party CSS file :/

docs/assets/styles.css
```diff
@@ -53,11 +53,11 @@ aside.sidebar {
   border: none;
   background-color: var(--theme-color-secondary);
   color: #fff;
-  width: 20%;
+  /* width: 20%; */
 }
 
 body.close .sidebar {
-  transform: translateX(-100%);
+  /* transform: translateX(-100%); */
```
dbanksdesign: I need to just remove these comments

docs/assets/styles.css
dbanksdesign: Same, it's because I'm overriding a 3rd party CSS file

chazzmoney:
Ah! LGTM!

Issue #195, opened 2018-11-08
CSS Comments


blackfalcon:
I noticed an issue with Style Dictionary when using comments to Sass and CSS. In Sass we get 

```
$color-type-theme-classic-light-link: #2774ae; // deprecated classic token color
```

In CSS we also get 

```
:root {
 --color-type-theme-classic-light-link: #2774ae; // deprecated classic token color
 --color-type-theme-classic-light-link-hover: #2774ae; // deprecated classic token color
}
```

This is not valid CSS and when using other post processors I get the following error

```
(node:85291) UnhandledPromiseRejectionWarning: CssSyntaxError: /Users/.../tokens/example/tokensBuild/TokenVariables.css:8:56: Unknown word
```

Also, this ONLY happens when there is an illegal comment as the last line. For example, when this is the variable set, there is no error. 

```
:root {
 --color-type-theme-classic-light-link: #2774ae; // deprecated classic token color
 --color-type-theme-classic-light-link-hover: #2774ae; // deprecated classic token color
 --color-type-disabled: #c8c9c7;
}
```

dbanksdesign:
Interesting, you are correct I guess we have always been doing CSS comments wrong. Opened a PR to fix it: #196

chazzmoney:
Fix in develop, will be in the next release

PR #196, opened 2018-11-09
Fix: #195

dbanksdesign:
*Issue #, if available:* #195

*Description of changes:* Fixing CSS variables templates because we were using invalid comments. Switching all web-based comments to block style so they work anywhere (less, scss, css, js). Updating snapshots to include a comment in the property. 


chazzmoney:
LGTM

Issue #187, opened 2018-11-01
Feature Request: Interactive Scaffolding


dbanksdesign:
## Customer Problem
Getting set up on Style Dictionary takes time and is not always very straightforward. To lower the barrier of entry we can make an interactive scaffolding CLI that sets up a Style Dictionary package suited to the needs of the user (at least getting them started). Just like when you run `npm init` it asks you questions about your project, we could do the same.

### Some ideas for configuration
* Platforms to target/types of output
* Structure of design tokens/design token files
* Themes?
* Starting point (bootstrap, foundation, material, scratch, import from css, etc.)

vvscode:
@dbanksdesign  could you provide examples for mentioned options? I mean I'm ready to take tool development, but who will provide structures ?

bolora:
Perhaps this kind CLI interaction?

> style-dictionary create acme-corp
> Where would you like to begin?
   - Start from scratch
   - From existing css
   - Bootstrap
   - Foundation
   - Semantic UI
   - Material

bolora:
@vvscode For example, in Bootstrap it is a matter of expressing a SD json file structure that delivers this file: https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss

chazzmoney:
I think that it makes sense to ask the user if they are starting from scratch or an existing project first. If they are starting from scratch there is a semi-straightforward method (as per @bolora).

If they are starting from an existing project, we probably need them to point us to platform output one by one. The first time we probably have to understand the structure they want to use for the SD, but after that each platform is similar: detect platform output type, add platform info to SD config, find existing styles in platform output and replace with SD variables, import styles into SD. May also need a tool to help dedupe styles (or near styles, like similar colors or sizes)).

There is a lot to parse here for sure, and the existing project being the more complicated but also probably more useful.

dbanksdesign:
Found this js library we could potentially use for the CLI interaction: https://github.com/enquirer/enquirer

didoo:
@dbanksdesign I strongly suggest https://github.com/SBoudrias/Inquirer.js to build a CLI prompt: I have used it for a long time and is perfect.

chazzmoney:
@vvscode Ahhhh, got it! Thanks. :-)

Not sure we are there yet... Maybe for the starting from scratch?

davixyz:
If we haven't got into this; We could also look into:
https://plopjs.com/

[...]

Issue #120, opened 2018-09-04
Transform after replace references


kwivix:
As a developer I want to execute certain transforms after the references are replaced. In my case, perform certain actions on color's such as darken or lighten.

**The why**
I want to theme the design tokens, by having colors a few base colors and calculate the shades from them.

kwivix:
I will keep an eye open for the PR. There is nothing wrong with first performing the transforms and then the references. Intact it makes the code a lot easier. Resolving references (recursive) 'on the go' while performing transforms will be a complex task.

chazzmoney:
@kwivix Can you mention a little more about how you are thinking about using this?

The first part of your first message made me think you wanted to customize colors for specific output - say one color for iOS and another color for web/css.

The second part of that message made me think that you wanted to have a few starting colors and then modify them automatically to be in multiple shades. Something maybe like https://palx.jxnblk.com/00a

It might be possible to have two transforms happen at different times - one on initial token values and another immediately before output via format/template. I want to make sure that we are taking your use case into consideration.

kwivix:
My use case concerns the second suggestion: I would like to start with three base colors that define the brand. From the brand colors I want to calculate the different shades and use the shades in the design tokens, such as tokens for button/heading/accent colors.

1. Define base brand colors
1. Calculate shades of the base colors
1. Use shades in design tokens

By changing the initial three base colors its possible to generate a new brand (or theme).

In the end there will be one design system for multiple brands. Each brand will have it's own colors. The brand colors are brand specific and won't be shared between brands.

chazzmoney:
Thanks for the info, very helpful!

To clarify one more thing based on this use case:

> As a developer I want to execute certain transforms after the references are replaced.

You specifically mention executing transforms after the references are replaced in your first message. For the three step process you listed out in the most recent comment, the main requirement appears to be the ability to perform generative transformations. Was there a specific importance to doing this after the references are replaced?

kwivix:
My first thought was to transform the values after resolving. After giving it some more thought I figured out a more generic solution is needed. It doesn't make sense to have two steps: Define base colors, calculate shades in design tokens. With the two step 'solution' a transformation has to be defined on every token or the application itself needs to use the shades directly. The later one defeats the purpose of having design tokens.

**Example with two steps:**
```json
{
    "color": {
        "brand": {
            "primary": {
                "value": "#ff0000"
            }
        },
        "background": {
            "button": {
                "primary": {
                    "base": {
[...]

PR #198, opened 2018-11-17
New cut at documentation PR using current develop branch

chazzmoney:
This is intended to replace #103. It is not intended to be a perfect or complete documentation update. The spirit is to get this in as quickly as possible and then add TODOs for any other documentation updates we want to do.

@didoo and @dbanksdesign Please give this a look over. It is based off of the existing setup in the develop branch, avoiding all the conflicts and fixing up some of the TODOs in the other branch.

I'd like to get this done and pulled before Wednesday next week.

I'll add issues for any comments I can't handle directly in this PR and tackle them small and separately from there.


docs/config.md
```diff
@@ -0,0 +1,49 @@
...
+| platform.transforms | Array[String] (optional) | An array of [transforms](transforms.md) to be performed on the style properties object. These will transform the properties in a non-destructive way so each platform can transform the properties. Transforms to apply sequentially to all properties. Can be a built-in one or you can create your own. |
+| platform.transformGroup | String (optional) | A string that maps to an array of transforms. This makes it easier to reference transforms by grouping them together. You must either define this or `transforms`. |
+| platform.buildPath | String (optional) | Base path to build the files, must end with a trailing slash. |
+| platform.files | Array (optional) | Files to be generated for this platform. |
+| platform.file.destination | String (optional) | Location to build the file, will be appended to the buildPath. |
```
didoo: while moving the text around you missed these:

```

 | platform.file.filter | Function/Object (optional) | A function or object used to filter the properties that will be included in the file. If a function is provided, each property will be passed to the function and the result (true or false) will determine whether the property is included. If an object is provided, each property will be matched against the object using a partial deep comparison to determine whether the property is included. |
 | platform.file.options | Object (optional) | A set of extra options associated with the file. |
 | platform.file.options.showFileHeader | Boolean | If the generated file should have a "Do not edit + Timestamp" header (where the format supports it). By default is "true". |
```
can you double-check that everything has been copy&pasted correctly and nothing else has been missed?

docs/package_structure.md
```diff
@@ -18,123 +18,12 @@ Here is a basic example of what a style dictionary package looks like.
...
-  }
-}
-```
-
-The comment  will appear in the output files, where relevant or the output format supports comments.
+| Name | Description |
+| :--- | :--- |
+| config.json | This is where the [configuration](config.md) for the style dictionary lives, where you define what happens when Style Dictionary runs |
+| properties | [Properties](properties.md) are saved as a collection of JSON or JS module files. We usually keep them in a `properties` directory, but you can put them wherever you like - the path to them should be in the `source` attribute on your `config.json` file. |
+| assets (optional) | Assets can be included in your style dictionary package, allowing you can keep them in your style dictionary as a single source of truth. |
```
didoo:
```suggestion
| assets (optional) | Assets can be included in your style dictionary package, allowing you to keep them in your style dictionary as a single source of truth. |
```

docs/transform_groups.md
```diff
@@ -1,6 +1,6 @@
 # Transform Groups
 
-Transform Groups are a way to easily define and use groups of transforms. They are an array of transforms. You can define custom transform groups with the [`registerTransformGroup`](api.md#registertransformgroup).
+Transform Groups are a way to easily use multiple transforms at once. They are an array of transforms. You can define a custom transform group with the [`registerTransformGroup`](api.md#registertransformgroup).
```
didoo:
```suggestion
Transform Groups are a way to easily use multiple transforms at once. They are an array of transforms. You can define a custom transform group with the [`registerTransformGroup`](api.md#registertransformgroup) method.
```

docs/using_the_cli.md
```diff
@@ -0,0 +1,73 @@
...
+| Configuration Path | -c <path>, --config <path> | Set the path to the configuration file. Defaults to './config.json'. |
+| Platform | -p <platform>, --platform <platform> | Only clean a specific platform. If not supplied, cleans all platform found in the configuration file. |
+
+
+# init
+Generates a starter style dictionary, based on the supplied example type. Usage:
+```bash
+$ style-dictionary init <example-type>
+```
+Available examples can be found [here](https://github.com/amzn/style-dictionary/tree/master/examples/).
```
didoo: Not true. You can install only a limited list of examples. See here: https://github.com/amzn/style-dictionary/blob/eba4970065941aab4403b4f544a805ae3846fcd9/bin/style-dictionary#L41

This will be even more true with the release of the new examples:
https://github.com/didoo/style-dictionary/blob/cba2672bcc31788b1d07b7d79ce7608b92e6b1b8/bin/style-dictionary#L41

docs/architecture.md
```diff
@@ -0,0 +1,13 @@
+# Architecture Overview
+
+This is how Style Dictionary works under the hood.
+
+## TODO
+
+1. We need to replace the picture with one that is more representative of the process.
```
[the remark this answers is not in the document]
chazzmoney: @dbanksdesign I agree with Didoo's view - it needs to be simple to understand what is happening.

docs/assets/properties.md
```diff
@@ -0,0 +1,129 @@
+# Properties
```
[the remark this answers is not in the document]
chazzmoney: whoaaaa.... how did that happen. Good catch!

docs/assets/properties.md
```diff
@@ -0,0 +1,129 @@
...
+
+> Synonyms: design token, design variable, design constant, atom
+
+Style properties are stored a collection of JSON or JS module files. We usually keep them in a `properties` directory, but you can put them wherever you like,
+they just need to be referenced in the `source` attribute on your `config.json` file.
+
+A property is collection of attributes that describe any fundamental/atomic visual style. Each attribute is a key:value pair. A property name and its value are considered a design token (or design variable, design variable, design constant, design atom).
+
+![](assets/property-definitions.png)
```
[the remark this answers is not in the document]
chazzmoney: Added

docs/assets/properties.md
```diff
@@ -0,0 +1,129 @@
...
+      "medium": { "value": "16" },
+      "large" : { "value": "24" },
+      "base"  : { "value": "{size.font.medium.value}" }
+    }
+  }
+}
+```
+
+# Property reference / alias
+You can reference (alias) any part of the Style Dictionary the same way
```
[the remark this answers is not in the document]
chazzmoney: @dbanksdesign whoa - it doesn't? This is a bug right?

docs/build_process.md
```diff
@@ -4,39 +4,10 @@ Here is what the build system is doing under the hood.
...
-
-// This will perform step 3 above, for each platform:
-// 1. Apply transforms
-// 2. Build files
-// 3. Perform actions
-styleDictionary.buildAllPlatforms();
-```
+  1. Performs all transforms, in order, defined in the transforms attribute or transformGroup
+  1. Builds all files defined in the files array
+  1. Performs any actions defined in the actions attribute
```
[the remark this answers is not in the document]
chazzmoney: I deleted that / broke it into two - using the CLI and using the npm module. You think that works ok?

docs/config.md
chazzmoney: Not sure how to double check. I tried to be very careful when copying over before. Any suggestions on how best to do this?

docs/using_the_cli.md
chazzmoney: Urg...

Do we really want to not allow people to use a different starting example? I can change this, but I think it might make more sense to remove the code you are pointing to and have it do a check to see if the directory exists?

The types list could just be the directory list in the examples folder.

@dbanksdesign What do you think?

docs/build_process.md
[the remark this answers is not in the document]
didoo: It works ok :)

docs/config.md
didoo: No, just have a second look.

docs/using_the_cli.md
didoo: The reasoning about this is that, if I am a user that needs to use the `init` command, probably it's the first time they I Style Dictionary. Giving me too many options can be counter-productive. As a first time user, probably what I want is to see/have some very basic setup, hence the "basic", or maybe I feel bolder and I try the "complete" to see how it looks like. If I have already started to use a little SD, probably I have read some documentation, maybe I had a look in the repo (and saw there is an "example" folder) and so on. So making the "init" a simple choice is better for me. 

Think about the "npm" example. If I really need to distribute my tokens as npm module (which means, I am using SD in a quite advanced way) the example will never be what I need. Is a good starting point, but the final result will be quite different. So, if you think to the two use cases (newbie and experienced) the "init" of an npm example can result confusing in the first case, and insufficient in the second case.

No strong opinions on this, just a gut feeling that in this case "less is better" :)

didoo: For example, I have just discovered (and updated the example) that the "prepublish" script is deprecated, now is better to use "prepublishOnly". Our example was not "perfect" to be used as a real-world base for an npm module distribution.

docs/config.md
chazzmoney: I double checked this one, didn't miss anything else.

docs/using_the_cli.md
chazzmoney: I see what you are saying. I guess my feelings are that the same argument could be applied to having examples at all - maybe we should only have the basic ones then. Or we should make sure we keep them up-to-date.

Maybe it is the command "init" causing some of the problem here. Really we mean "example".

dbanksdesign: I think we should think about 'starter' projects as different than 'examples'. Maybe in the documentation we can say you can get started with the types of init projects with the CLI command, or take a look at all the examples. I think because we are breaking out our examples into single concepts like base64ing assets, npm module, etc. those are better served as reference/example code rather than a starter project. You would want to add that functionality after you get your Style Dictionary project up and running. Long-term, I think a scaffolding CLI should replace `init` anyways to give better starter projects than just copying a simple directory. Thoughts?

chazzmoney: > Long-term, I think a scaffolding CLI should replace init anyways to give better starter projects than just copying a simple directory. Thoughts?

This was my exact thinking and means that we should change this to a different command besides `init`. We use `example` for this and then build a simple `init` command as per #187.

Also, I'm confused about us thinking that we should only let people generate and use the `basic` examples. Examples exist to play with things and see how they work. Maybe I've used Style dictionary for 2 weeks and I understand the basics but now want to learn how base64ing assets work without risking messing up my current project - being able to generate and build an example gives me the playground I need to learn how to use it.

docs/properties.md
```diff
@@ -0,0 +1,127 @@
+# Properties
+
+> Synonyms: design tokens, design variables, design constants, atoms
+
+Style properties are stored in a collection of JSON or JS module files. We usually keep them in a `properties` directory, but you can put them wherever you like, they just need to be referenced in the `source` attribute on your `config.json` file.
+
+A property is a collection of attributes that describe any fundamental/atomic visual style. Each attribute is a `key:value` pair. A property name and its value are considered a design token (or design variable/constant/atom).
+
+![Terminology for different parts of a JSON property](assets/property-definitions.png)
```
dbanksdesign: Does this file exist? I don't see it in the PR.

docs/properties.md
```diff
@@ -0,0 +1,127 @@
...
+      "tertiary": { "value": "#666666" },
+      "inverse": {
+        "base": { "value": "#ffffff" }
+      }
+    }
+  }
+}
+```
+
+Any object in the JSON that has a `value` attribute on it is a property, so in this example there are 4 style properties: `color.font.base`, `color.font.secondary`, `color.font.tertiary`, and `color.font.inverse.base`.
```
dbanksdesign: I like this explanation.

docs/properties.md
```diff
@@ -0,0 +1,127 @@
...
+
+For any properties you wish to output, the "value" attribute is required. This provides the data that will be used throughout the build process (and ultimately used for styling in your deliverables). You can optionally include any custom attributes you would like (e.g. "comment" with a string or "metadata" as an object with its own attributes).
+
+# Example Property
```
dbanksdesign: Should we use different headings for this and the ones below? These would be H1s

docs/properties.md
```diff
@@ -0,0 +1,127 @@
...
+        "value": "10",
+        "comment": "the smallest font allowed for readability"
+      },
+    }
+  }
+}
+```
+
+# Multiple Properties "size.font.\*"
+Multiple properties in a single file are simple to read and understand using the [`Category / Type / Item (CTI)`](#category-type-item-(cti)) method
```
dbanksdesign: Maybe add the caveat here that this structure is not required, you can structure your properties however you want?

docs/properties.md
```diff
@@ -0,0 +1,127 @@
...
+      "small" : { "value": "10" },
+      "medium": { "value": "16" },
+      "large" : { "value": "24" },
+    }
+  }
+}
+```
+
+# Attribute reference / alias
+You can reference (alias) existing attributes by using the property name in brackets
```
dbanksdesign: Maybe change "property name" to something about dot-notation object path?

docs/quick_start.md
```diff
@@ -195,7 +186,7 @@ const StyleDictionary = require('style-dictionary').extend('config.json');
 StyleDictionary.buildAllPlatforms();
 ```
 
-The `.extend()` method is an overloaded method that can also take an object with the configuration in the same format as a config.json file.
+The `.extend()` method is an overloaded method that can also take a [configuration](config) object.
```
dbanksdesign: Missing ".md"

docs/using_the_cli.md
dbanksdesign: Have an example command like: `style-dictionary example assets-base64-embed` and then turn `style-dictionary init` into the scaffolding script? I like that. Maybe let's wait until we get the scaffolding to make the change in documentation?

docs/quick_start.md
chazzmoney: weirdly it works without the md. I was putting them back in for clarity / consistency.

docs/properties.md
chazzmoney: done

chazzmoney:
@dbanksdesign , found this in the documentation:

> Assets
>Assets are not required, but can be useful to include in your style dictionary. If you don't want to manage having assets like images, vectors, font files, etc. in multiple locations, you can keep them in your style dictionary as a single source of truth.
>Coming soon: how to generate image assets based on your style dictionary

Is there something to add here or should I leave as is?

docs/properties.md
chazzmoney: Called it 'recommended'

docs/properties.md
chazzmoney: fixed

docs/properties.md
chazzmoney: Fixed

docs/README.md
```diff
@@ -8,102 +8,37 @@
...
-
-In this example there are 4 style properties: `color.font.base`, `color.font.secondary`, `color.font.tertiary`, and `color.font.inverse.base`.
-A style property is any object in the JSON that has a `value` attribute on it. In this way you can nest properties at different levels.
-This allows you to easily access the property as well as do things like get all the inverse font colors.
+1. _Literally anything you want styles or style data in_
+
+The value of using Style Dictionary to build all of these is that they are all consistent and up to date.
+
+The Style Dictionary framework is fully extensible and modular so you can create any type of file from a style dictionary.
+If there is a new language, platform, or file type you need, you can extend the style dictionary framework to create the necessary files.
```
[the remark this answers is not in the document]
chazzmoney: fixed

bin/style-dictionary
```diff
@@ -29,7 +29,7 @@ program
 
 program
   .command('clean')
-  .description('Removes files specificed in the config of the style dictionary package of the current directory.')
+  .description('Removes files specified in the config of the style dictionary package of the current directory.')
```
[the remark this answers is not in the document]
chazzmoney: tsk tsk

docs/using_the_cli.md
chazzmoney: Sounds good. I set up the documentation to reflect the current state.

docs/architecture.md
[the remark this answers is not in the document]
chazzmoney: This is now updated with Danny's PR.

chazzmoney:
@didoo @dbanksdesign I think we are good here, unless you have any final comments?

chazzmoney:
@didoo @dbanksdesign Am planning on squash and merging on Monday or Tuesday - let me know if you have any other thoughts or concerns.

docs/properties.md
didoo: That one was a quick mock. Let me know if you want me to prepare a better version.

didoo:
@chazzmoney I had a new look and seems ok to me. Only thing, I am still not 100% sure if this can generate confusion if we don't specify that it works **only** if we are targeting a "value":
https://github.com/amzn/style-dictionary/pull/198/files#diff-202d3626d667c87a265182819faab5e0R82

chazzmoney:
Yeah, thats a bug. Tested and it creates a duplicate name. Let me see if I can find the issue and I'll submit a PR for it.

docs/architecture.md
```diff
@@ -0,0 +1,33 @@
...
+In your [config](config.md) file you define a `source`, which is an array of file paths. This tells Style Dictionary where to find your token files. You can have them anywhere and in any folder structure as long as you tell Style Dictionary where to find them.
+
+## 2. Deep merge token files
+
+Style Dictionary takes all the files it found and performs a deep merge. This allows you to split your token files in any way you like, without worrying about accidentally overriding groups of tokens. This gives Style Dictionary a single, complete token object to work from.
+
+## 3. Split the platforms
+
+For each platform defined in your [config](config.md), Style Dictionary will do a few steps to get it ready to be consumed on that platform. Everything that happens in a platform is non-destructive so you don't need to worry about one platform affecting another.
+
```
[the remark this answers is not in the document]
chazzmoney: @dbanksdesign After reviewing the image with @didoo's comments, can you modify it to show that the iOS, android, and web are just examples?. I think maybe if you replaced them with purple bubbles  of "Start Platform Build" and then put the value of "iOS" underneath maybe?

Also maybe the final step should be "built files can then be consumed in each platform"?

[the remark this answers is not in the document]
chazzmoney: Maybe we should add an initial step for reading or processing the configuration?

chazzmoney:
Well... thats not going to happen. Because we replace references at the end, after everything is all done... there is no way to go back and reformat the names correctly (see #120). So we need to do references first / at the same time as transforms if we want this functionality to work.

chazzmoney:
Updated the documentation to correctly reflect the current state of affairs. Should we create an issue for the alias thing? Or for the redo in @dbanksdesign's first comment in #120?

I'm leaning towards the latter; we had that conversation a year ago now and seems like it makes of sense to do for a 3.0.

dbanksdesign:
Let's leave the aliasing stuff as it is for now, we can think about how to solve it in 3.0. I think the main issue is what *should* happen if you try to reference a whole object that contains values? What are the problems we are trying to solve with aliasing a whole object with values in it?

PR #200, opened 2018-11-19
Updating the architecture documentation page

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Updating the build diagram and added in-depth explanation of steps. (I did this kind of quickly so any feedback is welcome)


chazzmoney:
LGTM!

docs/architecture.md
```diff
@@ -2,12 +2,32 @@
...
+In your [config](config.md) file you define a `source`, which is an array of file paths. This tells Style Dictionary where to find your token files. You can have them anywhere and in any folder structure as long as you tell Style Dictionary where to find them.
+
+## 2. Deep merge token files
+
+Style Dictionary takes all the files it found and performs a deep merge. This allows you to split your token files in any way you like, without worrying about accidentally overriding groups of tokens. This gives Style Dictionary a single, complete token object to work from.
+
+## 3. Split the platforms
+
+For each platform defined in your [config](config.md), Style Dictionary will do a few steps to get it ready to be consumed on that platform. Everything that happens in a platform is non-destructive so you don't need to worry about one platform affecting another.
+
```
didoo: Probably I would add a comment here, to clarify that the "platforms" in the schema are just an example of the most common configuration, but that in general you can have as many platforms as you want, and for each one define specific transformations if needed. In other words, I fear that if someone looks at the documentation and at the schema (the _build-diagram.png_ image) he may think that SD works only for iOS/Android/Web and can generate only the formats shown in the picture.

dbanksdesign: I agree with both of these. I'll try to work them in this week.

PR #206, opened 2018-11-29
Final touches on build diagram and architecture

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Final touches based on comments. 


chazzmoney:
@dbanksdesign I'm happy with it, but I wanted to make sure that you intentionally left "build files" in the diagram

chazzmoney:
Not a lot of space is there.  Hmm.... “write formatted tokens into file”?

chazzmoney:
Write format to file?
Write file formats?

Also, we show the multiple token files above fanned out, then show the build files below as a stack. I realize that the above version was to show the deep merge, but this could cause confusion maybe.

@didoo thoughts?

didoo:
@chazzmoney I have created a quick comp of what I would change, easier than explaining by words:
build-diagram
No strong opinions on the text, the only thing I am sure is that the stack of output files is confusing, better to use a generic flat symbol.

chazzmoney:
Yeah, the problem is that it is sometimes generating more than one file for a platform. Like iOS and android generate multiple files, but there is only one CSS generated usually. Want people to understand that whatever output is needed for a platform is what is created.

PR #210, opened 2018-12-01
v2.6.0 release

dbanksdesign:
*Issue #, if available:*

*Description of changes:* 2.6.0 release

Posting a PR on develop as a sanity check. We can then merge this, then merge develop into master. 

Note: I will fix the documentation whitespace and weirdness when we do a version. Some of the documentation gets lost b/c some documentation markdown files are generated and should not be edited. I'll try to fix that.

### DIFF
--- a/docs/formats.md
+++ b/docs/formats.md
@@ -92,7 +92,7 @@ You created a format and think it should be included? [Send us a PR](https://git
 
 Creates a CSS file with variable definitions based on the style dictionary
 
-**Example**  
+**Example**
 ```css
 :root {
   --color-background-base: #f0f0f0;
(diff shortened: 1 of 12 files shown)


didoo:
> b/c some documentation markdown files are generated and should not be edited. 
@dbanksdesign probably it was me. but I don't know which ones. how can I know which ones are automatically generated? (also, I suppose there is no way to avoid committing it to the repo, right?)

dbanksdesign:
Correct, we have to commit them to the repo to show up in the docs. The auto-generated ones are formats, transforms, and transformGroups markdown files, which get generated half from jsdocs and half templated content. We have hooks that run the scripts to generate those files to run during an npm version. I think I just need to do a better job at commenting on files and documenting how to document stuff. I'll take that up this week.

Issue #224, opened 2019-01-07
No Changelog


rjoaopereira:
Thank for this library. I would like to make a request for a `CHANGELOG`. The file exists but it is empty.

A way to achieve this without too much hassle is by using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/) and [conventional changelog](https://github.com/conventional-changelog/standard-version) but this is just a suggestion.

dbanksdesign:
YES! This has been on the back of my mind for a while but never got around to doing it. Auto-generating and releasing versions has been a sore spot. How does conventional commits work with branching? Right now we have a develop branch and then merge into master, honestly a bit ad hoc, whenever we want to do a minor/major release. We commit directly to master if it is changing docs or patch commits. Would we instead change to getting rid of develop and always committing on master? Are there any good articles to read on this subject?

rjoaopereira:
`How does conventional commits work with branching?`

Doesn't change it but I would recommend checking [git flow](https://datasift.github.io/gitflow/IntroducingGitFlow.html) and having 1 commit per PR (use rebase with squash + fixup)

[A quick search on the subject](http://devboosts.com/2017/03/28/conventional-commits/)

chazzmoney:
Added via #225 , thanks @rjoaopereira !

PR #225, opened 2019-01-07
feat: add automatic changelog generation

rjoaopereira:
#224

Asking for comments on this.

A possible solution for the issue.

Added support for automatic `changelog` generation through conventional-commits and standard-version.

Also added linting for pre-commit and commit-msg.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,15 +35,19 @@
     "NOTICE"
   ],
   "scripts": {
-    "lint": "eslint index.js lib/**/*.js test/*.js test/**/*.js",
+    "lint": "eslint --fix index.js lib/**/*.js test/*.js test/**/*.js",
     "test": "npm run lint && jest --runInBand",
     "test-watch": "npm run lint && jest --runInBand --watch",
     "coverage": "istanbul cover ./node_modules/mocha/bin/_mocha -- test/*.js test/**/*.js",
-    "preversion": "npm test",
-    "version": "node ./scripts/version.js && npm run generate-docs",
     "generate-docs": "node ./scripts/generateDocs.js",
     "serve-docs": "docsify serve docs -p 3000 -P 12345",
-    "install-cli": "npm install -g $(npm pack)"
+    "install-cli": "npm install -g $(npm pack)",
+    "release": "git add . && standard-version -a"
+  },
+  "standard-version": {
+    "scripts": {
+      "postbump": "node ./scripts/version.js && npm run generate-docs"
+    }
   },
   "jest": {
     "verbose": false,
@@ -65,8 +69,20 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm test"
+      "pre-commit": "npm test && lint-staged",
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
+  },
+  "lint-staged": {
+    "*.js": [
+      "npm run lint",
+      "git add"
+    ]
+  },
+  "commitlint": {
+    "extends": [
+      "@commitlint/config-conventional"
+    ]
   },
   "repository": {
     "type": "git",
@@ -95,6 +111,8 @@
     "tinycolor2": "^1.4.1"
   },
   "devDependencies": {
+    "@commitlint/cli": "^7.2.1",
+    "@commitlint/config-conventional": "^7.1.2",
     "babel-jest": "^23.6.0",
     "babel-preset-env": "^1.7.0",
     "chai": "^4.1.2",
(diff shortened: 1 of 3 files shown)


dbanksdesign:
I like this a lot, I just need to kind of wrap my head around it a bit. Are there other alternatives, and why is this way the best? It would be nice to also update the contributing docs to talk about how to commit properly using this new way, both for future contributors and for my own edification as well 😄

rjoaopereira:
Other options are available such as semantic-release. Some [info](https://github.com/conventional-changelog/standard-version#how-is-standard-version-different-from-semantic-release) on the differences.

For linting there is also [precise-commits](https://github.com/nrwl/precise-commits)

rjoaopereira:
I added a small reference in `CONTRIBUTING.md` to commit rules.
These basically just point to the conventional commits specification.

chazzmoney:
I think this could be good - and it seems like a two-way door. If it doesn't work for us, we could back this change out easily.

didoo:
Two things I am not sure about:

> Please follow the spec in order to have a successful commit.

I have just submitted a PR, and I had **no idea** of what tags to use (also, I have to remember to do it in every commit). I am not sure about putting this onus on the contributors, this could mean creating a remarkable barrier to entry.

> squash commits when merging pull requests

Personally, I prefer to see the history of the commits, to see the evolution of a feature, see the changes and why they've been done, etc. - I can live with it, anyway :)

rjoaopereira:
I understand your concerns. Let me see if I can help my case.

`I have just submitted a PR, and I had no idea of what tags to use`

Generally, `fix`and `feat` are the common ones. `BREAKING CHANGE` is mandatory for a `major`.
The normal config follows [angular's](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) norm.

`also, I have to remember to do it in every commit`

husky will remind you in every commit if you don't follow the spec. It just won't let you do it.

`Personally, I prefer to see the history of the commits, to see the evolution of a feature, see the changes and why they've been done, etc.`

I have lived in both ends of the spectrum and I prefer to have a clean history. Project management becomes easier. However, there are drawbacks like you say.

[this](https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/) blogpost does a good job of explaining the pros and cons.

rjoaopereira:
Kent C. Dodds made a nice [course](https://egghead.io/lessons/javascript-automating-releases-with-semantic-release) on egghead regarding open source js libraries.

A nice watch with reference to changelogs, commits and releases

dbanksdesign:
Apologies this is taking a bit longer. I'm diving more deeply into this. One issue with this (not your fault or the fault of conventional commits or anything) is we have an issue with our documentation generation step. In formats we have less formats, and each format has an `@example` block in the JSDoc which shows an example of the output of the format. Less uses the '@' character for variables, which breaks the jsdoc parsing because it thinks it is a JSDoc tag. I have an [open issue](https://github.com/jsdoc3/jsdoc/issues/1521) on the JSDoc library, which has not been responded to.. 😔

I'll try to do some more digging around..

Thank you for sending me that course, I watched it yesterday and it was very helpful.

davixyz:
I love this; Definitely automating the changelog through `Conventional Commits` is a big win.
Here's a nicer to the eyes spec of the Conventional Commits:
https://www.conventionalcommits.org/en/v1.0.0-beta.3/#specification

also, to defend @rjoaopereira. If it's absolutely necessary, people that contribute to this library can keep writing the commits how they want. It's the job of the maintainers to add the correct tag like `fix`, `feat`, etc when they merge squash the commits.

package.json
```diff
@@ -35,15 +35,15 @@
     "NOTICE"
   ],
   "scripts": {
-    "lint": "eslint index.js lib/**/*.js test/*.js test/**/*.js",
+    "lint": "eslint --fix index.js lib/**/*.js test/*.js test/**/*.js",
     "test": "npm run lint && jest --runInBand",
     "test-watch": "npm run lint && jest --runInBand --watch",
     "coverage": "istanbul cover ./node_modules/mocha/bin/_mocha -- test/*.js test/**/*.js",
-    "preversion": "npm test",
-    "version": "node ./scripts/version.js && npm run generate-docs",
```
dbanksdesign: Could we still call the `./scripts/version.js` file somehow in the release? All that file is doing is updating the version number of style dictionary in the examples to be the new version number. Or if there is a better way to do this lets do that.

rjoaopereira: Calling it and doc generation after version bump.
A suggestion would be to create a monorepo with lernajs and make examples a package that depends on style-dictionary. When bumping SD, all others would be updated. But this is another task

rjoaopereira: you can call `npm run release -- --dry-run` to test things out. It will run the postbump though.

dbanksdesign: Ohhh using lerna is a good idea!

Issue #215, opened 2018-12-12
config's "Include" and "source" properties behave differently.


rjoaopereira:
Not sure if this is intended but it leads to confusion.

`include` expects an array of paths but which cannot be globs.
https://github.com/amzn/style-dictionary/blob/master/lib/extend.js#L106

`source` expects an array of paths which can be globs.
https://github.com/amzn/style-dictionary/blob/master/lib/extend.js#L119

I believe it would be benificial to have the same behaviour. I came across this when promoting some properties to be base properties under include.

dbanksdesign:
You are totally right... I wrote that code and I don't know what I was thinking. I honestly thought before looking at the code just now that they did behave the same, but I guess I am wrong. I'll work on a PR later today to get that fixed.

rjoaopereira:
Thank you for your quick reply and for your work.

rjoaopereira:
No worries, this is not an impediment at all. Just opened the issue for consistency.
I'm the one who is thankful 😄

dbanksdesign:
Quick question, because you opened the conventional commits, do you think this is a bug or a feature?

rjoaopereira:
I can't find anywhere in the docs or README that include can be used with globs. Only `An array of paths to Style Dictionary property files that contain default styles`. I would go with feature.

rjoaopereira:
`source` on the other hand, and although not a strict reference, has `"source": ["properties/**/*.json"],` in the docs.

dbanksdesign:
I figured it out: #227

rjoaopereira:
Awesome, thanks!

chazzmoney:
Fixed in v2.7.0

PR #227, opened 2019-01-08
feat: Making include accept an array of globs, like source does

dbanksdesign:
*Issue #, if available:* #215 

*Description of changes:* Changing the behavior of `include` attribute on the config to match the behavior of `source`. Now `include` accepts an array of path globs, instead of just file paths. This change should be a superset change, as normal file paths still function the same. 

### DIFF
--- a/__tests__/extend.test.js
+++ b/__tests__/extend.test.js
@@ -82,7 +82,7 @@ describe('extend', () => {
       expect(StyleDictionaryExtended.properties).toEqual(helpers.fileToJSON(__dirname + "/__properties/paddings.json"));
     });
 
-    it('should override existing properties include is given', () => {
+    it('should override existing properties if include is given', () => {
       var StyleDictionaryExtended = StyleDictionary.extend({
         properties: test_props,
         include: [__dirname + "/__properties/paddings.json"]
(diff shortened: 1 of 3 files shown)


__tests__/extend.test.js
```diff
@@ -61,10 +61,33 @@ describe('extend', () => {
...
+    });
+
+    it('should build the properties object if an include is given', () => {
+      var StyleDictionaryExtended = StyleDictionary.extend({
+        "include": [__dirname + "/__properties/paddings.json"]
+      });
+      expect(StyleDictionaryExtended.properties).toEqual(helpers.fileToJSON(__dirname + "/__properties/paddings.json"));
+    });
+
+    it('should override existing properties include is given', () => {
```
rjoaopereira: Not my native language but is this description correct?

__tests__/extend.test.js
```diff
@@ -91,10 +114,11 @@ describe('extend', () => {
       ).toThrow('source must be an array');
     });
 
-    it('should throw if a path in the source array doesnt resolve', () => {
-      expect(
-        StyleDictionary.extend.bind(null, {include: ['foo']})
-      ).toThrow("Cannot find module 'foo'");
+    it('should not update properties if sourche glob paths dont resolve to anything', () => {
```
rjoaopereira: typos in `sourche` and `dont`

__tests__/extend.test.js
dbanksdesign: Good catch, looks like it is missing 'if' in between 'properties' and 'include'.

chazzmoney:
LGTM

Issue #207, opened 2018-11-30
Create new method 'registerFilter' to enable custom filtering


chazzmoney:
From #199:

Define a new registerFilter method in Style Dictionary (similar to what we have with registerFormat, registerTemplate, registerAction) so that we can do:

"filter": "name_of_the_filter"
(in the same way as we do "template": "name_of_template" or "format": "name_of_format").

Then in the filterProperties function declaration we can easily detect if the filter is a "custom" filter by the fact that is a string.

This creates consistency across how we define our configuration (and customizations of it).

didoo:
FYI I have started to work on this one.

chazzmoney:
Added in v2.7.0

Thanks @didoo !

Issue #199, opened 2018-11-17
Allow "filter" option in config file to support array of objects


didoo:
Calling in @elliotdickison here, since you worked on #101.
In the documentation we say that: 

>  platform.file.filter | Function/Object (optional) | A function or object used to filter the properties that will be included in the file. If a function is provided, each property will be passed to the function and the result (true or false) will determine whether the property is included. If an object is provided, each property will be matched against the object using a partial deep comparison to determine whether the property is included.

But if the config is a JSON file, there is no way to pass a function to the `filter` property (maybe we should update the documentation?), so the only way to apply a filter is via an object:
```
"filter": {
    "attributes": {
      "category": "color"
    }
}
```
which in the code is transformed to a matching filter function via the Lodash `matches` method:
```
const filterFunc = _.isFunction(filter) ? filter : _.matches(filter)
```
This also means we can filter **only** for one CTI property (eg. color, or size) at the time. In reality often is useful to have a filter that can work with multiple CTI properties (eg. "give me all the color and size properties") but as is now is not possible.

What I am proposing here is to extend a little bit the capabilities of the filter, and accept an array of objects. Something like this:
```
"filter": [{
    "attributes": {
      "category": "size"
    }
},{
    "attributes": {
      "category": "color"
    }
}]
```
where the `filterProperties` function would look something like this (notice: the array filter function is probably wrong, I have to investigate what is the correct way to write it, but it gives you a gist of what I mean):

```
function filterProperties(dictionary, filter) {
  if (!filter) {
    return dictionary
  } else {
    let filterFunc;
    if (_.isFunction(filter)) {
        filterFunc = filter;
    } else if (_.isArray(filter) && filter.every(_.isObject(filter))) {
        filterFunc = function (filter) {
            return filter.some(function(obj) { return _.matches(filter); });
        };
    } else if (_.isObject(filter)) {
        filterFunc = _.matches(filter);
    } else {
        // maybe throw an error here?
    }
    return _.assign({}, dictionary, {
      allProperties: filterPropertyArray(dictionary.allProperties, filterFunc),
      properties: filterPropertyObject(dictionary.properties, filterFunc)
    })
  }
}
```

What do you think of this? Adding too much complexity?
(pinging @dbanksdesign + @chazzmoney if they have some ideas/comments)

didoo:
Forgot to add: this is motivated by this: https://github.com/amzn/style-dictionary/pull/164#issuecomment-438219133

[...]

PR #233, opened 2019-01-19
Add registerFilter method

didoo:
*Issue #, if available:* #207 #199 

*Description of changes:*
- introduced a `registerFilter` method for the StyleDictionary module
- updated the `filterProperties.js` to receive the `filter` parameter only as a function
- moved the logic to handle the case when the `filter` parameter is an object in the `transformConfig` method, and added the handling of the case where the filter is a string (supposedly a filter registered via `registerFilter`, otherwise throw an error)
- updated the documentation accordingly


didoo:
@dbanksdesign @chazzmoney before continuing this PR (updating tests and documentation) I need some feedback from you. Have a look at the code, it works but maybe (probably) can be better.

Things I am not 100% sure:

1) I have moved the logic where we convert the filter from object/function to function (and added the case for the string (registered function) from the `filterProperties` method to the `transformConfig` one. The reason for that is that in filterProperties I don't have access to the StyleDictionary object (where the filters are registered). I have tried to pass it down, or to use `this` but without success. Do you think there would be a way to do it? Where is the "correct" location for this logic?
https://github.com/amzn/style-dictionary/blob/b7d1672b163e5941e860bd49ba7e3b8ce55031a0/lib/transform/config.js#L55-L70

2) I have left a comment, about the initialization of the `filter` object as a property of StyleDictionary: 
https://github.com/amzn/style-dictionary/blob/b7d1672b163e5941e860bd49ba7e3b8ce55031a0/lib/register/filter.js#L38
I get why we need to initialise it (templates and formats are always present in SD, filter not necessarily) but I am not sure if there is a place where is better to initialise it. Any better ideas?

dbanksdesign:
For question 1, you are correct in the placement. `transform/config.js` is meant to 'hydrate' the config object so it has functions rather than string keys for transforms/formats/actions/etc. Adding filters, this would be the proper place to do it. 

For question 2, the initialization happens in the main index.js file:
https://github.com/amzn/style-dictionary/blob/master/index.js#L36
The file that it requires returns an object. So we could just add:
```javascript
filter: {}
```
in there and call it a day. Or do we want to include built-in filters? Maybe not if we are trying to distribute/open up the transforms/formats/actions. Sorry for stream of conscious here.

didoo:
> question 1

Cool, I am happy then! :)

> question 2

Ahh, now I get it! Now that you have explained it, makes total sense. OK, so we have these options:

1.  we leave the "initialisation" in the `lib/register/filter.js` file, but is an antipattern
2. we add a line in `index.js` like you suggested (with a comment why is done like this)
3. we create an "empty" module `lib/common/filters` that returns an empty object, so is consistent with the others (we need to add a comment in the module anyway, and possibly also in the `index.js` to signal that that is a special module

Probably 2 is the simplest and clearest one. What do you think?

dbanksdesign:
I like 2 or 3. We can just go with 2, and then if we do add built-in filters in the future we can update it when that happens.

__tests__/filterProperties.test.js
```diff
@@ -112,7 +112,7 @@ describe('filterProperties', () => {
     expect(filteredDictionary.properties).not.toHaveProperty('color');
   });
 
-  it('should work with a filter object', () => {
+  xit('should work with a filter object', () => {
```
dbanksdesign: typo?

didoo: It was a skipped test (this is how you tell Jest to skip a test). I had forgot to fix it 😬
Now I have updated it correctly for the way the function works now.

dbanksdesign: Ohhh... TIL

didoo:
@dbanksdesign wait before merging, I still have to update the documentation :)

dbanksdesign:
Good call, I might have merged today by accident

didoo:
@dbanksdesign I have updated the documentation files. Do you think is enough?

chazzmoney:
LGTM

Issue #251, opened 2019-02-22
Discussion: outputting files 1:1


dbanksdesign:
This issue is meant to be an open discussion about a topic to see if this is something we want to support in Style Dictionary and how to do that. 

Because of the way Style Dictionary works, by merging all source token files together first, there is no easy way to have a 1-to-1 mapping of source token file to build artifact. For example, having a JSON file for each component like 'button.json', 'input.json', 'heading.json', etc. and you wanted output files like 'button.scss', 'input.scss', and 'heading.scss' for the web, and maybe something similar for other platforms. 

This is actually the way [Theo](https://github.com/salesforce-ux/theo) does things, you input a single file and output files from that. 

Is this use-case something Style Dictionary should support, and if so, what is the best way to do that?

Thanks to @elliotdickison for bringing this up!

elliotdickison:
Thanks for getting this conversation started! A few details about our use-case:

We do 1-1 source-to-build token files in our project specifically when generating web variables that are resolved at runtime (e.g. JavaScript variables or CSS3 variables). Since these variables are resolved at runtime they contribute to the final size of the application bundles that import them (unlike sass variables, which boil out at build time). Keeping our token modules granular allows us to only import the tokens that we need in our apps (e.g. import button css variables and not date picker css variables) and helps us avoid shipping unnecessary bytes to the browser.

chazzmoney:
@elliotdickison correct me if I am wrong, but this seems like it could be related to #104 as well?

If so, I don't see why we couldn't add in two processing options. The first would do exactly what you had requested in 104. The second would allow a specification of how to output files, including outputting at a specific depth level or based on some metadata property.

Using both of these together could enable the 1:1 functionality you are describing. The important question is how to incorporate this correctly into the codebase. The input setting as either a config option or a separate metadata item? The output setting maybe as a platform option?

I know this might seem like a lot of added functionality, but it seems to me to be working towards being less opinionated about how to specify tokens. @didoo I think you don't use CTI at all - would this be a step in the right direction in your opinion?

elliotdickison:
Re: #104 - definitely related, keeping our token files granular means that we end up with a lot more files and rely more on file structure for organization. This makes the token paths pretty redundant in most cases (e.g. "/my/token/module.js" would contain something like `{ my: { token: { module: { a, b, c } } }`).

davixyz:
[...]

Issue #104, opened 2018-06-30
Treat file path as prefix for property path


elliotdickison:
### Proposal

Include the file path to a property (relative to the source dir provided to style-dictionary) as a prefix to the property path. For example this file:
`core/color.json`
```json
{
  "red": { "value": "#F00" }
}
```
would result in a property with the path: `["core", "color", "red"]` .

This would be a breaking change.

### Reasoning

Currently the benefits of splitting properties into multiple files are limited since property keys still have to be specified in full in each file, and the keys in a file may not actually match the path of a file. This proposed change would allow larger projects to take advantage of nested folders and multiple property files to significantly simplify the json required and ensure consistency between file paths and property keys.

### Background

While style-dictionary allows you to split properties into multiple files/folders, this currently has no effect on the way you key properties within the files. For example, splitting `all.json` into `size.json` and `color.json` still requires you to key properties by `size` and `color`:

`all.json`
```json
{
  "size": {
    "small": { "value": "4px" }
  },
  "color": {
    "red": "#F00"
  }
}
```
`size.json`
```json
{
  "size": {
    "small": { "value": "4px" },
    "large": { "value": "8px" }
  },
}
```
`color.json`
```json
{
  "color": {
    "red": "#F00",
    "blue": "#00F"
  }
}
```

In our project I found that this caused significant redundancy in our property files. For example I had a file `base/icon/checkmark.json` that looked something like this:
```json
{
  "base": {
    "icon": {
      "checkmark": {
        "url": "<cdn url here>"
      }
    }
  }
}
```
Multiplied across dozens of property files nested in multiple folders, this created significant overhead and opportunities for typos etc. I considered two solutions: use fewer files or pre-process our property files to automatically generate the needed keys from the file path. I really like the ability to organize properties in different files since we have hundreds of properties (400+ and counting) so I didn't want to start merging them back into large json files. I went the route of preprocessing our property files and generating json keys from file paths, and it's been working very well. We now have a setup where the following file:
`base/icon/checkmark.json`
```
{
  "url": "<cdn url here>"
}
```
is preprocessed so that the actual property path for the `url` property is `["base", "icon", "checkmark", "url"]`. This has simplified our json and allows you to quickly understand the path of a property by looking at the directory structure.

chazzmoney:
Like this a lot. There is something that is weird / bothers me about how deep merge loses all context of where things are coming from. I have a similar feeling about 'includes' vs 'source'.

A few thoughts that may have some relevance:

[...]

PR #253, opened 2019-02-27
feat: file path as cti path (example mechanism for discussion)

chazzmoney:
#104, part of #251

Added in options in config to use file path as CTI prefix.

This is meant to start discussion of value here.

The options to use in the config are 'sourceFilePathAsCTI' and 'includeFilePathAsCTI'.

Note that I did not attempt to solve the problem if you are using the current system, which would result in a duplication of information in the CTI as per @elliotdickison's comments. (e.g. A file in the path properties/fonts/size.json containing { fonts: { size: { small: value: '10px'; } } } using this option would result in fonts.size.fonts.size.small).

If we like this, I can add tests and add in functionality to detect and remove duplication. Also, let me know if I should remove that top level folder if it is called "properties", or what we should do to handle that.....

### DIFF
--- a/lib/utils/combineJSON.js
+++ b/lib/utils/combineJSON.js
@@ -28,7 +28,7 @@ var glob = require('glob'),
  * @param {Function} collision - A function to be called when a name collision happens that isn't a normal deep merge of objects
  * @returns {Object}
  */
-function combineJSON(arr, deep, collision) {
+function combineJSON(arr, deep, collision, useFilePathAsCTI) {
   var i, files = [],
     to_ret = {};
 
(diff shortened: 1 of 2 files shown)


lib/utils/combineJSON.js
```diff
@@ -48,6 +48,34 @@ function combineJSON(arr, deep, collision) {
       throw e;
     }
 
+    if(useFilePathAsCTI) {
+      let filePath = path.relative('./', files[i]);
```
elliotdickison: I think it might be too restricting to assume that the names should start from the current working directory. For example we would run the command from our package root, but our properties were in `<root>/src/tokens/<properties>`. In that case everything would be prefixed with "srcTokens". What if "rootPath" or something was a configuration option? That would also solve the issue you commented about on line 59 with assuming a "properties" folder name.

lib/utils/combineJSON.js
```diff
@@ -48,6 +48,34 @@ function combineJSON(arr, deep, collision) {
...
+      let ctisetup = filePath.split('.');
+      if(ctisetup.length > 1) {
+        ctisetup.pop();
+      }
+      ctisetup = ctisetup.join('.').split('/');
```
elliotdickison: Minor, but I think `.split(path.sep)` is more cross-platform-proof.

elliotdickison:
FWIW I don't think the duplication is a big deal. It doesn't feel like something that needs to be built in since the only purpose would be to help with migrating, right? If it's an issue maybe a migration script would be nice...

lib/utils/combineJSON.js
chazzmoney: 👍🏽 Good call on all accounts

crypt096 (approved):

jorenbroekema:
Since DTCG spec, token.type tends to be used to determine token type and users can use the token.filePath to create a name transform that takes that into account as well as the token hierarchy/path to determine the final token name. I think that makes this PR redundant and since it's 5 years old, I'll close for now

Issue #246, opened 2019-02-07
Add an advanced example with the gist shared by Dbanksdesign about CTI structure


didoo:
In a thread on Slack @dbanksdesign shared this gist:
https://gist.github.com/dbanksdesign/dd67b7c265b75089c03d969dceaa41f1

It may be an interesting example to add to the "advanced" examples. What do you think?

davixyz:
I would love to do this; It's actually what I'm using at work thanks to @dbanksdesign.
Here's my project:
https://github.com/davixyz/design-tokens

I might be adding something similar to this

chazzmoney:
In progress, #267 by @dbanksdesign 

Will close this once merged.

PR #267, opened 2019-04-22
docs(example): adding component cti example

dbanksdesign:
*Issue #, if available:* #246

*Description of changes:* Adding component CTI example which shows how to monkey patch built-in transforms and formats as well as showing one way to organize component tokens. 

### DIFF
--- a/examples/advanced/component-cti/README.md
+++ b/examples/advanced/component-cti/README.md
@@ -1,6 +1,58 @@
 ## Component CTI Structure
 
-This example will show you one way to define component tokens in an easy way while still using the CTI attributes.
+This example will show you one way to define component tokens in an easy way while still using the CTI attributes for transforming the tokens. The CTI structure for Style Dictionary tokens makes defining component-level tokens really cumbersome and not user-friendly. This is because the CTI of tokens are based on the object path of the token. So if you wanted to write tokens for a button component it would have to look something like this:
+
+```json
+{
+  "size": {
+    "padding" : {
+      "button": { "value": "{size.padding.base.value}" }
+    },
+    "font": {
+      "button": { "value": 2 }
+    }
+  },
+
+  "color": {
+    "font": {
+      "button": {
+        "primary": { "value": "{color.font.inverse.value}" },
+        "secondary": { "value": "{color.font.link.value}" }
+      }
+    },
+
+    "background": {
+      "button": {
+        "primary": { "value": "hsl(10, 80, 50)" },
+        "secondary": { "value": "{color.background.primary.value}" }
+      }
+    }
+  }
+}
+```
+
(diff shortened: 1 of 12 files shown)


dbanksdesign:
@didoo thoughts?

examples/advanced/component-cti/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "style-dictionary-tokens-deprecation",
```
chazzmoney: not sure the name is correct

examples/advanced/component-cti/README.md
```diff
@@ -0,0 +1,40 @@
+## Component CTI Structure
+
+This example will show you one way to define component tokens in an easy way while still using the CTI attributes.
```
chazzmoney: Is this a sufficient explanation for the user problem that this solves? I think we should lay out the issue in clarity. A user may not understand the solution, but they should be able to identify with the problem to know that this is the example they should be looking at.

examples/advanced/component-cti/README.md
```diff
@@ -0,0 +1,40 @@
...
+// }
+```
+
+In this example we monkey patch the `attribute/cti` transform to look at the top-level namespace of the token and if it is 'component', instead of running the default `attribute/cti` transform, it instead looks at the last part of the object path to generate the equivalent category and type.
+
+For example, if the key of the token (the last part of the object path) is "background-color", the monkey patched `attribute/cti` transform will return a category of 'color' and type of 'background'. This example uses CSS property names, but you could also change it to use similar React/CSS-in-JS names like 'backgroundColor' instead of 'background-color'. 
+
+### Why do this?
+
+The CTI structure for Style Dictionary tokens makes defining component-level tokens really cumbersome and not user-friendly. Instead, when defining component tokens it makes more sense to structure them more like CSS. `component.button.background-color` makes more sense than `color.background.button`
```
chazzmoney: Can we describe how it is cumbersome? Either here or up above; maybe move this section up there if you need more space. Organization of tokens seems to be one of the core challenges that people are trying to solve with Style Dictionary and we should make it easy to find and use this if this is the solution we are suggesting.

chazzmoney:
Also, might want to add a bullet item to docs/examples.md

examples/advanced/component-cti/package.json
dbanksdesign: Fixed

examples/advanced/component-cti/README.md
dbanksdesign: I moved the 'why do this section' up here because it better explains the problem. I also added an example of what the problem is with a token file in here. Hopefully that explains the issue a bit better.

didoo:
Sorry for coming late. LGTM.

chazzmoney:
LGTM! :shipit:

Issue #257, opened 2019-03-15
Camelcase in `json/nested`


jduthon:
Hello `style-dictionnary` team,

We are having a small problem, not sure if it is a bug or if we are incorrectly using the project.
In our properties, we have colors defined something like:

```
{
  "color": {
      "space-gray": {
          "value": "#ccc"
      },
      "space-gray-w50": {
          "value": "#dadada"
      }
  }
}
```

As an output from the `json-nested` transform group, and more particularly from the `name/cti/camel`, I would expect to find as an output:

```
{
  "color": {
      "spaceGray": "#ccc",
      "spaceGrayW50": "#dadada"
  }
}
```
Unfortunately the keys of the values themselves aren't camelCased and I get:

```
{
  "color": {
      "space-gray": "#ccc",
      "space-gray-w50": "#dadada"
  }
}
```

Is that expected behaviour?

I see how we could try to nest things more so that the path is picked up better, but we would loose the ability to have the first one being only `spaceGray` and not `spaceGrayBase` or something similar.

dbanksdesign:
Hmm that is expected behavior. Let me try to explain why. 

Because design tokens are defined in a nested object structure in style dictionary, we need to be able to flatten that object for platforms and filetypes that expect a flat array of variables, like and an Android resource XML file, or a CSS variables file. Internally what happens is any node in the object structure that has a "value" attribute is added to an array of tokens. To be able to turn the object structure into variable names, Style Dictionary adds a path array on the token with the object path to get to that token. In your example the object path would be `['color', 'space-gray']`. The name transforms that are included, such as [name/cti/camel](https://github.com/amzn/style-dictionary/blob/master/lib/common/transforms.js#L139) use this path to create a name attribute on the token, so after being transformed, the token would look like:
```json
{
  "name": "colorSpaceGray",
  "value": "#ccc",
  "path": ["color", "space-gray"]
}
```

Name transforms edit the "name" attribute on the token object itself, and not the overall object structure itself. 

Nested formats will keep the object structure of your original JSON, but because the name transforms only affect the tokens themselves and not any other part of the object, that is why it is "space-gray" is not camel-cased. If you want to take a look at the raw, transformed style dictionary object, you can use the 'javascript/module' format, which just outputs the whole object.

[...]

Issue #153, opened 2018-10-04
Rewrite to ES6


davixyz:
Had a talk with @dbanksdesign and basically we could start migrating the library little by little to use ES6 features. Of course this needs some infrastructure work (that I would love to do) and also to take some decisions about how do we do this. Opening this issue to start the conversation for this.

### Some Thoughts
- Some of the features are already available in node 6 and node 8, probably the most important ones we would want like template literals, spreads, destructuring etc
- Replacing and using arrow functions everywhere might be tricky because of how style dictionary API has been designed with `this` in mind
- A build system will be required to ship the code for older versions

### Some build systems
- **Webpack**: https://webpack.js.org/guides/author-libraries/
- **Rollup**: https://rollupjs.org/guide/en

vvscode:
There is https://github.com/5to6/5to6

I'm ready to do it

chazzmoney:
Thats cool! I didn't know that was even a thing that existed.

If you are ready, go for it. Otherwise I can take a look next week.

chazzmoney:
Probably best done in conjunction with #154.

I think we should wait on this until we merge #169 before doing this.

vvscode:
So I'm waiting for merge https://github.com/amzn/style-dictionary/pull/169 to make this one and https://github.com/amzn/style-dictionary/issues/154

To which branch?

dbanksdesign:
develop-3.0

vvscode:
@dbanksdesign @chazzmoney  are you going to merge `develop` to `develop-3.0` ?

chazzmoney:
Yes, sorry for the delay. Going to try to get #158 and #152 merged into develop in the next couple days and then merge develop into develop-3.0

dbanksdesign:
develop-3.0 is now up to date with develop

chazzmoney:
Promise to get #152 merged soon - almost there. Thanks for your patience @vvscode.

dbanksdesign:
#500 is the beginnings of this

dbanksdesign:
@chazzmoney you think we can close this?

davixyz:
If this is finished, we should close it! Hello everyone!

dbanksdesign:
This is done (for the most part)! Closing!

PR #273, opened 2019-05-17
first cut at catching property name collisions during file output

chazzmoney:
Bug found in #257, but not solving #257

Code to catch collisions in the array of properties immediately before file output.


chazzmoney:
This it NOT ready. Two issues:

1. I changed the basic example to cause errors so you can see what the output looks like
2. If y'all like it, I need to add tests

chazzmoney:
@didoo and @dbanksdesign let me know what you think

lib/utils/groupMessages.js
```diff
@@ -20,6 +20,7 @@ var GroupMessages = {
     RegisterTemplateDeprecationWarnings: 'Register Template Deprecation Warnings',
     SassMapFormatDeprecationWarnings: 'Sass Map Format Deprecation Warnings',
     MissingRegisterTransformErrors: 'Missing Register Transform Errors',
+    PropertyNameCollisionWarnings: 'Property Name Collision Warnings',
```
dbanksdesign: We are getting some real use out of the groupMessages! I never thought we'd have this many warnings and errors, but it follows our tenet to fail loudly! Super cool.

lib/buildFile.js
```diff
@@ -42,8 +45,49 @@ function buildFile(destination, format, platform, dictionary, filter) {
...
+  });
+
+  let propertyNamesCollisionCount = GroupMessages.count(PROPERTY_NAME_COLLISION_WARNINGS);
+
+  fs.writeFileSync(destination, format(filteredProperties, platform));
+  console.log(chalk.bold.green((propertyNamesCollisionCount > 0) ? '⚠️ ' : '✔︎ ') + ' ' + destination);
+
+  if(propertyNamesCollisionCount > 0) {
+    var propertyNamesCollisionWarnings = GroupMessages.flush(PROPERTY_NAME_COLLISION_WARNINGS).join('\n    ');
+    let title = `While building ${chalk.keyword(HIGHLIGHT_COLOR).bold(dest)}, token collisions were found; output may be unexpected.`;
```
dbanksdesign: "output may be unexpected" love it

lib/buildFile.js
```diff
@@ -42,8 +45,49 @@ function buildFile(destination, format, platform, dictionary, filter) {
...
+    }
+  });
+
+  let propertyNamesCollisionCount = GroupMessages.count(PROPERTY_NAME_COLLISION_WARNINGS);
+
+  fs.writeFileSync(destination, format(filteredProperties, platform));
+  console.log(chalk.bold.green((propertyNamesCollisionCount > 0) ? '⚠️ ' : '✔︎ ') + ' ' + destination);
+
+  if(propertyNamesCollisionCount > 0) {
+    var propertyNamesCollisionWarnings = GroupMessages.flush(PROPERTY_NAME_COLLISION_WARNINGS).join('\n    ');
```
dbanksdesign: Nit: you are using `let` in other places but `var` here, should we be consistent?

lib/buildFile.js
```diff
@@ -42,8 +45,49 @@ function buildFile(destination, format, platform, dictionary, filter) {
   if (!fs.existsSync(dirname))
     fs.mkdirsSync(dirname);
 
-  fs.writeFileSync(destination, format(filterProperties(dictionary, filter), platform));
-  console.log(chalk.bold.green('✔︎') + ' ' + destination);
+  let filteredProperties = filterProperties(dictionary, filter);
+
+  // Check for property name Collisions
+  var nameCollisionObj = {};
+  filteredProperties.allProperties && filteredProperties.allProperties.forEach((propertyData) => {
```
dbanksdesign: Do we need this check? I think when we are constructing `allProperties` it will always be at least an empty array?

lib/buildFile.js
```diff
@@ -42,8 +45,49 @@ function buildFile(destination, format, platform, dictionary, filter) {
...
+      nameCollisionObj[propertyName] = [];
+    }
+    nameCollisionObj[propertyName].push(propertyData);
+  });
+
+  var PROPERTY_NAME_COLLISION_WARNINGS = GroupMessages.GROUP.PropertyNameCollisionWarnings;
+  var WARNING_COLOR = 'darkorange';
+  var HIGHLIGHT_COLOR = 'orangered';
+  var HELP_COLOR = 'orange';
+  Object.keys(nameCollisionObj).map((propertyName) => {
```
dbanksdesign: Change `.map` here to `.forEach` since we aren't actually mapping the array to some variable.

lib/buildFile.js
```diff
@@ -42,8 +45,49 @@ function buildFile(destination, format, platform, dictionary, filter) {
...
+    if(nameCollisionObj[propertyName].length > 1) {
+      let collisions = nameCollisionObj[propertyName].map(properties => chalk.keyword(HIGHLIGHT_COLOR)(properties.path.join('.')) + '   ' + chalk.keyword(WARNING_COLOR)(properties.value)).join('\n        ');
```
dbanksdesign: This line is a bit long and hard to read. Maybe break this into a few lines?

lib/buildFile.js
```diff
@@ -42,8 +45,49 @@ function buildFile(destination, format, platform, dictionary, filter) {
...
+
+  let propertyNamesCollisionCount = GroupMessages.count(PROPERTY_NAME_COLLISION_WARNINGS);
+
+  fs.writeFileSync(destination, format(filteredProperties, platform));
+  console.log(chalk.bold.green((propertyNamesCollisionCount > 0) ? '⚠️ ' : '✔︎ ') + ' ' + destination);
+
+  if(propertyNamesCollisionCount > 0) {
+    var propertyNamesCollisionWarnings = GroupMessages.flush(PROPERTY_NAME_COLLISION_WARNINGS).join('\n    ');
+    let title = `While building ${chalk.keyword(HIGHLIGHT_COLOR).bold(dest)}, token collisions were found; output may be unexpected.`;
+    let help = chalk.keyword(HELP_COLOR)([
```
dbanksdesign: Maybe use back-ticks here since all you need the array for is to put new line characters in between each line.  I like this message though. You could probably put `title` and `help` into a single string literal to make this cleaner.

lib/transform/config.js
```diff
@@ -29,6 +29,7 @@ var MISSING_TRANSFORM_ERRORS = GroupMessages.GROUP.MissingRegisterTransformError
  */
 function transformConfig(config, dictionary, platformName) {
   var to_ret = _.clone(config);
+  to_ret.platformName = platformName;
```
dbanksdesign: Is this being used elsewhere? I can't find this being used in this PR.

lib/buildFile.js
```diff
@@ -33,6 +34,8 @@ function buildFile(destination, format, platform, dictionary, filter) {
   if (!destination || typeof destination !== 'string')
     throw new Error('Please enter a valid destination');
 
+  var dest = destination;
```
didoo: probably I would not do this, but create a variable `fullDestination` (or something similar, I am not sure about the name) which is the one used and manipulated, and leave the value `destination` untouched, so it can be used directly in the construction of the title if propertyNamesCollisionCount > 0

lib/buildFile.js
```diff
@@ -42,8 +45,49 @@ function buildFile(destination, format, platform, dictionary, filter) {
...
+  filteredProperties.allProperties && filteredProperties.allProperties.forEach((propertyData) => {
+    let propertyName = propertyData.name;
+    if(!nameCollisionObj[propertyName]) {
+      nameCollisionObj[propertyName] = [];
+    }
+    nameCollisionObj[propertyName].push(propertyData);
+  });
+
+  var PROPERTY_NAME_COLLISION_WARNINGS = GroupMessages.GROUP.PropertyNameCollisionWarnings;
+  var WARNING_COLOR = 'darkorange';
```
didoo: I would not create variables just for this, at the end are just colors in Chalk, even if they are copy&pasted no big deal

didoo:
I have just added a couple of comments, but overall seems good to me.

lib/buildFile.js
chazzmoney: Whoops. Good call.

lib/buildFile.js
chazzmoney: I've had massive problems with formatting trying to use multiple backtick sections into each other. This was my solution.

I'm going to hold off on this change for now, and when we get everything else to a good place if you have a good suggestion maybe you can do a PR into this one - it was literally the piece that took me the longest.

lib/transform/config.js
chazzmoney: No it is not. I was using it to generate the message originally, but then realized I could just use the current output stream instead of generating something unique.

I thought it made sense to have the platform name present in the platform (that data is currently not available), so I left it. I can pull to a different PR if you want.

lib/buildFile.js
chazzmoney: If you guys like the current coloring, I can make this change. It as super painful changing colors when I was creating the output message so I used this to make it simpler.

If everyone confirms they like the output, I'll change this out.

lib/buildFile.js
chazzmoney: Unfortunately this is required. The tests do not generate allProperties at all in some cases, causing an error to throw here and the tests to fail.

lib/buildFile.js
chazzmoney: We should switch everything to ES6 as per #153 . I'm trying to stay aligned with the technique used in each file.

chazzmoney: Crap, nm - using both in this file. I was moving things around a bit per our conversation and didn't change after a copy and paste. Will fix.

chazzmoney: Ehhhhhhh.... so I just set all root level functional variables to use var and the block scoped variable to use let. Whatcha think?

lib/buildFile.js
chazzmoney: agreed

lib/transform/config.js
dbanksdesign: Yea let's pull it to a different PR, just so this one is focused on a single problem.

lib/buildFile.js
dbanksdesign: I guess that is probably fine for now, and tackle it all in #153

lib/buildFile.js
dbanksdesign: LGTM

lib/buildFile.js
dbanksdesign: Sounds good, after we merge this I can try to do a cleanup PR on this area.

lib/buildFile.js
chazzmoney: changed to in-line

lib/transform/config.js
chazzmoney: removed

chazzmoney:
Sorry, was adding the tests still - and adding them required me to change something (the key for grouping messages is now per destination). Can you peek one more time? @dbanksdesign

dbanksdesign:
Can you take a look at the test failure on travis?

chazzmoney:
Well 💩. So apparently color coding in chalk differs on different platforms. Soooo... I dunno. I’ll come up with a way to rewrite the test.

dbanksdesign:
You could use chalk methods in the test? Or how are we testing other console outputs that use chalk?

chazzmoney:
Yeah, exactly what I did. Let me know if you still approve and we can MERGE BABY MERGE. 🤓

PR #277, opened 2019-05-21
fix(cleanfile): add file check and log for non-existent file

gwyneplaine:
*Description of changes:*

This change is to ensure that style-dictionary doesn't throw when cleanFile() is invoked on a file that doesn't (yet) exist. 

For example in our project, we have a build script that always calls cleanAllPlatform() before buildAllPlatform(), with the current logic this will throw on a clean install (which is not ideal). 


__tests__/cleanFile.test.js
```diff
@@ -10,7 +10,6 @@
  * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
  * and limitations under the License.
  */
-
```
chazzmoney: Can we add this back in?

__tests__/cleanFile.test.js
```diff
@@ -35,4 +34,14 @@ describe('cleanFile', () => {
     expect(helpers.fileDoesNotExist('./__tests__/__output/test.txt')).toBeTruthy();
   });
 
+  describe('if a file does not exist', () => {
+    it('should not throw', () => {
+      try {
+        cleanFile('test.txt', format, {buildPath: '__tests__/__output/'}, {});
```
chazzmoney: Looks good, let’s rename to nonexistent.txt for clarity?

dbanksdesign: Could we also change this to use an `expect` method like we do in other tests checking if something should or shouldn't throw? For example:
https://github.com/amzn/style-dictionary/blob/fa6ebeb6d6120fda4ca04934b554758a3727988f/__tests__/exportPlatform.test.js#L42

dbanksdesign:
Just one comment on the test, and I agree with @chazzmoney's comments. Thank you for submitting this PR! I actually just ran into this issue recently...

gwyneplaine:
@dbanksdesign @chazzmoney  thanks guys, feedback enacted. 
Love the library btw, fantastic work  🎉

Issue #285, opened 2019-05-30
Sketch color palette


MarjorieSIX:
Hello,
I don't think I have an issue per say, but I'm having a hard time creating the sketch color palette via the transform function sketch/palette.
My scss works fine, Android and IOS too.
But as for the sketch palette, it creates an empty file with just
```
{
  "compatibleVersion": "1.0",
  "pluginVersion": "1.1",
  "colors": []
}
```
If there's another place I can ask, please forward me :-D

Thanks for your help,
M.

dbanksdesign:
The 'sketch/palette' format filters tokens that have

```json
{
  attribute: {
    category: 'color',
    type: 'base'
  }
}
```

Here is the source: https://github.com/amzn/style-dictionary/blob/master/lib/common/formats.js#L710

We recently-ish added 'filters' for you to filter your tokens before formatting them, so we should remove that filter in the format itself. In the meantime, if your tokens have those attributes (category === 'color' and type === 'base') it will work then. You can do this either by adding an attribute object on your tokens directly with those properties, or set up your tokens to be like this:

```json
{
  "color": {
    "base": {
      "white": {
        //...
```

So the fix would be to remove the `.filter` code in the 'sketch/palette` format, and that is actually it. Once we do that, you will need to make sure to add a 'filter' in your file config to filter only the color tokens or you might get some weirdness.

dbanksdesign:
A bit of a bigger issue, when trying out sketch palettes, I had an old version of the plugin so I needed to update it. The plugin is now a major version above (2.2 vs. 1.1), which has an impact on the format because now the .sketchpalette file takes a slightly different form.

PR #287, opened 2019-05-30
fix(sketch): fix sketch palette format to use new filters

dbanksdesign:
*Issue #, if available:* #285

*Description of changes:* Removing filtering from the sketch/palette format because we handle filtering before formats now.


dbanksdesign:
There is a new version of the sketchpalette plugin, and I can't get the old version to work. I'm thinking to not break backwards compatibility, we could instead create a new format with 'sketch/palette/v2'?

chazzmoney:
👍🏾

custa1200:
I agree with the idea to create a new one is probably the best path forward. The naming is something I think we'd need in Sketch Palettes, I can see no value in putting effort into the old one. Unfortunately I had to park my SD integration for the moment so can't get back to look at it this sprint.

chazzmoney:
LGTM :shipit:

Issue #290, opened 2019-06-12
Allow scss/map-deep and scss/map-flat to customize the map name


dbanksdesign:
Change this line:
https://github.com/amzn/style-dictionary/blob/master/lib/common/templates/scss/map-deep.template#L47

To accept a configurable name with `$tokens` as the fallback

PR #291, opened 2019-06-12
fix(format): adding configurable name to sass map name

dbanksdesign:
*Issue #, if available:* #290 

*Description of changes:* Allowing configurable sass map names

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -98,6 +98,8 @@ module.exports = {
   /**
    * Creates a SCSS file with a flat map based on the style dictionary
    *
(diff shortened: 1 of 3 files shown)


dbanksdesign:
There is... good point. I will add it in the JSDoc comments here: 
https://github.com/amzn/style-dictionary/blob/master/lib/common/formats.js#L98

dbanksdesign:
And it looks like I already had a commit ready, take a look now. I didn't run the command to generate the docs, that will happen automatically in the next release.

chazzmoney:
LGTM :shipit:

Issue #305, opened 2019-07-03
CSS style comments used in scss/variables format


nategreen:
Since the comments used in the scss/variables format are `/* ... */` style, they show up (oddly) in the final output.

`_variables.scss` (generated by style-dictionary)
```
/**
 * Do not edit directly
 * Generated on Wed, 03 Jul 2019 14:49:57 GMT
 */

$some-variable: blue; /* The color of the sky */
```

`button.scss`
```
@import 'variables';

button {
  background: $some-variable;
}
```

`button.css` (compiled output from Sass)
```
/**
 * Do not edit directly
 * Generated on Wed, 03 Jul 2019 14:49:57 GMT
 */
/* The color of the sky */

button {
  background: blue;
}
```

Looking at #195 and #196 it seems this was done to ensure that the comments would not break CSS, but as you can see, it really doesn't make sense to have CSS-style comments on the Sass variables when the variables themselves will not be present in the final output.

dbanksdesign:
Interesting... The fix would be to use `/* */` style comments in CSS and `//` style comments in Sass formats? Sounds easy enough to fix.

dbanksdesign:
Do you know if Less has the same problem?

dbanksdesign:
Just started a PR to fix this. 👆

nategreen:
> Do you know if Less has the same problem?

It looks like Less has the same comment style as Sass. Their docs don't seem to say explicitly that the `//` comments are removed, but when I tried it on [this online preview tool](http://lesscss.org/less-preview/), the comments behaved like Sass's do: block comments are preserved, inline comments are removed. 👍

dbanksdesign:
Awesome I'll change the less variables format too!

PR #306, opened 2019-07-03
fix(formats): change scss comments to short version

dbanksdesign:
*Issue #, if available:* #305 

*Description of changes:* Change scss variables format to use shorthand comments `//` rather than block comments so they don't pollute the compiled CSS. 


didoo:
LGTM!

chazzmoney:
We should make the same change to the less format as per @nategreen comment in #305

Otherwise, LGTM!

chazzmoney:
👍

PR #314, opened 2019-08-14
[FIX] UIColor transformer - Increase to 3 decimals to retain 8bit precision

patriknyblad:
*Description of changes:*
Two decimals only allow for 100 different shades for each channel. Using a gray scale color of #1d with only 2 decimals of precision would produce 0.11 for each channel, converting this number back to hex would result in #1c which is incorrect. This is now fixed by using 3 decimals of precision allowing a theoretical 1000 different shades.

### DIFF
--- a/__tests__/common/transforms.test.js
+++ b/__tests__/common/transforms.test.js
@@ -286,14 +286,21 @@ describe('common', () => {
         var value = transforms["color/UIColor"].transformer({
           value: "#aaaaaa"
         });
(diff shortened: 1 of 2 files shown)


chazzmoney:
Since we are in the range of 0/255 to 255/255, fixed(3) will work, although swift supports any float value (included those longer than 3 decimals).

:shipit:

shaunbent:
Hej @dbanksdesign, do you have an idea of when you plan to release SD next? The last release was back in July.

dbanksdesign:
I think we can do a release this week. We really should get a more consistent mechanism for releasing... So I created an issue for that: #321

Issue #324, opened 2019-09-06
Does not resolve references to zero


dbanksdesign:
Scenario:
If a design token/style property has a value of `0` (a number, not a string) and you have a reference/alias to that value, when you build Style Dictionary says the reference doesn't exist. 

Here is a quick test:

```javascript
const StyleDictionary = require('style-dictionary');

StyleDictionary.extend({
  properties: {
    test: { value: "{zero.value}" },
    zero: { value: 0 }
  },
  platforms: {
    css: {
      transformGroup: 'web',
      files: [{
        destination: "test.css",
        format: "css/variables"
      }]
    }
  }
}).buildAllPlatforms();
```

PR #325, opened 2019-09-06
fix(utils): handle 0 as a reference value

dbanksdesign:
*Issue #, if available:* #324

*Description of changes:* Adding proper checking for alias/reference resolution so that `0` is a valid value. Now we check for undefined rather than "truthiness", because 0 is "false". 

### DIFF
--- a/lib/utils/resolveObject.js
+++ b/lib/utils/resolveObject.js
@@ -88,7 +88,7 @@ function compile_value(value, stack) {
     // Find what the value is referencing
     ref = selfRef(variable.trim(), updated_object);
 
-    if (ref) {
+    if (typeof ref !== 'undefined') {
       if (typeof ref === 'string') {
         to_ret = value.replace(match, ref);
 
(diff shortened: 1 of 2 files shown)


lib/utils/resolveObject.js
```diff
@@ -150,24 +150,24 @@ function selfRef(str, obj) {
...
+  if (typeof ref === 'undefined') {
     GroupMessages.add(
       PROPERTY_REFERENCE_WARNINGS,
       "Reference doesn't exist: " + context + " tries to reference " + str + ", which is not defined"
     );
+  } else {
+    return ref;
   }
-
-  return ref;
```
elliotdickison: It looks like in the previous version a bad ref would result in this function returning "obj" (ref = obj on line 156, return ref here). Now it doesn't return anything if there's a bad ref. Don't know if this is good or bad, just wondering if it might have the potential to break something else.

dbanksdesign: Oh good point I forgot to mention that part. So I don't know why we were returning the whole object, if you look at line 91 we were testing for truthiness, which if we passed an object back would always be true. Then in line 131 we set `to_ret = ref` which would be that object as well and return it. This is inside a string.replace function, and returning an object does nothing to the string. So this was always "working", but not how we thought it was...

dbanksdesign: And hopefully all our glorious unit tests would catch anything major, but seems everything is in order.

chazzmoney: This file is so hard to understand - I'm concerned about the out-of-scope variable (updated_object) used within. I'm pretty sure I added the code that sets ref = obj as part of the work to provide better error output (although blame does not seem to think so), but I'm not clear how it interfaces with the updated_object variable.

chazzmoney: ok so updated_object is created immediate upon calling resolveObject. It is a deep clone of the object being resolved, to prevent any changes that are made during this process from trickling back outside the resolveObject scope. I think everything is ok as is...

chazzmoney:
LGTM! :shipit:

PR #336, opened 2019-10-10
ci: add support for node v10 & v12, remove v6

dishankmehta:
*Issue #, if available:*
302

*Description of changes:*
Removed support for node v6. Added node v10 & v12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
 node_js:
   - "8"
-  - "6"
+  - "10"
+  - "12"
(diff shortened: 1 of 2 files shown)


dbanksdesign:
Some data for this, Node v8 will end its long-term support by the end of this year. Node v6 is already past long-term support and AWS Lambda no longer supports v6 as a runtime. 

https://nodejs.org/en/about/releases/
https://nodejs.org/metrics/

dbanksdesign:
@dishankmehta we should probably update the engines attribute in the package.json file to show support for >= 8 rather than 6 as well.

dishankmehta:
@dbanksdesign Sure. On it.

dishankmehta:
Done. Updated the engines attribute in package.json for node >=8.

Issue #327, opened 2019-10-01
Referencing alias without generating


7studio:
Hi,

Is there a way to include/reference alias without generating entries into output files?

I don't find any solution into documentation, examples or different projects which use Style Dictionnary to load JSON and assign them like alias without `include` or `source`.

At this time, I manage to get what I need with the help of `filter` by excluding tokens with some `prop.path` but it's very untenable.

Thank you for you help.
Xavier.

dbanksdesign:
Right now no. Style Dictionary merges all the source files together into 1 big object before any alias  resolution happens. Then filtering is the way to remove tokens from the output files. If you have any ideas on how we can make this better we would love to hear it!

7studio:
Danny,

I think about two ideas :

1. **The simplier**: enhancing `prop` object into filter function. For the moment, the object looks like below:
```js
{ 
  value: '#fef1cc',
  original: { value: '#fef1cc' },
  name: 'SecondaryYellow200',
  attributes:
   { category: 'secondary',
     type: 'yellow',
     item: '200',
     subitem: undefined,
     state: undefined 
  },
  path: [ 'secondary', 'yellow', '200' ]
}
```
IMO, if we add the file path where the property comes from or add the fact that the prop is an include instead of a source (adding the two, will be really awesome :grin:) will help filter more correctly than it is possible for now.

2. **More complex**: this should require to publish a new version because it will change the actual behaviour. The idea is that the files added by the `include` attribute don't output properties. I don't understand the differences and needs about `include` and `source` attributes in fact :sweat_smile: All files and properties are deep merged all together.

I hope it will help you improve this project.
Xavier.

dbanksdesign:
Thank you so much for the suggestions. I like the first idea, especially adding if a token was in 'include' or 'source'. 

The thinking behind 'include' and 'source' is for creating parent-child style dictionaries. In my team at Amazon, we have a core style dictionary that both web and mobile inherit from, core is the parent and web and mobile are children. The mobile and web style dictionaries 'include' the tokens from core, and have their 'source' as any additions/overrides. One of the differences between 'include' and 'source' is that Style Dictionary will warn you (or throw an error depending on the log type) if 2 files found in 'source' would override parts of the object, but it won't if a file in 'source' overrides something in 'include' as that is what is intended. Hopefully that helps!

7studio:
Thank you for your detailed explanation :blush: I had not imagined this use for my project but it's really interesting.

[...]

PR #356, opened 2019-11-22
Add new entries on property object

7studio:
Hi,

This is the PR to fix the issue #327.

Finally, I choose to follow Danny's comment about naming and tests problem and publish this PR.
It is not perfect but it's a good try I think :wink:

I hope we will find the problem into `__tests__/extend.test.js` together because my brain is stuck :dizzy_face:

Thank you for your help and your encouragements.


lib/utils/combineJSON.js
```diff
@@ -48,6 +58,15 @@ function combineJSON(arr, deep, collision) {
       throw e;
     }
 
+    // Add some side data on each property to make filtering easier
+    traverseObj(file_content, (obj) => {
+      if (obj.value && !obj.filePath) {
```
dbanksdesign: I can't comment on lines you don't change, but we will need to change the line `file_content = require(resolvedPath);`. Because we are requiring the file rather than loading it with `fs.readFile`, any mutations we make to the object stay in memory, which is why some tests are failing and showing isSource as true when it should be false. We can fix this a few ways, first we could remove the `&& !obj.filePath` so that it will always override tokens. Although this could cause weird issues in the future because we are mutating a module. Or we could change a line above from:

```
    var file_content;
    try {
      file_content = require(resolvedPath);
```

to:

```
    var file_content = {};
    try {
      deepExtend([file_content, require(resolvedPath)]);
```

Now mutating `file_content` would not result in mutating the original module. Or there might be some more elegant solution to create a deep copy of a module.

dbanksdesign:
@7studio just checking in on this PR, have you had a chance to take a look at it? Let me know if you need any help!

7studio:
@dbanksdesign I am really sorry Danny, I fixed all the problems (all tests pass) before Christmas but I didn't take the time to push them. I rebased the branch with the last version of master and I hope I haven't forgotten something otherwise I will fix it again.

I would like to congratulate you and your partner for your baby 👶 Good luck!

7studio:
@dbanksdesign Have I missed something about the PR not being merged since January? If I should apply some modifications, let me know :relaxed:

Issue #307, opened 2019-07-08
Sass variables should (optionally?) have `!default` flag


nategreen:
In the `scss/map-deep` format, the variables it generates (for use in the map) have the `!default` flag:

```scss
// '!default' means if the variable is/was already defined, this declaration gets skipped
$my-var: #123ABC !default; 
$tokens: (
  my: (
    var: $my-var,
  )
);
```

…whereas the `scss/variables` format doesn't have that:

```scss
// No !default, so this variable can't be overridden during Sass compilation
$my-var: #123ABC;
```

Is there a reason why one has it and the other doesn't? To me, it seems like:

- These should agree…It's not necessarily a huge deal, but currently if we build both formats, downstream Sass compilations will either be able to override the values or not, based on which format they choose.
- `!default` should be added as an option so we can choose whether or not we want them to be able to override the variables from the configuration. (I guess that's a feature request. 😄)

dbanksdesign:
I like having `!default` as an optional configuration. Do you think it would make sense to do that at the file level or at the token level?

nategreen:
I was just picturing it being an option on the format in the config—beyond that, not sure. I guess the token level makes sense for overrides, not sure about file level…are there other options/features that work at the file level?

nategreen:
> so we can choose whether or not we want them to be able to override the variables from the configuration

Oh, I think I was multi-tasking when I wrote this, I just meant "so we can choose whether or not downstream users can override the variables," full stop. I don't know if I can conjure a reason we'd want the default flag on some tokens but not others, but it doesn't seem out of the question.

dbanksdesign:
We merged in #359 onto the 3.0 branch targeting the next major release. We will keep this issue open until the release. We want to see if this flag `themeable: true` on the token can be used in other built-in formats so that this is not SCSS specific. If anyone has ideas for this please open a PR or comment on this issue: #474 thanks!

dbanksdesign:
Closing this issue as this is now a feature in 3.0. You can get it today with `npm install style-dictionary@next`

PR #359, opened 2019-11-23
feat: add support for !default in SCSS variables format

jmmcduffie:
*Issue #, if available:* #307

*Description of changes:*
This addresses the feature request made in #307 in one way. For my use case I'd like to make some variables "themeable" with `!default` but not all. For example, base colors should be immutable but the default text color could be configurable by the consumer of the design tokens.

I chose the key "themeable" because of seeing that language in the design systems community (for example with [the Lightning Design System](https://lightningdesignsystem.com/design-tokens/) or from [Brad Frost](https://bradfrost.com/blog/post/creating-themeable-design-systems/)). And that way other formats could implement their own version of what theme-ability is.

This doesn't address setting `!default` on all variables through the configuration, but it should be easy enough to add that to what I've implemented.

Thanks in advance for considering this PR!

### DIFF
--- a/__tests__/formats/scssVariables.test.js
+++ b/__tests__/formats/scssVariables.test.js
@@ -63,7 +63,7 @@ describe('formats', () => {
       });
     });
 
-    it('should optionally use !default', done => {
+    it('should optionally use !default', () => {
       var themeableDictionary = _.cloneDeep(dictionary),
         formattedScss = formatter(dictionary),
         themeableScss = "";
@@ -74,7 +74,6 @@ describe('formats', () => {
       themeableScss = formatter(themeableDictionary);
 
       expect(themeableScss).toMatch("#EF5350 !default;");
-      return done();
     });
   });
 });
--- a/__tests__/formats/scssVariables.test.js
+++ b/__tests__/formats/scssVariables.test.js
@@ -13,6 +13,7 @@
 
 var formats = require('../../lib/common/formats');
 var scss = require('node-sass');
+var _ = require('lodash');
 
 var file = {
   "destination": "__output/",
@@ -62,5 +63,18 @@ describe('formats', () => {
       });
     });
 
+    it('should optionally use !default', done => {
+      var themeableDictionary = _.cloneDeep(dictionary),
+        formattedScss = formatter(dictionary),
+        themeableScss = "";
+
+      expect(formattedScss).not.toMatch("!default");
+
+      themeableDictionary.allProperties[0].themeable = true;
+      themeableScss = formatter(themeableDictionary);
+
+      expect(themeableScss).toMatch("#EF5350 !default;");
+      return done();
+    });
   });
 });
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -40,7 +40,14 @@ function fileHeader(options, commentStyle) {
 
 function variablesWithPrefix(prefix, properties, commentStyle) {
   return properties.map(function(prop) {
-      var to_ret_prop = prefix + prop.name + ': ' + (prop.attributes.category==='asset' ? '"'+prop.value+'"' : prop.value) + ';';
+      var to_ret_prop = prefix + prop.name + ': ';
+      to_ret_prop += (prop.attributes.category==='asset' ? '"'+prop.value+'"' : prop.value);
+
+      if (prop.themeable === true) {
+        to_ret_prop += ' !default';
+      }
+
+      to_ret_prop += ';';
 
       if (prop.comment) {
         if (commentStyle === 'short') {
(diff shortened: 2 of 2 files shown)


lib/common/formats.js
```diff
@@ -40,7 +40,14 @@ function fileHeader(options, commentStyle) {
 
 function variablesWithPrefix(prefix, properties, commentStyle) {
   return properties.map(function(prop) {
-      var to_ret_prop = prefix + prop.name + ': ' + (prop.attributes.category==='asset' ? '"'+prop.value+'"' : prop.value) + ';';
+      var to_ret_prop = prefix + prop.name + ': ';
+      to_ret_prop += (prop.attributes.category==='asset' ? '"'+prop.value+'"' : prop.value);
+
+      if (prop.themeable === true) {
+        to_ret_prop += ' !default';
```
dbanksdesign: This function, variablesWithPrefix, is used to generate less and CSS variables as well. Would this addition break those formats? We might need to rethink this function...

__tests__/formats/scssVariables.test.js
```diff
@@ -62,5 +63,18 @@ describe('formats', () => {
...
+        formattedScss = formatter(dictionary),
+        themeableScss = "";
+
+      expect(formattedScss).not.toMatch("!default");
+
+      themeableDictionary.allProperties[0].themeable = true;
+      themeableScss = formatter(themeableDictionary);
+
+      expect(themeableScss).toMatch("#EF5350 !default;");
+      return done();
```
dbanksdesign: We don't need `return done();` here, that is used if the test has some asynchronous code which we do for testing scss/less validity.

jmmcduffie: Good to know! I should have looked more closely at what that was for in the other test.

lib/common/formats.js
jmmcduffie: Great catch. I assumed it was just shared between the different Sass formatters so that's my bad.

It would definitely break both of those other formats without some more work. I could add an additional argument to this function to output Sass-specific properties like `!default`. Technically the "short" comment version is only valid in Sass and Less, as native CSS only supports `/* ... */`.

__tests__/formats/scssVariables.test.js
jmmcduffie: This has been resolved.

lib/common/formats.js
jmmcduffie: I took a stab at updating the `variablesWithPrefix` function to encapsulate the logic around prefixes, comment formats and `!default`. Let me know what you think!

jmmcduffie:
Awesome! Thank you so much for reviewing this.

kevinmpowell:
+1 for this feature. I'm just getting started w/style dictionary. Good to see @jmmcduffie making it better. Nice work!

jmmcduffie:
Thanks, @kevinmpowell! I'll be curious to hear what you think of Style Dictionary.

chazzmoney:
I think this is a really good direction and important functionality. However, I have two concerns:

1. The themeable property is only applicable to one format. I think it might have general use in languages or spaces where a default syntax exists. If we can find other places (especially with currently supported languages) where we can use this to generate syntactically appropriate defaults, I think that would be a very large value to the commit.
2. I'm worried about adding this without doing the previous. I realize this is a lot of extra work and maybe even outside your/our specialties, but we risk polluting the property space with single-format relevant properties. If we can aim to ensure property relevance to all formats, that would be preferred.

Thoughts?

jmmcduffie:
Your concerns make a ton of sense, @chazzmoney.

My hope for this new property was exactly what you suggested in your first point, that other platforms would be able to use the same data with a different implementation. By giving it an open-ended name, my thought was that it would not be viewed as a single-format property and allow new use cases to arise.

But I totally get your worry about moving forward without knowing if that will really happen. And I also don't disagree that a better understanding of that would add value to this PR. Do you think that we could still provide that value by doing just enough research to validate that another format could use this?

My concern in lumping them together would be that we bloat the scope of this change. And if we determine that it truly is a single-format property, we can course-correct my implementation into something more appropriate.

chazzmoney:
For those who are looking for someplace to help out - if you know if there are tags for defaults in [the main library formats](https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L90), we would love to have them added to this.

dbanksdesign:
We added this PR to the 3.0 branch targeting the next major release (later this year) because it is adding something to the core token object itself. Before releasing 3.0 we want to see how this flag can work in other built-in formats and write some documentation about it.

lib/common/formats.js
afebbraro: Hey all, I'm working on adding a custom transform for my DS for this feature and found an edge case with my stuff that I thought I would share because It looks like you might run into the same issue here with the current code. The edge case is when I have this:
```
// token json file
 "content-padding": {
      "comment": "The padding around the main content icon and message in the Alert component.",
      "value": "{space.m.value} {space.m.value} {space.m.value} {space.misc-a.value}",
      "type": "settings",
      "themable": true
    },
```
It outputs this: 
```
$sprk-alert-content-padding: 16px !default 16px !default 16px !default 24px !default; // The padding around the main content icon and message in the Alert component.

The `!default` gets repeated unless we check for that case and make sure not to repeat it. Thought this might be helpful to y'all! Hope you are having a great week!

Issue #208, opened 2018-11-30
Process references first / at the same time as transforms


chazzmoney:
https://github.com/amzn/style-dictionary/pull/198#issuecomment-441364947
https://github.com/amzn/style-dictionary/issues/120#issuecomment-418522257

We've had a couple issues come up in regard to handling references/aliases before transforms (or at the same time). We also are looking at #205, which would require this change (depending on what we choose to do).

chazzmoney:
Adding to 3.0 as this could be considered a breaking change.

lancetharper:
I have found this somewhat counter-intuitive when first starting out with SD. It would go a long way toward increasing flexibility of token creation if you don't follow strict CTI (me! 😄).

Loving SD but this is my biggest sticking point so far. Hope to see this land soon!

lancetharper:
@didoo Yes! I've read your article and haven't had a problem ignoring the CTI structure; I've just re-wired it to abstract it out of naming.

I just want to be able to do transforms on referenced/aliased tokens

mhewson:
as an example, It would be great maintain alpha channel values separate from foundational color tokens and combine these underlaying values into tokens

shaunbent:
I have just ran into this problem by trying to exactly what @mhewson is describing above.

Do you have a clear plan of what you intend to implement @chazzmoney? I'd be happy to try and help out with this if I can find sometime.

chazzmoney:
@shaunbent thanks for bringing this back up, I think this is actually something we need to think about soon.

I don't have a clear plan. [The way SD works is pretty step by step currently](https://amzn.github.io/style-dictionary/#/build_process); I think there would be some heavy changes involved (and likely breaking) to get everything to run in a new order or all at the same time.

If you find some time, i would _LOVE_ to see someone outside take a stab at it (diversity of thinking results in better solutions). I'm happy to answer any questions as they come up!

mfal:
I really like the idea of creating new tokens from other tokens by transforms (like color transformation e.g. with polished).

#371 enabels this feature and now is ready to be reviewed!

chazzmoney:
Just wanted to put a huge acknowledgment of @mfal

Your diverse thinking and hard work created a more flexible (and better) solution. Your contribution solved a variety of issues people were running into trying to have SD work for their use cases. You moved SD forward in a really significant way that will impact the ease of managing styles in code.

THANK YOU!

PR #371, opened 2020-01-10
Make transitive transforms & resolves possible

mfal:
Issue #208 

Transformations can only be executed, if the value to be transformed, has no references in it. So resolving may lead to enable further transformations, and sub sequent resolving may enable even more transformations - and so on.

Unfortunately the current implementation only transforms and resolves once, and transformation on values containing references *do not work*. This change enables this transitive dependencies, by executing "transform & resolve" until all transformations are executed.

Take the following example (the `color: {...}` object might setup your custom color transformation)

```
color.brand = {
  value: "{color.base.green}"
}

color.background.button.primary.base = {
  value: "{color.brand.value}",
  color: {
    desaturate: 0.5
  }
}

color.background.button.primary.hover = {
  value: "{color.background.button.primary.base}",
  color: {
    darken: 0.2
  }
}
```

As you can see `color.background.button.primary.hover` is a variation of `color.background.button.primary.base` which is a variation of `color.base.green`. This transitive dependencies are solved by running this loop.

### DIFF
--- a/__tests__/__properties/colors.json
+++ b/__tests__/__properties/colors.json
@@ -85,6 +85,16 @@
 
     "font": {
       "link": { "value": "{color.base.blue.100.value}" }
+    },
+
+    "button": {
+      "base": { "value": "{color.base.blue.100.value}" },
+      "active": { "value": "{color.button.base.value}", "transformColor": {
+        "darken": 0.5
+      } },
+      "hover": { "value": "{color.button.active.value}", "transformColor": {
+        "darken": 0.2
+      } }
     }
   }
 }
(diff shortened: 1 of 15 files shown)


dbanksdesign:
Thank you for submitting this! This is a pretty big PR, it'll take some time to review. Please have some patience on this.

dbanksdesign:
Hey @mfal ! Sorry it took so long to get to this, it has been a busy few months. I finally dug into this PR and we are super excited to get this into the next major release (3.0). I made some changes/proposals and created a pull request on your branch, https://github.com/mfal/style-dictionary/pull/1. Let me know what you think and when that gets merged in, this PR should automatically update and we can merge it in. Let me know if you have any questions/concerns. 

Thank you so much for this contribution!

PR #372, opened 2020-01-13
Adds prefix option for scss/map-deep format

canvasplay:
*Description of changes:*

This PR adds an additional option `prefix` for the `scss/map-deep`.
Similar to the already existing option `mapName" the `prefix` option adds the ability to also customise the variable names.
I found this useful to prevent variable name conflicts when importing the tokens into existing codebases.

With this change one could get an output like this:
```
$token-color-white: #FFFFFF;
$token-font-weight-bold: 600;
tokens: (...)
```
using the prefix like this:
```
{
  destination: 'tokens.scss',
  format: 'scss/map-deep',
  mapName: 'tokens',
  prefix: 'token'
}
```

What do you think?


canvasplay:
I see tests are failing. I can have a look if the idea is accepted.

dbanksdesign:
We already do allow prefixes, but they are in the name transforms:
https://github.com/amzn/style-dictionary/blob/master/lib/common/transforms.js#L150


The only difference would be you need to add the "prefix" attribute to the platform object rather than the file object. Would this solve your case, or do you think something different is needed?

canvasplay:
Using the platform prefix works great! It is covering my needs so far!
Thank you.

PR #383, opened 2020-02-07
feat: adding color/hsl and color/hsl-4 transforms

cssinate:
*Issue #, if available:*

*Description of changes:*
Adding in support for color transforms to color/hsla (going to to 3-argument hsl format if no alpha change) and color/hsl with newer syntax, accepting 4 arguments in the same function.

### DIFF
--- a/__tests__/common/transforms.test.js
+++ b/__tests__/common/transforms.test.js
@@ -281,6 +281,38 @@ describe('common', () => {
       });
     });
 
+    describe('color/hsl-4', () => {
+      it('should handle normal colors', () => {
+        var value = transforms["color/hsl-4"].transformer({
+          value: "#009688"
+        });
+        expect(value).toBe("hsl(174 100% 29%)");
+      });
+
+      it('should handle colors with transparency', () => {
+        var value = transforms["color/hsl-4"].transformer({
+          value: "#00968899"
+        });
+        expect(value).toBe("hsl(174 100% 29% / 0.6)");
+      });
+    });
+
+    describe('color/hsl', () => {
+      it('should handle normal colors', () => {
+        var value = transforms["color/hsl"].transformer({
+          value: "#009688"
+        });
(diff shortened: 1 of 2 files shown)


lib/common/transforms.js
```diff
@@ -287,6 +287,53 @@ module.exports = {
...
+  },
+
+    /**
+   * Transforms the value into an HSL string, using fourth argument if alpha is present.
+   *
+   * ```js
+   * // Matches: prop.attributes.category === 'color'
+   * // Returns:
+   * "hsl(174 100% 29%)"
+   * "hsl(174 100% 29% / .5)"
```
dbanksdesign: This might be a dumb question, what is the difference in the formatting of this output versus the 'color/hsla' transform? Do both work and is one preferable over the other? Both transforms seem to handle alpha and both output `hsl()`, do we need both?

__tests__/common/transforms.test.js
```diff
@@ -281,6 +281,38 @@ describe('common', () => {
       });
     });
 
+    describe('color/hsl', () => {
```
dbanksdesign: Yes! Unit tests! 🙌

cssinate:
The hsla transform has better browser support, but the hsl transform uses
the CSS color module level 4 color spec syntax. It's analogous to wanting
to use a newer JavaScript syntax like Array.map or something.

https://www.w3.org/TR/css-color-4/#the-hsl-notation

dbanksdesign:
Sorry just getting back in the swing of things. I was wondering since both transforms handle transparency and the only difference is which CSS color module spec syntax they use, would it make sense to rename the transforms to something like `color/hsl` and `color/hsl-4` or something?

cssinate:
Perhaps `hsla` and `hsl-4`?

dbanksdesign:
That makes sense to me, what do you think?

cssinate:
Let's do it

chazzmoney:
LGTM :shipit:

Issue #298, opened 2019-06-19
Invalid scss/map-flat when using strings


siiron:
New to Style Dictionary here.

I'm trying to convert some font styles from json to scss using `scss/map-flat`. I'm especially having trouble with font-family where my webfont Montserrat needs to be in single quotes in the finished scss map.

My config is pretty simple:

```json
{
    "source": ["src/properties/**/*.json"],
    "platforms": {
        "scss": {
            "transformGroup": "scss",
            "buildPath": "dist/scss/",
            "files": [              
                {
                    "format": "scss/map-flat",
                    "destination": "_font.map.scss",
                    "filter": {
                        "attributes": {
                            "category": "font"
                        }
                    }
                }
            ]
        }
    }
}
```

My `font.json` looks like this:

```json
{
    "font": {
        "family": {
            "brand": { "value": "'Montserrat', Arial, sans-serif" }
        },
        "style": {
            "normal": { "value": "normal" },
            "italic": { "value": "italic" }
        },
        "weight": {
            "regular": { "value": "regular" },
            "medium": { "value": "medium" },
            "bold": { "value": "bold" }
        }
    }
}
```

After transform I get a `_font.map.scss` that does not validate since the font-family stack consists of three values not encapsulated in quotes:
```scss
$tokens: (
  'font-family-brand': 'Montserrat', Arial, sans-serif,
  'font-style-normal': normal,
  'font-style-italic': italic,
  'font-weight-regular': regular,
  'font-weight-medium': medium,
  'font-weight-bold': bold
);
```

If I manually encapsulate the values in the map in `()`, it validates:

```scss
$tokens: (
  'font-family-brand': ('Montserrat', Arial, sans-serif),
  'font-style-normal': (normal),
  'font-style-italic': (italic),
  'font-weight-regular': (regular),
  'font-weight-medium': (medium),
  'font-weight-bold': (bold)
);
```

How can I do this using Style Dictionary? Do I need a custom template or can I somehow escape the font-family string in the json?

dbanksdesign:
Hmm looks like a bug on our end for not handling font-family properly in that format. I'll do some investigation.. In the meantime, you could override the default [sass/map-flat format](https://github.com/amzn/style-dictionary/blob/master/lib/common/templates/scss/map-flat.template)

```js
const StyleDictionary = require('style-dictionary');
const _ = require('losdash');
const fs = require('fs');

StyleDictionary.registerFormat({
  name: 'scss/map-flat',
[...]

PR #394, opened 2020-02-27
fix(formats): align scss/map-* with scss/variables on asset category

7studio:
*Issue #, if available:*

A part of #298 :blush:

*Description of changes:*

To be more in accordance with the expected output in the same language (Scss), it is important that the behaviours of all formats try to follow the same way.

Now, `scss/map-*` formats handle the "asset" category like `scss/variables` and wrap the values in a double-quoted string.

### DIFF
--- a/__tests__/formats/__snapshots__/scssMaps.test.js.snap
+++ b/__tests__/formats/__snapshots__/scssMaps.test.js.snap
@@ -11,6 +11,7 @@ $size-font-small: 12rem !default;
 $size-font-large: 18rem !default;
 $color-base-red: #ff0000 !default; // comment
 $color-white: #ffffff !default;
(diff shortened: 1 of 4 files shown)


__tests__/formats/__snapshots__/scssMaps.test.js.snap
```diff
@@ -11,6 +11,7 @@ $size-font-small: 12rem !default;
 $size-font-large: 18rem !default;
 $color-base-red: #ff0000 !default; // comment
 $color-white: #ffffff !default;
+$asset-icon-book: \\"PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0IiBmaWxsPSJub25lIiBzdHJva2U9ImN1cnJlbnRDb2xvciIgc3Ryb2tlLXdpZHRoPSIyIiBzdHJva2UtbGluZWNhcD0icm91bmQiIHN0cm9rZS1saW5lam9pbj0icm91bmQiIGNsYXNzPSJmZWF0aGVyIGZlYXRoZXItYm9vayI+PHBhdGggZD0iTTQgMTkuNUEyLjUgMi41IDAgMCAxIDYuNSAxN0gyMCI+PC9wYXRoPjxwYXRoIGQ9Ik02LjUgMkgyMHYyMEg2LjVBMi41IDIuNSAwIDAgMSA0IDE5LjV2LTE1QTIuNSAyLjUgMCAwIDEgNi41IDJ6Ij48L3BhdGg+PC9zdmc+\\" !default;
```
dbanksdesign: Is the double slash here correct?

dbanksdesign: Pinging this, @7studio is this valid scss?

7studio: @danielstocks It's a valid SCSS syntax but I don't know why I need to use the double `\` :disappointed: to pass the tests. If I remove them, some tests fail... The first `\` escapes the `"` and the second one escapes the first `\`. This trick/hack seems to be needed by the JS test engine.

dbanksdesign: Ohhhhh I think it has to do with our snapshot tests... it is escaping the quote in JS, and NOT in actual output SCSS. --____--

PR #415, opened 2020-04-04
feat(attribute/cti): make attribute/cti respect manually set attributes

tonyjwalt:
*Issue #, if available:* 414

*Description of changes:*
- Make attribute/cti not overwrite existing attributes
- Only build out cti attributes to the length of the path
- Add tests for attribute/cti

### DIFF
--- a/lib/common/transforms.js
+++ b/lib/common/transforms.js
@@ -76,9 +76,9 @@ module.exports = {
       const originalAttrs = prop.attributes || {};
       const generatedAttrs =  {}
 
-      prop.path.forEach((leaf, i) => {
-        generatedAttrs[attrNames[i]] = leaf;
-      });
+      for(let i=0; i<prop.path.length && i<attrNames.length; i++) {
+        generatedAttrs[attrNames[i]] = prop.path[i];
+      }
 
       return Object.assign(generatedAttrs, originalAttrs);
     }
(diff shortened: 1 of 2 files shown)


lib/common/transforms.js
```diff
@@ -71,13 +71,16 @@ module.exports = {
...
-        state:    prop.path[4]
-      }
+
+      const attrNames = ['category', 'type', 'item', 'subitem', 'state'];
+      const originalAttrs = prop.attributes || {};
+      const generatedAttrs =  {}
+
+      prop.path.forEach((leaf, i) => {
+        generatedAttrs[attrNames[i]] = leaf;
+      });
```
chazzmoney, line 81: if prop.path.length > attrNames.length, then line 80 will set generatedAttrs[undefined] to the last leaf value in prop.path.

This would be new behavior. Maybe the loop should be forced to only run a maximum of 5 times?

chazzmoney:
Hi Tony,

Thank you for this! I agree that this would be a better situation than our current setup.

Can you check out the comment I posted? Its just a thought about managing the length of the two arrays being used together.

Thanks!

lib/common/transforms.js
tonyjwalt: Great catch. I updated the loop to terminate if either array runs out of values so that you can change the length of the attrNames array in the future. I'll commit that now.

chazzmoney: Looks great!

chazzmoney:
LGTM :shipit:

PR #419, opened 2020-04-27
Extended path trailing slash check to also work on Windows

JDansercoer:
On Windows, the passed in paths don't end with a forward slash, but instead with a backward slash. This causes `styled-dictionary` to fail when building on Windows. New check is less strict, and also allows the path to end in a backslash.

### DIFF
--- a/lib/buildFiles.js
+++ b/lib/buildFiles.js
@@ -25,7 +25,7 @@ var _ = require('lodash'),
  * @returns {null}
  */
 function buildFiles(dictionary, platform) {
-  if (platform.buildPath && platform.buildPath.slice(-1) !== '/') {
+  if (platform.buildPath && (platform.buildPath.slice(-1) !== '/' || platform.buildPath.slice(-1) !== '\')) {
     throw new Error('Build path must end in a trailing slash or you will get weird file names.')
   }
 
--- a/lib/buildFiles.js
+++ b/lib/buildFiles.js
@@ -25,7 +25,7 @@ var _ = require('lodash'),
  * @returns {null}
  */
 function buildFiles(dictionary, platform) {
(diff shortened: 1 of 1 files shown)


lib/buildFiles.js
```diff
@@ -25,7 +25,7 @@ var _ = require('lodash'),
  * @returns {null}
  */
 function buildFiles(dictionary, platform) {
-  if (platform.buildPath && platform.buildPath.slice(-1) !== '/') {
+  if (platform.buildPath && (platform.buildPath.slice(-1) !== '/' || platform.buildPath.slice(-1) !== '\\')) {
```
dbanksdesign, line 28: I think the logic should be `&&` instead of `||` because the positive case is the one that throws the error (so the logic is a bit backwards, the positive case is the error case).

chazzmoney:
@JDansercoer if you can update based on @dbanksdesign comments before May 22nd, we can get this into the last 2.x release of StyleDictionary.

We are starting 3.0 branch and otherwise this may have to wait until we get some other large changes in.

chazzmoney:
Hi @JDansercoer,

Can you modify the logic so we can merge this in? Thank you!

chazzmoney:
Hi @JDansercoer,

Would like to merge in this PR but need the change mentioned by DBanksDesign. Let us know if you are still working on this or need help.

Thanks!

JDansercoer:
Sorry for the late reply on this guys, but I lost track of this.

I added the check that was proposed, and hope that that fixes the failing tests.

If not, I'm not sure when I'll have time to thoroughly check it.

chazzmoney:
Thank you @JDansercoer - Looks great!

:shipit:

dbanksdesign (approved):
LGTM! :shipit:

PR #423, opened 2020-05-15
bugfix: fix conditional to ensure we catch properties with a falsy value

### LINKED ISSUE #406: Values of number zero are always filtered if filter is set
Hi,

For a token defined with value zero, for example `opacity: {value: 0}`, and any `filter` function set in config it appears that token is always filtered out.

I believe the cause is https://github.com/amzn/style-dictionary/blob/master/lib/filterProperties.js#L37, which perhaps should be `} else if (typeof value.value !== 'undefined') {`

This line only kicks in if a filter is set in config, as the entire filtering check is bypassed with no filter config.

Thoughts?

chazzmoney:
Fixes #406


__tests__/__output/json-nested.json
```diff
@@ -0,0 +1,12 @@
...
+    "base": {
+      "red": {
+        "primary": "#611D1C",
+        "secondary": {
+          "inverse": "#000000"
+        }
+      }
+    }
+  }
+}
```
chazzmoney, line 12: I'm not sure where this came from... Is this ok @dbanksdesign ?

dbanksdesign: I think there might be a race condition in our unit tests somewhere.. there shouldn't be anything checked into **__tests__/__output/** that is the folder where we put files generated by the unit tests. It should get cleared after tests run. You can remove this file or just try running the tests again and see if it gets removed.

Issue #370, opened 2020-01-08
Pass `file` as an argument in formatter


dbanksdesign:
Based on the conversation in #363 it would probably make more sent to pass the `file` object as an argument in the formatter function rather than bind `this` to it. It could be done in a backwards-compatible way by sending one more argument and keeping the binding in place.

dbanksdesign:
The code is merged in, we just need to update the documentation in the 3.0 branch to remove the mention of `this` and instead use the file object.

dbanksdesign:
Closing this as it is in 3.0

PR #439, opened 2020-06-19
feat(formats): add file object to formatter method

dbanksdesign:
*Issue #, if available:* #370 

*Description of changes:* Passing the file object to the formatter function so it can be accessed as an argument rather than being bound to `this`. I kept the binding as-is to ensure backwards compatibility though. 


chazzmoney:
I'm good with this. It does change a couple function signatures, but they are all internal. Do you think this should be a 3.0 thing?

dbanksdesign:
Yea this should be 3.0 I think.

chazzmoney:
LGTM! :shipit:

PR #454, opened 2020-08-23
fix: update `style-dictionary clean` config path

tehkaiyu:
`style-dictionary build` and `style-dictionary clean` should use the same default config paths

*Description of changes:*
Ran into an issue where running `style-dictionary build` worked but `style-dictionary clean` threw an error because I was using config.js instead of config.json

### DIFF
--- a/bin/style-dictionary
+++ b/bin/style-dictionary
@@ -14,6 +14,25 @@ function collect(val, arr) {
   return arr;
 }
 
+function getConfigPath(options) {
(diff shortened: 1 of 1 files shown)


chazzmoney:
This looks like a solid bugfix to me - @dbanksdesign is there a reason that the clean would want to use a different method to get configs?

dbanksdesign (approved):
LGTM! :shipit:

Issue #464, opened 2020-09-28
ERROR: Maximum call stack size exceeded when using json/nested


lukasoppermann:
Hey, I am getting the following error when trying to transform my json file. I am a bit at a loss what to do about it. 😅

Please let me know how I can provide some more helpful information.

> /Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:75
  if (obj.hasOwnProperty('value')) {
          ^

RangeError: Maximum call stack size exceeded
    at String.hasOwnProperty (<anonymous>)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:75:11)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:80:23)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:80:23)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:80:23)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:80:23)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:80:23)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:80:23)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:80:23)
    at minifyDictionary (/Users/lukasoppermann/Repos/design-token-transformer/node_modules/style-dictionary/lib/common/formats.js:80:23)

chazzmoney:
@lukasoppermann 
Looks like there may be an infinite loop. Are you running vanilla SD or are you using some of the custom work you've built?

What would enable us to help:
1. The config you are using
2. Any custom code you've built
3. The source dictionaries

lukasoppermann:
I think I have found what causes it, in the json below the `comment` key on line 3 (as a child of the h3) causes the issue. Removing this makes it work. However having additional comments & meta info on a group is important for documenting the system.

```json5
"body": {
    "h3": {
[...]

PR #465, opened 2020-09-29
fix(formats): fix max call stack issue on json/nested format

dbanksdesign:
*Issue #, if available:* #464 

*Description of changes:* Fixing max call stack on `json/nested` format when the dictionary object contains non-token data. The issue was we were using `for (var name in obj)` assuming `obj` was an object, but if it was a string it would get in an infinite loop. 

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -71,7 +71,12 @@ function iconsWithPrefix(prefix, properties, config) {
 }
 
 function minifyDictionary(obj) {
+  if (typeof obj !== 'object' || Array.isArray(obj)) {
+    return obj;
+  }
+
   var toRet = {};
+
   if (obj.hasOwnProperty('value')) {
     return obj.value;
   } else {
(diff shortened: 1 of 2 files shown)


lib/common/formats.js
```diff
@@ -784,14 +789,14 @@ module.exports = {
    * @example
    * ```dart
    * import 'package:flutter/material.dart';
-   *
```
chazzmoney, line 787: 👀

lib/common/formats.js
```diff
@@ -784,14 +789,14 @@ module.exports = {
    * @example
    * ```dart
    * import 'package:flutter/material.dart';
-   * 
+   *
    * class StyleDictionary {
    *   StyleDictionary._();
-   *
```
chazzmoney, line 790: 👀

lib/common/formats.js
```diff
@@ -784,14 +789,14 @@ module.exports = {
...
-   * 
+   *
    * class StyleDictionary {
    *   StyleDictionary._();
-   *   
+   *
    *     static const colorBrandPrimary = Color(0x00ff5fff);
    *     static const sizeFontSizeMedium = 16.00;
    *     static const contentFontFamily1 = "NewJune";
-   * ```
```
chazzmoney, line 794: 👀

chazzmoney:
LGTM - and you made some whitespace modifications. I'm totally good merging as is, but wanted to make sure you saw them.

:shipit:

dbanksdesign:
I did not see those whitespace changes... thanks for bringing them up. Let me double check them, maybe it was prettier doing something...

dbanksdesign:
Ah I have the prettier plugin on VSCode and it was removing trailing whitespace. So it's a change we probably should make, but we should also probably have prettier as a dev dependency and have it run as a pre-commit hook or something.

PR #467, opened 2020-09-30
feat(formats): add javascript/es6-ts-declarations formats for .d.ts

### LINKED ISSUE #409: Provide TypeScript typings

### LINKED ISSUE #425: Declaring TypeScript types for design tokens that use the ES6 module format
I'm exporting my design tokens as ES6 modules and I'm having troubles figuring out a way to type these. Currently, I type them with a `module.d.ts` file as such:

tokens.js
```ts
export const ColorWhite = "#ffffff";
export const ColorBlack = "#222222";
...
```

module.d.ts
```ts
declare module 'design-tokens' {
  export const ColorWhite: string;
  export const ColorBlack:string;
  ...
}
```

As the tokens are generated automatically and a lot of tokens will be added in the future it's kinda tedious. I haven't seen a way to export type declarations atm and also [this issue](https://github.com/amzn/style-dictionary/issues/409) doesn't appear to cover my use case.

Has anyone had the same problem and knows a way around it?

Maybe there is a TypeScript way of declaring the module's exports dynamically? I'm thinking something like this:

```ts
declare module 'design-tokens' {
  export const [token]: string;
}
```

tychi-nflx:
Possibly resolves #425 

*Description of changes:*
Adds a format for exporting es6-ts-declarations

Usage (config.json):
```
...
  {
    "format": "javascript/es6-ts-declarations",
    "destination": "colors.d.ts"
  }
...
```

Output (colors.d.ts):
```
export const ColorBackgroundBase : string;
export const ColorBackgroundAlt : string;
```

### DIFF
--- a/__tests__/formats/__snapshots__/all.test.js.snap
+++ b/__tests__/formats/__snapshots__/all.test.js.snap
@@ -414,6 +414,15 @@ exports[`formats all should match javascript/es6 snapshot 1`] = `
 export const color_red = \\"#FF0000\\"; // comment"
 `;
 
+exports[`formats all should match javascript/es6-ts-declarations snapshot 1`] = `
+"/**
+ * Do not edit directly
+ * Generated on Sat, 01 Jan 2000 00:00:00 GMT
+ */
+
+export const color_red : string; // comment"
+`;
+
 exports[`formats all should match javascript/module snapshot 1`] = `
 "/**
  * Do not edit directly
--- a/__tests__/formats/es6TypeScriptDeclarations.test.js
+++ b/__tests__/formats/es6TypeScriptDeclarations.test.js
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+ * the License. A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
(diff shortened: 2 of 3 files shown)


chazzmoney:
Thank you for the contribution!

Two thoughts:
1. Usually in the test, (where possible) we like to actually execute the library and attempt to import the styles. This could be doable in this case potentially, but I leave it to you. Alternatively, executing a test against the expected string literal for each of the styles listed could be a good alternative to the RegEx technique, especially regarding the next thought...
2. Would all styles be of type 'string'? Maybe this makes sense, but it seems like maybe different style types might have different TS types. Maybe this is fine as is for a first step and can be changed later as usage develops.

chazzmoney:
Hi @tychi-nflx,

We are looking at merging this but would like to hear your thoughts before we do.

Thanks!

tychi-nflx:
Hey @chazzmoney! Sorry for the delay, I ended up switching teams in the middle of the last month, so I'm not actually doing any further work on the design system I was helping out with.

Anything you decide, I'll be happy with :)

jimmyandrade (approved):

ojassvi:
Any updates on this when will this be merged to the base branch.

calvinf:
Functionally, this looks good to me. Semantically, I'd recommend naming the format as "typescript/declarations" as it is specifically for the TypeScript language (ES is up to es2020 now and this really doesn't have anything in it that's specific to es6).  Other then that though, it would be great to have this merged as it would simplify TS declaration generation and avoids the need for creating one-off custom format for my project (or others who need to do the same thing).

chazzmoney:
@tychi-nflx can you change the format to "typescript/declarations"?

Thanks!

chazzmoney:
Hi @tychi-nflx - if you have a quick second to just update the name to "typescript/declarations", then we can get it merged.

Thank you!

dbanksdesign:
Closing this in favor of #557 Thank you!

PR #484, opened 2020-11-23
chore(export platform): cleaning up transitive transform code

dbanksdesign:
*Issue #, if available:*

*Description of changes:* I forgot to add this change to #476... This change will reduce the number of passes the loop that transforms and resolves does, the old code ran too many times (no effect on output, just unnecessary). Also better while loop logic so we don't have to disable eslint on that line. 


chazzmoney:
LGTM :shipit:

PR #486, opened 2020-11-25
chore(tests): add eslint-jst and fix lint errors

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Adding eslint-jest so we can better lint our tests. Also fixing the errors and warnings that came up from linting. 


__tests__/formats/scssMaps.test.js
```diff
@@ -207,7 +207,7 @@ var dictionary = {
 describe('formats', () => {
   _.each(['scss/map-flat', 'scss/map-deep'], function(key) {
 
-    describe(key, () => {
+    describe(`${key}`, () => {
```
chazzmoney, line 210: ?

chazzmoney:
Minus that one template string that I don't understand, LGTM.

__tests__/formats/scssMaps.test.js
dbanksdesign: This might be a rule we could disable... There is a lint rule that `describe` blocks must have a string as the title and the linter isn't smart enough to know that those variables are strings...

PR #487, opened 2020-11-25
docs(example): adding transitive transform example

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Adding a transitive transform example. Also, tweaking the usesReference method to check if the value is an object, check the object's attributes for references as well. 


chazzmoney:
Huge 👍 from me.

PR #488, opened 2020-11-25
docs: add page about upcoming release

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Adding a page in the docs about the upcoming release.

### DIFF
--- a/docs/version_3.md
+++ b/docs/version_3.md
@@ -1,0 +1,136 @@
+# Version 3.0!
+
+We have been working hard the past few months on a number of features and improvements we wanted to combine into a major release. Our aim is that is next version is backwards compatible, we thought it was a good idea to move to a new major version because we are changing a core part of how Style Dictionary works. 
+
+If you would like to get a pre-release of 3.0, use the `next` tag to install Style Dictionary:
+
+`npm install --save-dev style-dictionary@next`
+
+If you find any bugs or issues, please file tickets so we can get things fixed before we release it to the wider community. Thanks!
+
+### Transitive transforms
+
+This is the big one that required a big re-architecture of how the Style Dictionary build process works. 
+
+Up until now the build process looked like this: https://amzn.github.io/style-dictionary/#/build_process
+After merging all of the token files, it would go through the merged object and transform tokens it found, but only do value transforms on tokens that did not reference another token. The original intent here was that if you have a color for example, the value of it should be the same for all references of it, so we only need to do a value transform once. Then after all tokens are transformed, resolve all aliases/references. 
+
+However, we heard from the community there were a number of reasons why someone might want to transform the value of an aliased token. 
+
+* https://github.com/amzn/style-dictionary/issues/451
+* https://github.com/amzn/style-dictionary/issues/452
+* https://github.com/amzn/style-dictionary/issues/208
+
+The new build process is similar, except that it does multiple passes of transforming and resolving aliases and defers transforms if the token has an unresolved alias. Each pass might reveal tokens that are ready to be transformed in the next pass. Take this example:
+
+```json
+{
+  "color": {
+    "black": { "value": "#000000" },
+    "font": {
+      "primary": { "value": "{color.black.value}" },
+      "input": { "value": "{color.font.primary.value}" }
+    }
+  }
+}
+```
+
+The first pass will transform the `color.black` token because it doesn't have a reference. It will defer the transforms of `color.font.primary` and `color.font.input` because they do have references. Then it will resolve references to `color.black.value` because it has been transformed.
+
+The second pass will transform `color.font.primary` because it no longer has a reference. It will then resolve the reference to `color.font.primary.value` because it has been transformed. 
+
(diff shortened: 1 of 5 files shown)


docs/version_3.md
```diff
@@ -0,0 +1,136 @@
+# Version 3.0!
+
+We have been working hard the past few months on a number of features and improvements we wanted to combine into a major release. Our aim is that is next version is backwards compatible, we thought it was a good idea to move to a new major version because we are changing a core part of how Style Dictionary works.
```
chazzmoney, line 3: Our aim is that is next version is -> Though we intend that 3.0 to be

docs/version_3.md
```diff
@@ -0,0 +1,136 @@
...
+`npm install --save-dev style-dictionary@next`
+
+If you find any bugs or issues, please file tickets so we can get things fixed before we release it to the wider community. Thanks!
+
+### Transitive transforms
+
+This is the big one that required a big re-architecture of how the Style Dictionary build process works. 
+
+Up until now the build process looked like this: https://amzn.github.io/style-dictionary/#/build_process
+After merging all of the token files, it would go through the merged object and transform tokens it found, but only do value transforms on tokens that did not reference another token. The original intent here was that if you have a color for example, the value of it should be the same for all references of it, so we only need to do a value transform once. Then after all tokens are transformed, resolve all aliases/references.
```
chazzmoney, line 16: go through -> iterate through

The original intent here was that if you have a color for example, the value of it should be the same for all references of it, so we only need to do a value transform once. -> The original intent here was that a value of any reference should be the same for all references of it, so we only need to do a value transform once.

(Put any example at the end after the description)

docs/version_3.md
```diff
@@ -0,0 +1,136 @@
...
+Up until now the build process looked like this: https://amzn.github.io/style-dictionary/#/build_process
+After merging all of the token files, it would go through the merged object and transform tokens it found, but only do value transforms on tokens that did not reference another token. The original intent here was that if you have a color for example, the value of it should be the same for all references of it, so we only need to do a value transform once. Then after all tokens are transformed, resolve all aliases/references. 
+
+However, we heard from the community there were a number of reasons why someone might want to transform the value of an aliased token. 
+
+* https://github.com/amzn/style-dictionary/issues/451
+* https://github.com/amzn/style-dictionary/issues/452
+* https://github.com/amzn/style-dictionary/issues/208
+
+The new build process is similar, except that it does multiple passes of transforming and resolving aliases and defers transforms if the token has an unresolved alias. Each pass might reveal tokens that are ready to be transformed in the next pass. Take this example:
```
chazzmoney, line 24: The new build process is similar, except that it does multiple passes of transforming and resolving aliases and defers transforms if the token has an unresolved alias. -> The new build process is similar, except that it recursively transforms and resolves aliases, only deferring a transform to the next cycle if the token has an unresolved alias.

(Lets be explicit. I think this is accurate. If not, lets correct it.)

docs/version_3.md
```diff
@@ -0,0 +1,136 @@
...
+
+Example (WIP): https://github.com/amzn/style-dictionary/pull/487
+
+Thanks [@mfal](https://github.com/mfal)!
+
+### Custom parser support
+
+https://github.com/amzn/style-dictionary/pull/429
+
+We are pretty excited about this. Until now you could only define your design tokens in either JSON, JSON5, or plain Node modules. The addition of custom parser support allows you to define your tokens in any language you like! The most obvious use-case is to use YAML which has a cleaner and less verbose syntax than JSON. But really the sky is the limit. You can now keep your source of truth in any file format you like as long as you can parse it into an object that Style Dictionary can understand. You register a custom parser like you would a custom transform or format. A parser consists of a pattern to match against files, similar to the test attribute in a loader in Webpack, and a parse function which gets the file path and its contents and is expected to return an object.
```
chazzmoney, line 58: JSON. But really the sky is the limit. You can now keep your source of truth in any file format -> JSON. <new paragraph> Now, the sky is the limit. Your source of truth can be in any file format

You register a custom parser like you would a custom transform or format. -> You register a custom parser the same way as you register a custom transform or format.

chazzmoney (changes requested):
This is great! I think it will help people understand what is changing and why.

I'm suggesting a few changes to help with clarity. We may need to do another pass or two - and hopefully we can get some external people folks to review the documentation as soon as we feel comfortable as they will point out more things we likely need to clarify.

docs/version_3.md
```diff
@@ -0,0 +1,136 @@
...
+{
+  "color": {
+    "red": { "value": "#ff0000" }
+  }
+}
+```
+
+Turns into this:
+
+```json
```
alvarocastro, line 82: You can use `json5` as the codeblock syntax to allow the comments to display nicely since `json` doesn't support them.

Using ` ```json `:
```json
{
  "key": "value" // sad comment
}
```
Using ` ```json5 `
```json5
{
  "key": "value" // happy comment
}
```

chazzmoney:
LGTM! :shipit:

chazzmoney (approved):
Looks great!

PR #457, opened 2020-09-02
feat(formats): javascript/module-flat format

noslouch:
*Description of changes:*
Adds a new `javascript/module-flat` format that's just like the `json/flat` format, but exported as a cjs module. We have a use case in a project, and this seemed generic enough to consider for the core.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #493, opened 2020-12-02
feat(formats): add CommonJS format

danlevy1:
*Description of changes:*

Added a CommonJS format (`javascript/cjs`) that outputs a file in a similar manner to the `javascript/es6` format. The only possible concern is that there already exists a similar format called `javascript/module`. I still believe that this new format should be added so that there is parity with the `javascript/es6` format (i.e. flattened names). If needed, I could rename the format to something like `javascript/module-flat`.

Note: This PR is very similar to https://github.com/amzn/style-dictionary/pull/47 that was never finished or merged.

### DIFF
--- a/__tests__/formats/__snapshots__/all.test.js.snap
+++ b/__tests__/formats/__snapshots__/all.test.js.snap
@@ -405,6 +405,15 @@ public enum  {
 "
 `;
 
+exports[`formats all should match javascript/cjs snapshot 1`] = `
+"/**
+ * Do not edit directly
+ * Generated on Sat, 01 Jan 2000 00:00:00 GMT
+ */
+
+module.exports.color_red = \\"#FF0000\\"; // comment"
+`;
+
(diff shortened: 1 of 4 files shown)


dbanksdesign:
Would this be in effect the same as #457? The output code is slightly different:

```javascript
module.exports = {
  "ColorBaseRed": "#ff0000"
}
```

I think consuming this file would be the same as the format you are creating (I might be mistaken though). This feature is in the next version of style dictionary (3.0), which is in pre-release right now. You can get it with `npm install style-dictionary@next` and the format is called "javascript/module-flat".

danlevy1:
Just checked out version 3.0 and it looks good to me! The only reason my formatting was different was because I was trying to stay in-line with the `javascript/es6` format, but there's honestly no good reason to do that. Plus, the format used in #457 is more space efficient because it only uses `module.exports` once. I'm going to close this PR.

PR #496, opened 2020-12-07
fix: add direct injection of props

levi-pires:
To be used in backend projects, to avoid unnecessary file generation.

*Description of changes:*

- Added a shortcut in lib/utils/combineJson ( lines 36 to 46 ) to inject properties directly.
- Changed lib/extend ( lines 114 to 117 ) to avoid throwing errors since `_.isArray({})` returns false


levi-pires:
I didn't realize that there was already a method to inject properties

Issue #449, opened 2020-07-29
Kebab case transform seems inconsistent with other name transforms


juanfabrega:
Hi, hope you're doing well @dbanksdesign! 

I noticed that because the `cti/name/kebab-case` transform uses lodash's `kebabCase` under the hood, it is prone to inserting dashes between numbers and letters as if they were different words. This is inherent in their implementation of kebab and does not appear to be up for changing any time soon.

Issues raised in lodash's repo here asking similar questions:
https://github.com/lodash/lodash/issues/2425
https://github.com/lodash/lodash/issues/4462
https://github.com/lodash/lodash/issues/4205
https://github.com/lodash/lodash/issues/4141

Here's an example of what is happening in our dictionary:

Where the json is:
```
"size": {
  "spacing": {
    "2xl": { "value": "2.5" }
  }
}
```

Our expectation for a css variable is `--size-spacing-2xl`
The actual output is `--size-spacing-2-xl`

I understand this is probably by design, but unfortunately it creates inconsistencies between css/scss variable outputs and js/json outputs, where the resulting objects still have the original non-kebab attribute.

We have already solved this with a custom transform on our dictionary but I thought it was worth bringing up to see if you'd be open to a PR to change the default kebab transforms. Let me know if there's a reason that this should **not** be changed as we are ok with the custom transform, but perhaps it's worth bringing up in the docs somewhere. (Perhaps it's already in the docs and I missed it)

All the best!

dbanksdesign:
Good catch! This is brought up in another issue as well: #300. I think it would be a good idea to migrate off lodash's string methods to get the behavior we want. I think this would be a good addition. We have a 3.0 branch for the next major release of style dictionary so this would make sense to go in there as it is a breaking change. If you'd like, please open a PR on the 3.0 branch and we can get this in!

juanfabrega:
> Good catch! This is brought up in another issue as well: #300. I think it would be a good idea to migrate off lodash's string methods to get the behavior we want. I think this would be a good addition. We have a 3.0 branch for the next major release of style dictionary so this would make sense to go in there as it is a breaking change. If you'd like, please open a PR on the 3.0 branch and we can get this in!

Will do, thanks for the quick response!

chazzmoney:
Hi @juanfabrega !

I was getting around to checking on this as part of our 3.0 work. If you have a copy of your custom transform, I'd love to borrow it to see if it might be useable to solve this as part of 3.0

Thanks!

[...]

Issue #300, opened 2019-06-27
Flat output turns "2x" into "2-x"


nategreen:
I have tokens like:
```
{
  space: {
    "2x-small": { value: "2px" },
    "x-small": { value: "4px" },
    "small": { value: "8px" },
    "medium": { value: "16px" },
    "large": { value: "32px" },
    "x-large": { value: "64px" }
  }
}
```
And when they get turned into Sass variables, the ones with "2x" come out as e.g.:
```
$space-2-x-small
```
Likewise in the flat Sass map format, but in the deep Sass map format, the map keys come out as `2x` even though the variable referenced is `2-x`:
```
'padding': (
      '2x-small': (
        'base': $mri-space-padding-2-x-small-base
      ),
```
🤔 Very strange, but is this intended? (Perhaps to ensure that the keys don't break some other output?) I'm using JSON5 format, by the way, if that's important.

```
style-dictionary: 2.8.0
```

dbanksdesign:
I think it isn't the output, but the 'name/cti/kebab' transform that is the culprit. We are using lodash's string method, [kebabCase method](https://lodash.com/docs/4.17.11#kebabCase) (and similar string functions for other name transforms). I'm guessing lodash see's "2x" and thinks those are 2 separate words... I just tested it out on the lodash documentation site's REPL

littlefengers:
I had to make a custom transform with this package: https://github.com/blakeembrey/change-case/tree/master/packages/param-case

`StyleDictionary.registerTransform({
  name: 'name/cti/kebabCustom',
  type: 'name',
  transformer: function(prop,options){
    return paramCase([options.prefix].concat(prop.path).join(' '));
  }

})`

dbanksdesign:
It might be a good idea to take a look at migrating off lodash's string methods to change-case if it fixes this issue (and does not introduce any new issues).

chazzmoney:
@nategreen @wiese @littlefengers 

We are changing our case-changing support in 3.0 to use [change-case](https://github.com/blakeembrey/change-case) and want to create some unit tests around it. If you have any challenging ones, please submit them on #500

chazzmoney:
This is now solved in 3.0

PR #500, opened 2020-12-08
Lodash removal

chazzmoney:
Initial efforts are replacing lodash. This is amateur hour at the moment; all suggestions welcome.


lib/utils/es6_.js
```diff
@@ -0,0 +1,57 @@
...
+
+let isArray = function(obj) {
+  return Array.isArray(obj)
+}
+
+module.exports = {
+  each: forEach,
+  forEach: forEach,
+  forIn: forEach,
+  keys: Object.keys,
```
dbanksdesign, line 53: For something like this, could we instead replace instances of `_.keys()` with `Object.keys()`? What do you think?

lib/utils/es6_.js
```diff
@@ -0,0 +1,57 @@
...
+      return cpObj;
+    }, {})
+  }
+}
+
+let isString = function(obj) {
+  return typeof obj === 'string' || obj instanceof String
+}
+
+let isArray = function(obj) {
```
dbanksdesign, line 45: Same question as `_.keys()`. What do you think?

lib/utils/es6_.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+let forEach = function(obj, f) {
+  Object.keys(obj).forEach((key) => {
+    let value = obj[key]
+    f(value, key)
+  });
+}
+
+let cloneDeep = function(obj) {
```
dbanksdesign, line 21: This method is very clean. ✨

dbanksdesign (commented):
❤️ this PR! 2 questions about `_.keys` and `_.isArray`, curious your thoughts

dbanksdesign:
It is fitting that this is PR # 500 😁

lib/utils/es6_.js
didoo: +1

chazzmoney: Yes, we should do this. My suggestion is to remove _. first, and then switch to native code everywhere else after.

chazzmoney:
@dbanksdesign 

So it is getting close... the few issues left:
1. My replacement for _.pull does not play well with transform/object.js line 86. It causes the tests to hang. Replace '_.pull' with '_es6.pull' to see it in action... Likely I just need to dive deeper on what is being passed in / out
2. Some very hefty work to be done to replace _.template
3. The case-changing stuff is basically copied from lodash and includes its quirks. So.... maybe we want to redo it another way (issues #449 #300 #257)

chazzmoney:
What if we switched from lodash to https://github.com/blakeembrey/change-case ?

dbanksdesign:
I am good swapping one dependency for another is ok, especially if this dependency is more configurable and generally a smaller and more narrow dependency, which this is.

chazzmoney:
moved to lodash-removal-3 #570

PR #499, opened 2020-12-08
fix(extend): use given file path for token data

dbanksdesign:
*Issue #, if available:* #498

*Description of changes:* Changed the `filePath` attribute Style Dictionary adds to all tokens to be the filePath provided by 'source' or 'includes'. If you provide a relative file path like `["tokens/**/*.json"]` then filePath will now be "tokens/color/core.json" rather than "/Users/djb/dev/my-style-dictionary/tokens/color/core.json" for example.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #501, opened 2020-12-08
chore(types): adds filePath and isSource to Prop and descriptions to Props keys

luke-john:
*Description of changes:*

Adds the new isSource and filePath to the type of Prop.

Also added descriptions to the keys of Prop interface based off the docs added in https://github.com/amzn/style-dictionary/pull/499.

(note: these were still usable pre this change due to the "catch all" index signature `[key: string]: any;` on `Prop`).

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -71,6 +71,10 @@ declare namespace StyleDictionary {
     attributes: Attributes;
     path: string[];
     value: string;
+    /** A string representing the absolute path of the file that defines the token. */
+    filePath: string;
+    /** Represents if this file was defined as ‘source’ in the configuration as opposed to ‘include’ (or directly setting the ‘properties’ object). */
+    isSource: boolean;
     [key: string]: any;
   }
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -66,15 +66,31 @@ declare namespace StyleDictionary {
   }
 
   interface Prop {
-    original: Property;
+    /** A default name of the property that is set to the key of the property. */
     name: string;
-    attributes: Attributes;
-    path: string[];
     value: string;
-    /** A string representing the absolute path of the file that defines the token. */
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -71,6 +71,10 @@ declare namespace StyleDictionary {
     attributes: Attributes;
     path: string[];
     value: string;
+    /** A string representing the absolute path of the file that defines the token. */
```
luke-john, line 74: note: this should change based on https://github.com/amzn/style-dictionary/pull/499

chazzmoney: agreed- update as you see fit (you can look at what we did in #499 and update as you see fit)

types/index.d.ts
```diff
@@ -66,11 +66,31 @@ declare namespace StyleDictionary {
...
+     */
+    original: Property;
+    /**
+     * The file path of the file the token is defined in.
+     *
+     * This file path is derived from the source or include file path arrays defined in the configuration.
+     */
+    filePath: string;
+    /**
+     * If the token is from a file defined in the source array as opposed to include in the [configuration](https://amzn.github.io/style-dictionary/#/config).
```
luke-john, line 90: In case the reviewer is unfamiliar with how these comments are presented, markdown links show as a link (which is clickable).

image

types/index.d.ts
luke-john: Updated thanks (and added descriptions for the additional keys that had docs added in #499).  Also fixed a typo in that file.

dbanksdesign (approved):
LGTM! :shipit:

chazzmoney:
Thank you for this. It is a very useful contribution and one we would not have gotten to ourselves. Your work here will make SD more usable for everyone. Big +1 and Heart from me.

❤️ 👍 

:shipit:

PR #502, opened 2020-12-09
chore(types): fix transform options type

amalik2:
*Issue #, if available:*

*Description of changes:*
Based on the code [here](https://github.com/amzn/style-dictionary/blob/3.0/lib/exportPlatform.js#L96) and in the tests for transforms, it looks like the correct type for transformer functions should be `Platform` instead of `Options`

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -77,19 +77,19 @@ declare namespace StyleDictionary {
   interface NameTransform {
     type: "name";
     matcher?: (prop: Prop) => boolean;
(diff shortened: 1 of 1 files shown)


chazzmoney:
Thank you for the fix - appreciate it! :shipit:

chazzmoney:
@dbanksdesign We should rename the variable to "platform" from "options" too...

chazzmoney (approved):
👍

Issue #396, opened 2020-03-06
Question: Is there a way to transform alias into css variables without resolving the actual value?


davidx1:
Is there a way to NOT resolve alias to values, but rather transform alias into css variable names?

For example, if I have defined in my properties like so: 
```
//base.json
{
   "color": {
           "g500": { "value": "green" },
       }
    }
}

//other.json
{
   "color": {
           "success": { "value": {color.g500} },
       }
    }
}
```

Is there a way to generate:
```
:root {
  --g500: green;  
  --success: var(--g500);
}
```

EDIT: 
The reason I ask is that we want to use StoryBooks' Design Token addon which uses css variable aliases to identify when two tokens are actually the same color and display them as such:

eg: here, both `success` and `published` are `g500`

![image](https://user-images.githubusercontent.com/12288086/76040155-8007c900-5fb3-11ea-8339-223c4d918221.png)

dbanksdesign:
There is no way in the built-in transforms and formats, but you could write a custom one to do that (in kind of a hacky way). You could write a custom format that iterates over each token and does a check to see if `token.original` contains `{}` meaning it references another token. If it does, get the name of that token and put it in a `var()` instead of outputting the token's resolved value.

chazzmoney:
If you build one, we would love a PR for a format that includes this!

avkvirtru:
FYI, I've been tinkering with the solution to this started in #428

GarthDB:
This is how Adobe handles aliases (keeping them intact) in Spectrum CSS. I am building a new design system for a personal project and was looking for the same functionality in Style Dictionary. If there's anything I can do to help, let me know.

avkvirtru:
@GarthDB I have a solution for this locally. Haven't had a chance to open a PR and contribute back here. Would be happy to coordinate on that. Can we start by seeing if my solution helps you: aklepchukov@virtru.com ?

tonyjwalt:
I have this in my PR - https://github.com/amzn/style-dictionary/pull/428
-Tony

chazzmoney:
This is now possible via the merge of #504 

Thanks @dbanksdesign

Issue #17, opened 2017-03-03
Add support for variable references in output


dbanksdesign:
This would be useful to see reference chain in the generated files. Although would need to only be used for files that have all properties in them or there might be errors.

chazzmoney:
The question here is whether this is a use case that we see as being valuable.  It isn't quite on the path of "enable end-to-end consistency of design atoms across platforms / deliverables.

sarahbethfederman:
I would definitely find this useful for use in supporting tooling. I would love to have a vscode plugin that shows the reference chain on hover when using a token

didoo:
Let's keep in mind also that a token "A" can reference as "alias" another token, let's say "B", who can reference another token, "C" , who can reference another token, "D", who finally references a key/value pair, let's say "XYZ / 12px". In that case, which one is the reference to show for "A"? "B" or "XYZ" ?

chazzmoney:
Or do we want to show the entire chain?

XYZ, D, C, B, A

dbanksdesign:
The issue in showing the entire chain in the output like this:
```scss
$A: $B;
$B: $C;
$C: $D;
$D: $XYZ;
$XYZ: 12px;
```

Is that file would need to include all tokens to ensure the file is valid. Or if we don't we at least need to make sure the full variable chain is there, but to also not duplicate variable definitions or that might cause issues. 

Now a VSCode plugin that shows the reference chain in a style dictionary...

lukasoppermann:
If I understand this correctly, it would mean in css I would have:

```css
--core-primary-color: #0022AA;
/*...*/
--interactive-color: var(--core-primary-color);
/*...*/
--button-color-primary: var(--interactive-color);
```

This would be very helpful for the goal of building a somewhat "self-documenting" and human readable system. Since the idea of design tokens is to provide `named entities` and bring meaning into the whole mess, I'd argue a referenced variable is very valuable. It helps me to understand why this color is used for the button color, vs. just `--button-color-primary: #0022AA`

dbanksdesign:
This is in 3.0 and you can get it today with `npm install style-dictionary@next`. For users of the future, this feature is available in Style Dictionary starting with 3.0.

Issue #268, opened 2019-04-23
Should we consider token organization a primary use case?


chazzmoney:
Especially per #246 , people are using style dictionary both to [create original design tokens AND to reference them into specific components](https://medium.com/@didoo/how-to-manage-your-design-tokens-with-style-dictionary-98c795b938aa#5faf). In this case it is clear that the file organization is representative of that "both" structure. 

This is all fine until you need to output those tokens. From that same article:

> Now, because these functions operate directly on the name, value and attributes of a token, it means that they are somehow coupled with the way you want to declare your properties and values...
> 
> Now, as explained above, I wanted to have a very specific organization of the generated files for the different platforms: not only have specific formats for each one of them (Web, iOS and Android) saved in their corresponding folder, but I also needed all the possible combinations/permutations of platform and brand made available for the style guide.
> 
> I also wanted to have complete freedom in how I organized and declared the key/values properties for the tokens, to be able to manipulate them at will.
> 
> So I decided to venture in creating a custom build script leveraging the APIs made available by the Style Dictionary library.

While #267 is an example of how to solve this today that we can provide our users, it is a very technically demanding method. So while we support the use case, we do so only in a round-about way and require technical knowledge.

It is my belief that if we want to enable simple design token control for people, we should probably make this a direct goal and engineer a simple solution.

I wanted to open the discussion. Let me know your thoughts.

chazzmoney:
@didoo I used your article as an example of what people want to be able to do with Style Dictionary. I know that you love us and are very happy, but I wanted to try to capture the work you did from the point of view of a new user and the logic they apply to how to do what they want using Style Dictionary.

I definitely do not want to misrepresent you, so you can clarify if I am. Just want to be as user centric as possible with the solution we provide.

didoo:
@chazzmoney no problem in using the article, is for public consumption, knowledge, and discussion :) 

Re. this issue, I am not sure what you're proposing. Clearly the tokens organisation (the file structure of the source files + the organisation of the JSON trees of keys/children and keys/values in these file) is fundamental when you use Style Dictionary. Is also true, that each one of us probably will organise them in different way, depending by their specific requirements, and personal preferences.

So, what can we do here to help people _do what they want using Style Dictionary_?

dbanksdesign:
What I want is to strike a balance of: flexibility, to @didoo's point everyone will organize differently, while providing a simple baseline that can handle most basic cases. (Granted "basic" and "simple" are hard to quantify)

[...]

PR #504, opened 2020-12-11
feat(format): output references

dbanksdesign:
*Issue #, if available:* #17 #396

*Description of changes:* Allow for variable references in output files. These changes will make it easier to use the *name* of a referenced token rather than the resolved value, so the output will keep the aliases/references intact.

* Adding `resolveReference()` and `usesReference()` methods on the dictionary object that is passed to formats and actions. This will allow custom formats and actions to test and resolve references the same way style dictionary does internally, allowing you to get the transformed name of a token for instance.
* Adding a `outputReferences` configuration on SCSS, CSS, and Less variables, as well as `ios/class.swift`, `flutter/class.dart`, and `android/resources` formats to make use of this
* Adding a 'output-references' example to show how to use these new methods
* Added documentation about this feature
* Added integration testing to properly test this


chazzmoney:
WOOHOO! #17

chazzmoney:
I really like that this empowers people on the route of #268, which is something that I have wanted for SD for a LONG time (even well before I created that issue).

I think it is super valuable that people can customize SD for how **they** think about tokens. Even so, it encourages base definitions in one place and output definitions in another.

It also enables client side debugging/tooling to have strong access to the underlying reference structure that powers SD.

Really really excited about this one.

chazzmoney:
Two improvements to consider:

1. I think this deserves documentation catering to the absolute newest / least experienced SD user. This is a very important / useful thing and making sure people know it is available and can understand how to use it easily will help them use SD to its fullest.
2. Based on the usefulness to downstream tooling, we may want to make this on by default... this will be included in a major version/breaking change so it might be ok... we should probably @ some of our most active folks here to get their opinion.
3. The configuration options here should likely be able to be set at three levels: global, included file, and property (maybe even inheritably so)

chazzmoney:
Also... we should probably update all default formats to use variables where appropriate.

It will be a lot of work, but it is weird to include this without updating all of them.

lib/common/formats.js
```diff
@@ -55,9 +55,20 @@ function formattedVariables(format, properties) {
...
-  return properties.map(function(prop) {
+  return dictionary.allProperties.map(function(prop) {
       var to_ret_prop = prefix + prop.name + ': ';
-      to_ret_prop += (prop.attributes.category==='asset' ? '"'+prop.value+'"' : prop.value);
+      let value = prop.value;
+
+      if (useVariables && dictionary.usesReference(prop.original.value)) {
+        var ref = dictionary.getReference(prop.original.value).name;
+        if (format === 'css') {
+          value = `var(${prefix}${ref})`
```
didoo, line 65: nitpicking here, but in the output it generates something like:
```
:root {
  --token-color-error: var(  --token-color-generic-red);
}
```
and that extra space inside the var() is... annoying :)

I think more correctly the `prefix` value at line 46 should be `--` and not `  --` (formally, more correct) and then we should add two extra spaces in output when we generate the line (`var to_ret_prop = prefix + prop.name + ': ';`)

lib/common/formats.js
```diff
@@ -55,9 +55,20 @@ function formattedVariables(format, properties) {
...
+        } else {
+          value = prefix + ref;
```
didoo, line 67: for consistency with two lines above I would do:
```suggestion
          value = `${prefix}${ref}`;
```

examples/advanced/variables-in-outputs/README.md
```diff
@@ -0,0 +1,48 @@
...
+#### How does it work
+
+The "build" command uses the `sd.config.js` file as the Style Dictionary configuration. It is configured to use JSON files in the `tokens/` directory as the source files. It adds a custom format directly in the configuration (as opposed to using the `.registerFormat()` method) that uses 2 new methods added onto the internal dictionary object that is passed to formats and actions: `.usesReference()` and `.resolveReference()`. Also, it uses a new configuration on some formats: `useVariables: true` to include variable references in the output.
+
+
+#### What to look at
+
+The `sd.config.js` file has everything you need to see. The tokens included in this example are just enough to show these new techniques.
+
+```javascript
```
didoo, line 24:
```suggestion
Here is an example that shows how to get an alias's name within a custom format:
```javascript
```

lib/utils/references/resolveReference.js
```diff
@@ -0,0 +1,36 @@
...
+
+function resolveReference(path, obj) {
+  let i,
+      ref = obj;
+
+  if (!Array.isArray(path)) {
+    return;
+  }
+
+  for (i = 0; i < path.length; i++) {
```
didoo, line 22: now that it's been decided to move away from lodash, and the "utils" folder has started to contain lodash-like functions, do you think we should isolate this as `deepFind` function there?

didoo:
First of all, amazing PR @dbanksdesign, really great idea 🙌.

Now, I did a quick code review and added a few comments there. I have some other thoughts here:
- I am not entirely sure about the name `useVariables`. maybe there is something that could be more explicit on what it does? I was thinking something like `preserve[References/Variables]` or `use[Variables|References]InOutput`
- agree with @chazzmoney that we should consider also other less-obvious default formats that may support this kind of referencing in output (which one? JS clearly yes, in some form; but what about others like formats for iOS or Android?)
- regarding setting it on by default for version 3.0, agree as well, makes sense (unless there are downsides we're overlooking)
- regarding the three levels configuration, not sure about it. why do you think we need this extra layer of complexity in the codebase? if you can achieve the result adding a set of `useVariables: true` wherever you need it for the output, I don't think is much of a burden for the final user, and in any case, you probably do it once at the initial setup and then forget.
- regarding the documentation, there are other parts that we may want to highlight for the new user (eg. my colleague didn't realize the input format for the tokens could be directly JavaScript code, not only JSON) so maybe this could be done in a separate task?

Let me know if I can help somehow on this.

dbanksdesign:
@didoo I would love some help with this! I think we need some more tests, more documentation, and updating applicable formats to use this new code. Whichever you'd like to help out on would be greatly appreciated. I also have some time tomorrow I can spend on this and you can take a look.

dbanksdesign:
Big update:
* Changed the option name from 'useVariables' to 'keepReferences' so it better matches the methods
* Moved the option to the options object in a file, to be consistent with other file-level configuration like `showFileHeader`
* Updated `android/resources` format to handle `keepReferences`
* Updated `scss/variables` format to handle `keepReferences`. In order to do this, because Sass is imperative, we are sorting the `allProperties` array so all references come last.
* Started to add integration testing to better test the output of formats. 
* Cleaned up some things in our testing

dbanksdesign:
Another update:
* Changed option name to `outputReferences` and moved it to the `options` object on a file
* Added Swift and Flutter support and integration tests
* Started some documentation

docs/formats.md
```diff
@@ -73,6 +73,54 @@ A special file configuration is `filter`, which will filter the tokens before th
...
+
+This would be the output:
+
+```css
+:root {
+  --color-red: #ff0000;
+  --color-danger: var(--color-red);
+  --color-error: var(--color-danger);
+}
+```
```
chazzmoney, line 120: I love this as an example. You might contrast it with what happens if outputReferences is false.

chazzmoney:
Nice work. Seriously. LGTM! :shipit:

dbanksdesign:
Update:
* Added documentation in the config section about this
* Platform level configuration can now define an `options` object with the same options as a file object, and the options cascade
* Renamed the example

uptonking:
@dbanksdesign Is there any way to support generate css variables with fallback values?
tokens source: 
```json
{ 
  "color": {
    "background": {
      "primary": { 
		"value": "{color.palette.neutral.0.value}" ,
		"fallback":"red"
		}
    }
  }
}
```

tokens output: 
```css
/** simple form */
--color-background-primary: var(--my-var, red);

/* complex var may not be necessary */
--color-background-primary: var(--my-var, var(--my-background, pink));
```

https://github.com/amzn/style-dictionary/issues/544 this issue seems similar.

Issue #453, opened 2020-08-19
Android empty exports -- am I missing something?


carbonology:
Hi everyone! Thank you for clicking into this issue, and thanks to the team for their support. 🍻 

**Problem:** Android transforms will export an empty-ish file if the categories aren't named correctly. For example, a category named `color` (and its types + tokens) will export using the 'android/colors' format correctly. However, a category named 'color**s**', with the same configuration, will not be recognized or exported at all. Any 'custom' category will be ignored.

**What I expect**: Disclaimer: I'm not an Android developer. But what I'd expect is for StyleDictionary to output each token into a file, regardless of its category name. No token should be ignored and perhaps exported to a more general xml file. Below you will see how the android formats will completely ignore unrecognized categories. Is there another format or documentation I'm missing that will enable this?

**Extra details**:
Here is an example StyleDictionary object that I'm using:
```
{
   "fun-colors":{
      "button":{
         "active":{
            "value":"red"
         },
         "hover":{
            "value":"blue"
         }
      }
   },
   "color":{
      "button":{
         "active":{
            "value":"red"
         },
         "hover":{
            "value":"blue"
         }
      }
   }
}

```

With the Android part of my config as:
```
android: {
            transformGroup: 'android',
            buildPath: `${buildPath}/android/`,
            files: [{
              destination: 'token_colors.xml',
              format: 'android/colors'
            },{
              destination: 'token_font_dimens.xml',
              format: 'android/fontDimens'
            },{
              destination: 'token_dimens.xml',
              format: 'android/dimens'
            },{
              destination: 'token_integers.xml',
              format: 'android/integers'
           },{
              destination: 'token_strings.xml',
              format: 'android/strings'
            }]
        },
```

Will export:
**token_colors.xml** (missing the "fun-colors" category)
```
<?xml version="1.0" encoding="UTF-8"?>

<!--
  Do not edit directly
  Generated on Wed, 19 Aug 2020 15:31:22 GMT
-->
<resources>
  <color name="color_button_active">#ffff0000</color>
  <color name="color_button_hover">#ff0000ff</color>
  
</resources>

```

**token_font_dimens.xml** (Empty)
```
<?xml version="1.0" encoding="UTF-8"?>

<!--
  Do not edit directly
  Generated on Wed, 19 Aug 2020 15:31:22 GMT
-->
<resources>
  
</resources>

```

**token_dimens.xml** (Empty)
```
<?xml version="1.0" encoding="UTF-8"?>

<!--
[...]

PR #509, opened 2020-12-15
feat(format): adding android/resources format

dbanksdesign:
*Issue #, if available:* #453

*Description of changes:* Adding a new format for android: 'android/resources'. This new format removes the filtering inside the format that the previous android formats had so you can filter your tokens however you like. This format allows for mixing multiple token types (string, dimen, color, etc.) in a single file. It is also customizable with these settings:
* resourceType: will force all tokens to be a specific resource type like color or dimen
* resourceMap: allows you to define a custom map of which token categories should map to which resource types. 


chazzmoney:
LGTM! :shipit:

Issue #510, opened 2020-12-17
yaml-tokens example doesn't produce expected results


jbarreiros:
Pulled down `3.0.0-rc.1` to test it out and it has some great stuff. The transitive feature is :hot_pepper: 

Testing out the "yaml-tokens" example, and when I run `npm run build`, it completes, but the final result (`build/variables.css`) looks like:
```css
/**
 * Do not edit directly
 * Generated on Thu, 17 Dec 2020 17:52:16 GMT
 */

:root {

}
```

Interestingly, if I put the following at the top of `sd.config.js`, it shows that the `yaml` package is interpreting the yaml file correctly.
```js
const yaml = require('yaml');
const fs = require('fs');

const yamlraw = fs.readFileSync('./tokens/color/core.yaml', {encoding: 'UTF-8'});
const result = yaml.parse(yamlraw);
console.log(result);
```
```js
{ color:
   { base:
      { white: [Object],
        red: [Object],
        green: [Object],
        success: [Object] } } }
```

dbanksdesign:
There was an issue that no parsers were being run because we changed the method signature of combineJSON.. but I posted a PR for that fix. Thanks for bringing this up!

chazzmoney:
This is fixed now in the 3.0 branch and will be part of 3.0.0-rc.2

PR #511, opened 2020-12-17
fix(parsers): fixed an error where parsers weren't running

dbanksdesign:
*Issue #, if available:* #510

*Description of changes:* The extend method was sending the wrong parameters to combineJSON so parsers never ran. This came up because of 2 PRs changing the combineJSON method signature. For future sanity we might want to consider changing methods to take a single object.

### DIFF
--- a/examples/advanced/yaml-tokens/package-lock.json
+++ b/examples/advanced/yaml-tokens/package-lock.json
@@ -5,12 +5,12 @@
   "requires": true,
   "dependencies": {
     "ansi-styles": {
-      "version": "3.2.1",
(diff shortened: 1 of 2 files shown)


chazzmoney:
LGTM :shipit:

SalimBensiali:
Hi @dbanksdesign, any idea on when this fix will make it to the next rc?

Thanks

SalimBensiali:
I just saw https://github.com/amzn/style-dictionary/issues/514#issuecomment-758318159, thanks for releasing rc.2 with the fix!

PR #521, opened 2021-01-22
chore(deps): fix security dependency in cra example

dbanksdesign:
*Issue #, if available:*

*Description of changes:* The immer dependency has a security vulnerability. However, it is a nested dependency of `react-scripts` which has not yet updated its dependency. To get around this we are going to use `npm-force-resolutions` to fix the immer version at 8.0.1


chazzmoney:
Silly to have to do this, but safer for our users. Thanks @dbanksdesign 

LGTM :shipit:

Issue #526, opened 2021-02-01
Support for Stylus format


klausbayrhammer:
Some of our internal projects use [Stylus](https://www.npmjs.com/package/stylus) as a preprocessor, and we've written a custom formatter to support this (which works great).

Since stylus is relatively popular we would happily contribute it. However we were not sure if based on the ["what should be included" section in the contributing guidelines](https://github.com/amzn/style-dictionary/blob/main/CONTRIBUTING.md#what-should-be-included) stylus support is something you would like to see within the style-dictionary.

dbanksdesign:
Stylus would definitely be something we would see within Style Dictionary. That section in the contribution doc is mainly so we don't get non-generic formats, but we always want to add support for languages. If you would like to contribute that, take a look at the Less, SCSS, and CSS formats. I'm not as familiar with Stylus, I assume we don't need to add any new transforms to support Stylus.

chazzmoney:
This will be included in the next release candidate for 3.0. Once that is done (probably next Monday, Feb 15th, 2021), you can try it out using:

`npm install --save-dev style-dictionary@next`

dbanksdesign:
Just released the next release candidate, it should now be available with the `next` tag like @chazzmoney said above

PR #527, opened 2021-02-02
feat(format): stylus support

klausbayrhammer:
*Issue #, if available:*

See #526 

*Description of changes:*
This PR adds support for `stylus/variables`.

Although leading dollar signs are not mandatory for stylus variables there is a trend to having this, so I also included it in this PR. See https://gist.github.com/zspecza/7220997#why

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -37,22 +37,31 @@ function fileHeader(showFileHeader = true, commentStyle) {
 }
 
 function formattedVariables(format, dictionary, outputReferences = false) {
-  var prefix, commentStyle, indentation;
+  var prefix, commentStyle, indentation, separator;
 
   switch(format) {
     case 'css':
       prefix = '--';
       indentation = '  ';
+      separator = ':';
       break;
     case 'sass':
       prefix = '$';
       commentStyle = 'short';
       indentation = '';
+      separator = ':';
       break;
     case 'less':
       prefix = '@';
       commentStyle = 'short';
       indentation = '';
+      separator = ':';
+      break;
+    case 'stylus':
+      prefix = '$';
+      commentStyle = 'short';
+      indentation = '';
+      separator = '=';
       break;
   }
 
@@ -72,7 +81,7 @@ function formattedVariables(format, dictionary, outputReferences = false) {
       }
     })
     .map(function(prop) {
-      var to_ret_prop = `${indentation}${prefix}${prop.name}: `;
+      var to_ret_prop = `${indentation}${prefix}${prop.name}${separator} `;
       let value = prop.value;
 
       if (outputReferences && dictionary.usesReference(prop.original.value)) {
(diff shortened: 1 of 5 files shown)


__tests__/formats/stylusVariables.test.js
```diff
@@ -0,0 +1,65 @@
...
+          if (err) {
+            return done(new Error(err));
+          }
+          expect(css).toBeDefined();
+          return done();
+        });
+    });
+
+  });
+});
```
klausbayrhammer, line 65: @dbanksdesign  this tests is pretty similar to the [lessVariables.test.js](https://github.com/amzn/style-dictionary/blob/main/__tests__/formats/lessVariables.test.js) or the [scssVariables.test.js](https://github.com/amzn/style-dictionary/blob/main/__tests__/formats/scssVariables.test.js) - would it make sense to combine the test setup?

lib/common/formats.js
```diff
@@ -38,20 +38,20 @@ function fileHeader(options, commentStyle) {
...
-        if (commentStyle === 'short') {
-          to_ret_prop = to_ret_prop.concat(' // ' + prop.comment);
-        } else {
-          to_ret_prop = to_ret_prop.concat(' /* ' + prop.comment + ' */');
-        }
+    if (prop.comment) {
+      if (commentStyle === 'short') {
+        to_ret_prop = to_ret_prop.concat(' // ' + prop.comment);
+      } else {
+        to_ret_prop = to_ret_prop.concat(' /* ' + prop.comment + ' */');
```
klausbayrhammer, line 49: I've moved the indentation alignment in a separate commit - if you feel like we should move this out of this PR I'll happily do this.

lib/common/formats.js
```diff
@@ -239,6 +240,21 @@ module.exports = {
...
+  /**
+   * Creates a Stylus file with variable definitions based on the style dictionary
+   *
+   * @memberof Formats
+   * @kind member
+   * @example
+   * ```stylus
+   * $color-background-base= #f0f0f0;
+   * $color-background-alt= #eeeeee;
+   * ```
```
klausbayrhammer, line 252: Although stylus does not require trailing semicolons, it does no harm. If we want to be more concise we could also remove them, but for the cost of adding another parameter to the `variablesWithPrefix` function.

__tests__/formats/stylusVariables.test.js
dbanksdesign: Potentially... I think there is definitely some cleanup and better testing we could be doing for formats for sure. I think we can tackle that in a separate PR though.

lib/common/formats.js
dbanksdesign: Looks good to me! Cleaning up indentation and alignment is always welcome. We should probably use prettier or something in the future though... That'll be a different PR

chazzmoney (approved):

dbanksdesign:
Thank you for this addition! We are working off the 3.0 branch to target the next major release. We switched the branch for this PR and there are some conflicts. If you could fix these we can merge this into the 3.0 branch. If you need any help with any of the conflicts let me know and I can help out.

lib/common/formats.js
klausbayrhammer: Sounds like a good idea! I've removed the indentation cleanup for now.

dbanksdesign (approved):
LGTM! :shipit: Thank you again for this addition!

klausbayrhammer:
Thanks for merging it! Really appreciate all the work you put into style-dictionary!

Issue #528, opened 2021-02-02
Allow testing with a mocked filesystem


tlouisse:
Currently I'm writing tests for my custom transforms, formatters and filters.

For simplicity and readability I use [mock-fs](https://www.npmjs.com/package/mock-fs) to mock the filesystem.
In a test, this looks something like:

```js
     mockFsAndRequire({
        '/my/dictionary/mixin.json': JSON.stringify({
          elevation: {
            l0: {
              shadow: { value: 'none', propRole: 'shadow' },
              bgcolor: { value: 'currentcolor', propRole: 'bgColor' },
            },
            l1: {
           ...etc..
```

However, it is currently not compatible. I get this error:
```js
Error: Cannot find module '/my/dictionary/mixin.json'
Require stack:
- /path/to/current/wd/noop.js
      at Function.Module._resolveFilename (node:internal/modules/cjs/loader:925:15)
      at resolveFileName (node_modules/resolve-cwd/node_modules/resolve-from/index.js:29:39)
      at resolveFrom (node_modules/resolve-cwd/node_modules/resolve-from/index.js:43:9)
      at module.exports (node_modules/resolve-cwd/node_modules/resolve-from/index.js:46:47)
      at module.exports (node_modules/resolve-cwd/index.js:4:30)
      at combineJSON (node_modules/style-dictionary/lib/utils/combineJSON.js:42:24)
      at Object.extend (node_modules/style-dictionary/lib/extend.js:117:17)
      at runDictionary (dictionary/runDictionary.js:119:41)
      at Context.<anonymous> (dictionary/test-node/runDictionary.test.js:110:7)
      at processImmediate (node:internal/timers:463:21)
```

When I dive into the code, I see that `resolve-cwd` creates this problem: https://github.com/amzn/style-dictionary/blob/main/lib/utils/combineJSON.js#L41
When I use `path.resolve(process.cwd(), myPath)` instead, everything just works as expected.
Also, all tests keep working as expected.

Would it be possible to do a small code change here?

tlouisse:
Closing the issue, since the pull request has been merged: https://github.com/amzn/style-dictionary/pull/529
Thanks for helping out! 👍

PR #529, opened 2021-02-02
chore: replace resolve-cwd with platform capabilities

tlouisse:
*Issue #, if available:*
In order to make it possible to mock the filesystem in my tests, I would need a small change. For an explanation, see: https://github.com/amzn/style-dictionary/issues/528

*Description of changes:*
Replaced all occurences of `resolveCwd(myPath)` with `path.resolve(process.cwd(), myPath)`

### DIFF
--- a/lib/extend.js
+++ b/lib/extend.js
@@ -15,7 +15,7 @@ require('json5/lib/register');
 
 var combineJSON = require('./utils/combineJSON'),
     deepExtend = require('./utils/deepExtend'),
-    resolveCwd = require('resolve-cwd'),
+    path = require('path'),
     _ = require('lodash'),
     GroupMessages = require('./utils/groupMessages');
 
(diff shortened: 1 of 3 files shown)


lib/extend.js
```diff
@@ -11,14 +11,14 @@
  * and limitations under the License.
  */
 
-require('json5/lib/register');
+require("json5/lib/register");
```
tlouisse, line 14: Autoformatting (Prettier) was a bit too aggressive. I will revert those 'quote changes' :)

tlouisse:
Hi @chazzmoney,

I see that you changed the target branch to 3.0. I updated my merge request and resolved all merge conflicts.
Wrt this line, I could have kept using resolveCwd if preferred:

```js
var resolvedPath = path.isAbsolute(files[i]) ? files[i] : path.resolve(process.cwd(), files[i]);
```

It's actually not the fact that `resolveCwd(...)` was replaced with `path.resolve(process.cwd, ...)` that makes it work for me(#528),but rather the fact that absolute paths are not resolved.

In other words, if you want to keep using `resolveCwd` package, this would also have worked:
```js
var resolvedPath = path.isAbsolute(files[i]) ? files[i] : resolvedCwd(files[i]);
```
However, I think it's still nice to get rid of an external package (resolve-cwd) if possible.


Let me know what your thoughts are. If needed, I could keep the resolveCwd package as well.

chazzmoney:
Thanks for resolving to the 3.0 branch!

chazzmoney (approved):

PR #530, opened 2021-02-02
fix(types): Change transforms to transform in Core

Tenga:
*Description of changes:*

By looking [at the source](https://github.com/amzn/style-dictionary/blob/3.0/index.js#L37), and by accessing the default import at runtime, it seems like the `transforms` property of `Core` should be `transform`.

This PR makes that tiny fix so there are no issues when trying to access the default transforms on the default import.

---


chazzmoney:
Thanks for catching this!

Issue #524, opened 2021-01-31
Reference global property from theme property (alias)


alehar9320:
# Intro
Let me start off by saying thanks to all the contributors of this repository. The StyleDictionary tool is wonderful and I look forward to what will follow for the Design Token community. How Style Dictionary may come to be packaged and extended to enable new innovations; with GUIs and more. From a personal perspective, I consider it almost as beneficial to the dev community as Git.

## The issue
With one of the more advanced setups for Multi Brand & Multi Platform, I'm having trouble referencing the global (general) properties from my theme restricted properties. 

### Project structure
The setup follows the structure proposed in the [official documentation for multi-theme Style Dictionary](https://github.com/amzn/style-dictionary/tree/master/examples/advanced/multi-brand-multi-platform).

The project structure follows the [documentation guidelines](https://amzn.github.io/style-dictionary/#/package_structure).

### How to re-create
Run Style Dictionary on the following definition.

#### _properties/themename/themename.json_

```JSON
  "theme": {
        "company": {
              "logo": {
                  "shell": {
                      "value": "{company.logo.url.new.full.negative.value}"
                  },
                  "welcome": {
                      "value": "{company.logo.url.new.helix.mono.white.value}"
                  }
              }
         }
    }
```
    
#### _properties/global/somefile.json
```JSON 
{
    "company": {
        "logo": {
            "url": {
                "new": {
                    "helix": {
                         "mono": {
                            "white": {
                                "value:": "#fff"
                             }
                          }
                    }
            }
       }
   }
}
```

_Note: the actual content of the value field is exceptionally long (up to hundreds of characters). Here it's been replaced by the hex code #fff. May have impact. Example: url(data:image/svg+xml;base64,adsadsdsadsadsadasdsadsadsadsadsadadsadsadsadsadsadsadsasadsadsajdhj21h3uh213h1371237ahsdhsadjhsadjsahdjsahdkjh1321h3j21h3j21h31jjsagdhgsadgsadagdhgsadhsagdhsahdg1312g3hgahsgdhasgdhgsajdgashdgsadhsagdhagdhsag131232132131sadsadasdadd4444466661337loremipsumdummydata);"_

#### Build script
```typescript

[...]

console.log('Build started...');

const themeFolderNames = getAllThemeFolderNames();

themeFolderNames.map((themeName) => {
    ['web', 'ios', 'android'].map((platform) => {
        console.log('\n==============================================');
[...]

PR #531, opened 2021-02-02
fix(export platform): fixing infinite loop when there are reference errors

dbanksdesign:
*Issue #, if available:* #524

*Description of changes:* When there were property reference errors (could not find reference or circular reference) Style Dictionary would get caught in an infinite loop. This fixes that by changing the logic of when loop within exportPlatform finishes by checking if it performed any transformations. This is because we keep the reference in the dictionary object when it is erroneous so it would keep thinking there is a deferred transform. 

### DIFF
--- a/lib/exportPlatform.js
+++ b/lib/exportPlatform.js
@@ -93,22 +93,14 @@ function exportPlatform(platform) {
     // values like "1px solid {color.border.base}" we want to
     // transform the original value (color.border.base) before
     // replacing that value in the string.
-    const transformationsCountBefore = transformedPropRefs.length;
     const transformed = transformObject(exportableResult, platformConfig, transformationContext);
 
     // referenced values, that have not (yet) been transformed should be excluded from resolving
     const ignorePathsToResolve = deferredPropValueTransforms.map(p => getName([p, 'value']));
     exportableResult = resolveObject(transformed, {ignorePaths: ignorePathsToResolve});
-    console.log(transformationsCountBefore, transformedPropRefs.length, deferredPropValueTransforms.length);
+
     // nothing left to transform -> ready
-    if (
-      // if there is no deferred value transforms left we are done
-      deferredPropValueTransforms.length === 0 ||
-      // or if we didn't transform anything since the last loop.
-      // This would happen if a token reference is not found or is circular
-      // because we keep the broken reference
-      transformedPropRefs.length === transformationsCountBefore
-    ) {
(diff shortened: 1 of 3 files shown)


lib/exportPlatform.js
```diff
@@ -93,14 +93,22 @@ function exportPlatform(platform) {
...
     // transform the original value (color.border.base) before
     // replacing that value in the string.
+    const transformationsCountBefore = transformedPropRefs.length;
     const transformed = transformObject(exportableResult, platformConfig, transformationContext);
 
     // referenced values, that have not (yet) been transformed should be excluded from resolving
     const ignorePathsToResolve = deferredPropValueTransforms.map(p => getName([p, 'value']));
     exportableResult = resolveObject(transformed, {ignorePaths: ignorePathsToResolve});
-
+    console.log(transformationsCountBefore, transformedPropRefs.length, deferredPropValueTransforms.length);
```
chazzmoney, line 102: I think this was an accidental include of some debug code?

lib/exportPlatform.js
```diff
@@ -93,14 +93,22 @@ function exportPlatform(platform) {
...
     // nothing left to transform -> ready
-    if (deferredPropValueTransforms.length === 0) {
+    if (
+      // if there is no deferred value transforms left we are done
+      deferredPropValueTransforms.length === 0 ||
+      // or if we didn't transform anything since the last loop.
+      // This would happen if a token reference is not found or is circular
+      // because we keep the broken reference
+      transformedPropRefs.length === transformationsCountBefore
+    ) {
```
chazzmoney, line 111: I trust your logic, and it makes sense...  but it seems unintuitive to me that this is a solid "or".

lib/utils/resolveObject.js
```diff
@@ -143,8 +143,7 @@ function compile_value(value, stack) {
         PROPERTY_REFERENCE_WARNINGS,
         "Reference doesn't exist: " + context + " tries to reference " + variable + ", which is not defined"
       );
-      // Leave the circular reference unchanged
-      to_ret = value;
+      to_ret = ref;
```
chazzmoney, line 146: ? help me understand this.

chazzmoney:
Do you think we need to add something to our release notes about the change in output if you have a broken reference?

Otherwise, LGTM! :shipit:

chazzmoney (approved):
Good work.

PR #533, opened 2021-02-03
feat(format): use named parameters in formatter functions

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Changed the method signature of the formatter function so that it can used named parameters rather than positional. This is done by overloading the first argument that is passed to the formatter function, so that users don't have to update their custom format code if they don't want to. This also fixes a lot of eslint errors.

TODO:
* Update documentation if this is direction we want to go.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,7 +18,12 @@
   ],
   "rules": {
     "no-console": 0,
-    "no-unused-vars": 1,
+    "no-unused-vars": [
+      "warn",
+      {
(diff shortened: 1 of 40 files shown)


chazzmoney:
This looks good to me. If this wasn't you, I'd require a third party to write up some documentation on this before we would merge it. So... can you add some documentation? 😁

chazzmoney:
LGTM! :shipit:

PR #534, opened 2021-02-03
fix(property setup): original property being mutated if the value is an object

dbanksdesign:
*Issue #, if available:*

*Description of changes:* In propertySetup we were using lodash's `clone` method which only performs a shallow clone of the object, so if a `value` was an object as well, the pristine `original` value was being mutated as well. This fixes that behavior. 


chazzmoney:
LGTM! :shipit:

Issue #545, opened 2021-02-19
Missing TypeScript types


yoniholmes:
Hello!

Thanks so much for sharing this library, nice work! 

I recently tried out using StyleDictionary with TypeScript (on the 3.0 brach), but noticed some missing types. Those are:

1. `parsers` & `transform` on the `Config` interface. This is causing a type error in the `component-cti` example.
2. `registerParser` doesn't have any types defined.
3. The Transform `options` are marked as required, but, again, in the `component-cti` example, these aren't defined, so this is causing a type error.

I'm raising a PR to address these – I thought it best to track these in a ticket.

dbanksdesign:
Thank you for this fix @yoniholmes !

PR #546, opened 2021-02-19
fix(types): introduce Parser, update Config, optional Transform options

yoniholmes:
*Issue #, if available:*
https://github.com/amzn/style-dictionary/issues/545

*Description of changes:*
1. Add `parsers` & `transform` to the Config interface. 
2. Add types for `registerParser`
3. Make `options` optional in the 3 Transform types ( `NameTransform` | `ValueTransform` | `AttributeTransform`)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -45,6 +45,8 @@ declare namespace StyleDictionary {
   }
 
   interface Config {
+    parsers?: Parser[];
+    transform?: Transforms;
     include?: string[];
     source: string[];
     platforms: { [platform: string]: Platform };
@@ -90,20 +92,20 @@ declare namespace StyleDictionary {
   interface NameTransform {
     type: "name";
     matcher?: (prop: Prop) => boolean;
-    transformer: (prop: Prop, options: Platform) => string;
+    transformer: (prop: Prop, options?: Platform) => string;
   }
 
   interface ValueTransform {
     type: "value";
     transitive?: boolean;
     matcher?: (prop: Prop) => boolean;
-    transformer: (prop: Prop, options: Platform) => string;
+    transformer: (prop: Prop, options?: Platform) => string;
   }
 
   interface AttributeTransform {
     type: "attribute";
     matcher?: (prop: Prop) => boolean;
-    transformer: (prop: Prop, options: Platform) => { [key: string]: any };
+    transformer: (prop: Prop, options?: Platform) => { [key: string]: any };
   }
 
   type Transform = NameTransform | ValueTransform | AttributeTransform;
@@ -151,6 +153,17 @@ declare namespace StyleDictionary {
   interface Filters {
     [name: string]: Matcher;
   }
+
+  interface ParserOptions {
+    contents: string;
+    filePath: string;
+  }
+
+  interface Parser {
+    pattern: RegExp;
+    parse: (props: ParserOptions) => Properties;
+  }
+
   type Named<T> = T & {
     name: string;
   };
@@ -173,6 +186,7 @@ declare namespace StyleDictionary {
     registerTemplate(this: Core, template: Named<{ template: string }>): this;
     registerAction(this: Core, action: Named<Action>): this;
     registerFilter(this: Core, filter: Filter): this;
+    registerParser(this: Core, parser: Parser): this;
 
     exportPlatform(this: Core, platform: string): Properties;
     buildPlatform(this: Core, platform: string): this;


types/index.d.ts
```diff
@@ -45,6 +45,8 @@ declare namespace StyleDictionary {
   }
 
   interface Config {
+    parsers?: Parser[];
+    transform?: Record<string, Partial<Named<Transform>>>;
```
yoniholmes, line 49: I'm not 100% sure what the expected usage of this `transform` record is. Based on the `component-cti` example, I see only `transformer` property has been defined: https://github.com/amzn/style-dictionary/blob/main/examples/advanced/component-cti/config.js#L26 (as opposed to including all possible fields: `name`, `matcher`, type`, `transitive`)

types/index.d.ts
```diff
@@ -151,6 +153,17 @@ declare namespace StyleDictionary {
...
   }
+
+  interface ParserOptions {
+    contents: string;
+    filePath: string;
+  }
+
+  interface Parser {
+    pattern: RegExp;
+    parse: (props: ParserOptions) => string | undefined;
```
yoniholmes, line 164: I'm not sure whether a Parser always must return a string in all circumstances, hence `string | undefined`.

chazzmoney: Parser should always return a plain JS object

chazzmoney: I'm a little amateur at TypeScript, but I believe it should be...

`parse: (props: ParserOptions) => Properties;`

types/index.d.ts
chazzmoney: So.... this is a substitute for calling `.registerTransform()`. It needs to match the internal structure for a transform which I believe is `Transform` in the current index.d.ts file.

chazzmoney (commented):
Thanks for this contribution and working to clean up the typescript.

I'm a bit amateur at TS, but I've tried to provide the best info I can. Let me know if anything I said doesn't make sense.

types/index.d.ts
chazzmoney: https://github.com/amzn/style-dictionary/blob/main/examples/advanced/component-cti/config.js#L40-L46

types/index.d.ts
yoniholmes: Thanks @chazzmoney, I've made this change. That makes a lot more sense!

types/index.d.ts
yoniholmes: Right... so I'm still a bit unclear about this I'm afraid! 

`registerTransform` takes `Named<Transform>>` as it's input: https://github.com/amzn/style-dictionary/blob/3.0/types/index.d.ts#L170 (i.e. the Transform, plus a field called `name`). 

However, `transform` in `Config` appears to take an object that maps the names of the transforms to the transform definitions. E.g.:

```ts
transform: {
  'attribute/cti': CTITransform
},
```

I typed this as `Record<string, Partial<Named<Transform>>>`. It's kind of equivalent to this which already exists:

```ts
interface Transforms {
  [name: string]: Transform;
}
```

So I've changed this to use `Transforms`, which I think is better. The only gotcha I think will be that the CTI example will have a type error, because only the `transformer` field is supplied, it should really be supplying `matcher` and `type` as well. _If_ however, it's possible to provide only some of those fields, then we should probably reflect this with `Partial` as in the original example (with or without `Named` as required). Sorry that's probably all rather confusing! 

Take a look at the new version and let me know what you think.

types/index.d.ts
```diff
@@ -173,6 +186,7 @@ declare namespace StyleDictionary {
     registerTemplate(this: Core, template: Named<{ template: string }>): this;
     registerAction(this: Core, action: Named<Action>): this;
     registerFilter(this: Core, filter: Filter): this;
+    registerParser(this: Core, parser: Parser): this;
```
yoniholmes, line 189: Incidentally... On further testing, I'm not sure exactly how `registerParser` is supposed to work. The types look ok. But when I initialised a style dictionary as follows:

```ts
const myDictionary = require('style-dictionary').extend(myConfigObject)
myDictionary.registerParser(myParser)
myDictionary.buildAllPlatforms()
```

Also this didn't work:
```ts
const myDictionary = require('style-dictionary')
myDictionary.registerParser(myParser)
myDictionary.extend(myConfigObject)
myDictionary.buildAllPlatforms()
```

The my customer parsers _are not_ executed. The only way I could get this to work was to put my custom parser inside `myConfigObject`, via: ` parsers: [myParser]`. 

It looks like the `extend` function calls `combineJSON` immediately (i.e. before we've run `registerParser`), rather than waiting for `buildAllPlatforms` to run, hence `registerParser` is called too late. I don't suppose you have any examples how I'm suppose to run this?

types/index.d.ts
chazzmoney: If this is not helpful, I apologize in advance, as I'm doing like 7 things at once and have about 15 minutes left here before I have to be on childcare duty. I didn't want to leave you hanging until tomorrow though.... so hopefully this helps.

You have already seen registerTransform and registerTransformGroup, but for some additional examples, there are some great ones here (specifically in build.js): https://github.com/amzn/style-dictionary/tree/main/examples/advanced/custom-transforms

Now what does this have to do with the transform property in the config? Great question. When registerTransform is called, it simply does this: https://github.com/amzn/style-dictionary/blob/main/lib/register/transform.js#L59-L63

Wait, but that still isn't config - except, config works very similarly:
- https://github.com/amzn/style-dictionary/blob/main/bin/style-dictionary#L84
- https://github.com/amzn/style-dictionary/blob/main/lib/extend.js#L59-L138

That last link is a bit hard to interpret, but it is just modifying the StyleDictionary object and extends it by adding the configuration key/value pairs as properties (without overwriting).

Thus the type for transform should be the same in both places... I think. My amateur typescript skills showing again.

I hope this helps. If not, let me know and I can try to take a deeper look later.

dbanksdesign: Hopping in here. Taking a look at the 'component-cti' example, @yoniholmes is correct and the example is wrong (sorry!). The example should have at least a `type` attribute, `matcher` on a Transform is optional. If `matcher` is omitted, it will match all tokens. It looks like the what you have currently looks good to me. The `transform` attribute in the config takes named Transforms, and a Transform is the same whether used on there or in `.registerTransform` the only difference is where the `name` of the Transform goes. In `.registerTransform` it is added in a `name` attribute, when applied directly to the config the `name` is the key in the `transform object. That is a long-winded way of saying I think this looks good and the 'component-cti' example is wrong and should be failing.

chazzmoney: Thanks @dbanksdesign I wasn't sure about the example. Appreciate the clarity.

types/index.d.ts
dbanksdesign: Can you confirm you are on `3.0.0-rc4` version? I know there was an issue with parsers not being run on a previous release candidate version... 

I just tried it out and these both work:

```javascript
// build.js
const StyleDictionary = require('style-dictionary');
const yaml = require('yaml');

const config = {
  // .. config
};

StyleDictionary.registerParser({
  pattern: /\.yaml$/,
  parse: ({contents, filePath}) => yaml.parse(contents)
});

StyleDictionary.extend(config).buildAllPlatforms();
```

Where the script to build would be `node build.js`

```javascript
// sd.config.js
const StyleDictionary = require('style-dictionary');
const yaml = require('yaml');

const config = {
  // .. config
};

StyleDictionary.registerParser({
  pattern: /\.yaml$/,
  parse: ({contents, filePath}) => yaml.parse(contents)
});

module.exports = config;
```

Where the script to run would be `style-dictionary build --config ./sd.config.js`

You are correct that you would need to register the parsers __before__ you called `.extend()`. This is specific for parsers, where registering a transform for example could happen after `.extend()` but would have to happen before `.buildAllPlatforms()`

dbanksdesign (approved):
Great work @yoniholmes ! :shipit:

Issue #551, opened 2021-03-01
custom filter not working with outputReferences. TypeError: Cannot read property 'name' of undefined


uptonking:
# overview

- [this](https://github.com/examples-hub/issue-style-dictionary-custom-filter-error) is a simplified example repo combining ideas from `multi-brand-multi-platform` and `node-modules-as-config-and-properties`.
  - all tokens are written in commonjs.
  - simplified with color only tokens
  - only output css variables with references
  - project is built with style-dictionary v3.0.0-rc.5

- my design system has three themes. 
  - pg is used as the default theme. 
  - halfmoon and spectrum override tokens in pg.

# my goal

- A. output theme-specific global css variables without components
  - all global vars start with `--pg-g` for all themes
- B. output theme-specific component only css variables
  - all comp vars start with `--pg-c` for all themes

## my expected output:

``` 

--dist/
  --pg.css
  --halfmoon.css
  --spectrum.css
  --pg-comp/
    --button.css
    --checkbox.css
  --halfmoon-comp/
  --spectrum-comp/

```

# my problem

- goal A works
- goal B fialed. 
  - I cannot output component only css variables

## how to reproduce

- `git clone https://github.com/examples-hub/issue-style-dictionary-custom-filter-error.git`
- `npm i && npm run build`.
  - look at `dist/pg.css`, for `globalCss` platfrom, button comp `--pg-c-button-primary-background-color: var(--pg-g-color-theme-primary, #db3236);` can be outputted
  - but for `compCss` platform, button comp css variables cannot be generated

``` 

==,  {"value":"#db3236","filePath":"tokens/theme/pg/component/index.js","isSource":true,"original":{"value":"{pg-g.color.theme.primary.value}"},"name":"pg-c-button-primary-background-color","attributes":{"category":"color","type":"background"},"path":["pg-c","button","primary","background-color"]}

/build.js:55
              value = reference.name;
                                ^

TypeError: Cannot read property 'name' of undefined

```
- I haven't looked into source code details of style-dictionary.

# simplified config

``` js
const StyleDictionary = require('style-dictionary');

// custom transfromer to map tokens to component-cti
StyleDictionary.registerTransform({
  name: 'attribute/cti',
  type: 'attribute',
  transformer: (prop) => {
 
    if (prop.path[0] === prefix4c) {
      // if token starts with pg-c，map the last css property to cti
      return propertiesToCTIMap[prop.path[prop.path.length - 1]];
    }

[...]

PR #553, opened 2021-03-02
fix(references): use unfiltered dictionary for reference resolution in formats

dbanksdesign:
*Issue #, if available:* #551

*Description of changes:* WIP! This is a first attempt at fixing #551. The issue is when using a filter and outputReferences, if the filtered object does not have a referenced token, it will fail. This adds an unfiltered properties object to the dictionary sent to formats as a way around this. One potential improvement is warning users if the filtered dictionary does not have the reference?


chazzmoney:
😳🤔

This is complicated but I think you have it right. Maybe we also need to output a warning when this comes up to make sure the user knows they are doing something a bit outside standard expectations?

chazzmoney (approved):
Looks GREAT. Very nicely done. Also, I'm glad GroupMessages is getting some more exercise.

:shipit:

uptonking:
@dbanksdesign 

when i use the filter with `outputReferences` to implement multi-tier design system tokens, the warning message is **TOO LONG**.    
it would be nice if u can make it short :upside_down_face:    

maybe just top 5 of `Here are the references that are used but not defined in the file: `

Issue #411, opened 2020-04-02
Token collisions were found


The-Code-Monkey:
```
⚠️  ./temp/time.json
While building time.json, token collisions were found; output may be unexpected.
    Output name 0 was generated by:
        time.transition.0   0
        time.delay.0   0
        time.duration.0   0
    Output name 1 was generated by:
        time.transition.1   120ms
        time.delay.1   50ms
        time.duration.1   1000ms

```

There really are unwanted is there a way to remove them or just hide them from the console via the config?

dbanksdesign:
Unfortunately not at this time. It is on our backlog to add more robust log levels that would fix this: #367

The-Code-Monkey:
any ideas as to when this will be implemented.

chazzmoney:
Probably in a 3.x release.

If you have time to put together a PR, we could probably get it done sooner.

If you do want to try that route, you can search the repo for `console.log` and `GroupMessages` in the code to see where logging is generated.

chazzmoney:
One more question - you are intentionally creating tokens with multiple values and don't mind if the output is any of them?

chazzmoney:
@The-Code-Monkey just checking in here; I wanted to see if you had further input.

I'd love to understand the use case so I can implement a solution correctly. It sounds like you are generating tokens with the same name on purpose - can you clarify what the intent is?

Thanks!

dbanksdesign:
I think part of this is that for 'nested' style formats like `json/nested`, they don't use the transformed names because they keep the original object structure. The way we test for collisions is to see if all token names are unique, and if the user doesn't use any name transform, token names might not be unique which would be fine in a nested format where the token names don't matter.  A quick fix would be to apply any name transform in the configuration, it won't have any effect on the nested formats, but will get rid of the warnings because all names would be unique. See the discussion here: #489

One quick fix for this that someone suggested was to mark formats as 'nested' or not, and if it is a nested format, don't show name collision warnings.

The-Code-Monkey:
@chazzmoney we aren't actually giving them the same name these values are used being generated for a theme object for our styled-components / styled-system ui library, and we wanted to be able to call them like color.primary.0 which all works fine the only issue we have is that it thinks that time.delay.0 is the same as time.transition.0 their values might be the same but the reason they are different keys is so that they are used in the correct CSS property, and also we might want to change transitions across the board but leave the delays as is.

The-Code-Monkey:
[...]

PR #567, opened 2021-03-12
chore(logs): removing name collision warnings for nested formats

dbanksdesign:
*Issue #, if available:* #411

*Description of changes:* Hiding name collision warnings for nested formats like `json/nested` because they do not use token names and the warnings have no effect on these formats. 


chazzmoney:
LGTM! :shipit:

Issue #519, opened 2021-01-20
Overriding an existing transform


nhoizey:
Based on https://github.com/amzn/style-dictionary/blob/master/examples/advanced/component-cti/config.js where I read that "Using .registerTransform() with the same transform name, 'attribute/cti', would work as well", I tried.

But I get a "RangeError: Maximum call stack size exceeded" error, as it seems the code loops the 'attribute/cti' transform.

dbanksdesign:
Ah you are correct. The quick fix is to keep a reference of the original transformer function:

```javascript
const transformer = StyleDictionary.transform['attribute/cti'].transformer;

StyleDictionary.registerTransform({
  name: 'attribute/cti',
  type: 'attribute',
  transformer: (prop) => {
    // Only do this custom functionality in the 'component' top-level namespace.
    if (prop.path[0] === 'component') {
      // When defining component tokens, the key of the token is the relevant CSS property
      // The key of the token is the last element in the path array
      return propertiesToCTI[prop.path[prop.path.length - 1]];
    } else {
      // Fallback to the original 'attribute/cti' transformer
      return transformer(prop);
    }
  }
});
```

nhoizey:
@dbanksdesign it works like that indeed, thanks! 👍

Do you think the example should use this syntax? I can make a PR if you want.

dbanksdesign:
Yea I think we should correct this in the example so others don't get confused. If you'd like to open a PR that would be wonderful. If you can, make it on the '3.0' branch because that will be the next release. Thanks!

dbanksdesign:
This is fixed now!

PR #569, opened 2021-03-15
chore(examples): update component cti example to work with register transform

dbanksdesign:
*Issue #, if available:* #519

*Description of changes:* Updating the component-cti example so that it can be used with `.registerTransform`. Currently it would throw a max call stack error if someone tried to do it that way. 


chazzmoney (commented):
👍

PR #570, opened 2021-03-17
lodash removal squashed

chazzmoney:
Replacing lodash


__tests__/utils/flattenProperties.test.js
```diff
@@ -12,7 +12,9 @@
  */
 
 var flattenProperties = require('../../lib/utils/flattenProperties');
-var _ = require('lodash');
+const sortBy = (key) => {
```
dbanksdesign, line 15: This function 🤩

lib/utils/es6_.js
```diff
@@ -0,0 +1,160 @@
...
+    f(value, key)
+  });
+}
+
+// Note: This is a crappy version to a certain extent... don't use with Strings, for example...
+const clone = function(object) {
+  return Object.assign(new object.constructor(), object)
+}
+
+const cloneDeep = function(obj) {
```
dbanksdesign, line 35: Do we need `clone` and `cloneDeep`? Could we use the `deepExtend` function we have in utils?

dbanksdesign (approved):
Tests pass! LGTM! :shipit:

lib/utils/es6_.js
chazzmoney: clone and cloneDeep do slightly different things. deepExtend is slightly different as well - copying onto a specific object.

I think to avoid bugs and refactors it is smart to just have all three used for what they are intended.

PR #583, opened 2021-03-26
fix(combine): filePath was missing for falsey values

dbanksdesign:
*Issue #, if available:*

*Description of changes:* When merging source token files we were doing a simple truthy check for `.value` which does not work if given something like `0`. Now we are checking if the token `.hasOwnProperty('value')` instead which will return true for any value. 


chazzmoney:
LGTM :shipit:

Issue #579, opened 2021-03-25
SD 3.0 output references


chemoish:
After following https://amzn.github.io/style-dictionary/#/version_3?id=output-references, I noticed this works most of the time, but for some reason it generates the following `scss` errors for nested aliased properties. Does the output account for alias dependencies?

This throws a `SassError: Undefined variable.`.

I couldn't find a relating issue, is this a known issue?

### Config

```js
files: [
  {
    destination: '_variables.scss',
    format: 'scss/variables',
    options: {
      outputReferences: true,
    },
  },
],
```

### Font

```json
{
  "font": {
    "family": {
      "custom": {
        "value": "'Helvetica Neue', Helvetica, Arial, sans-serif"
      },

      "default": { "value": "{font.family.custom.value}", "themeable": true }
    }
  }
}
```

### Component

```json
{
  "component": {
    "alert": {
      "fontFamily": {
        "value": "{font.family.default.value}",
        "themeable": true
      }
    }
  }
}
```

### _variables.scss

```
$font-family-custom: 'Helvetica Neue', Helvetica, Arial, sans-serif;
$component-alert-font-family: $font-family-default !default;
$font-family-default: $font-family-custom !default;
```

`$font-family-default` is placed at the end of the file, but is used prior to being defined.

dbanksdesign:
Ah, it looks like in the `scss/variables` format we are sorting all the tokens to make sure references are at the end, but not accounting for nested references... We need to fix that somehow

dbanksdesign:
This is fixed and will make it out in the next release candidate we will publish this week.

dbanksdesign:
Closing this as this should be fixed in the latest release candidate of 3.0.

PR #587, opened 2021-03-29
fix(formats): update output references in formats to handle nested references

dbanksdesign:
*Issue #, if available:* #579

*Description of changes:* The `formattedVariables` format helper method was not accounting for nested references (a -> b -> c) when ordering the output thus creating invalid code for imperative languages like Sass. This should fix that.


chazzmoney:
LGTM :shipit:

PR #594, opened 2021-04-08
chore(refactor): only including lodash template function

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Directly consuming the lodash template function and not the whole library. This does mean that code inside templates can no longer use lodash methods. 


chazzmoney:
LGTM :shipit:

PR #607, opened 2021-04-20
fix(references): fixing circular reference errors

dbanksdesign:
*Issue #, if available:* #608

*Description of changes:* Due to the changes we made with for transitive transforms, the code that checks for a circular reference was breaking because it was deferring resolution and therefore created an infinite loop of trying to resolve references but deferring some. This change fixes that by keeping track of how many deferred properties there are and if that number does not decrease with a pass then there is a circular reference. We then try to resolve the object one more time, without deferring any references so that the `resolveObject` method can record the circular references.

This also adds some integration testing for console output and doing the circular reference check in a real-world integration test rather than a unit test. This is because the unit test for `resolveObject` was working correctly with circular references, but when they happened in a real world case it broke. 


chazzmoney (approved):
Looks great! Good catch :)

(Also, I love all the hard work you've put into the integration tests)

Issue #610, opened 2021-04-24
3.0.0-rc.8 broke mapName functionality in scss/map-flat format


kevinmpowell:
Had to do a recent refresh of local node modules and got version bumped from `3.0.0-rc.7` to `3.0.0-rc.8` and my tests were failing because my scss map name changed to the default `$tokens`. The `mapName` key in the config appears to be skipped. Downgrading and pinning to `3.0.0-rc.7` restored the expected functionality. 

The relevant portion of my config:
```js
platforms: {
    scss: {
      transformGroup: 'scss',
      buildPath: `${path.join(destinationPath, 'scss')}/`,
      files: [
        {
          destination: '_tokens.scss',
          format: 'scss/map-flat',
          mapName: 'design-system-tokens',  // This is the feature that's not working in rc.8
        },
      ],
    },
    ...
```

dbanksdesign:
Good catch, just opened a PR to fix it!

dbanksdesign:
This is fixed in 3.0.0-rc.9!

PR #611, opened 2021-04-26
fix(formats): bringing mapName back to scss formats

dbanksdesign:
*Issue #, if available:* #610

*Description of changes:* When we were cleaning up the built-in formats, we broke `mapName` option on the `scss/map-flat` and `scss/map-deep` formats. This brings that back.

backstreet's back


chazzmoney:
LGTM! :shipit:

Issue #615, opened 2021-04-30
Object values with references don't work with outputReferences


dbanksdesign:
Hey, @dbanksdesign. Thank you so much for the examples. 

I don't understand enough about the build lifecycle to recommend a solve, but have you tested `transitive: true` with `outputReferences: true`? In this case the `value` is an `Object` and not a `String` and an error is thrown.

* What mechanism would be required if you wanted to pass a reference through a transform?
* Does it even make sense to generate references for custom values?

https://github.com/amzn/style-dictionary/blob/7a28f40b7f44b37e565b1360683b60268d044e9e/examples/advanced/transitive-transforms/sd.config.js#L53

```js
options: {
  outputReferences: true,
},
```

**Output**

```sh
⋊> ~/g/s/e/a/transitive-transforms on 3.0 ⨯ yarn build                                                                                                                                                                              14:04:33
yarn run v1.22.10
$ style-dictionary build --config ./sd.config.js

css
{ h: '{hue.red}', s: '{saturation.1}', l: '{lightness.1}' }
/Users/c0h04ba/git/style-dictionary/examples/advanced/transitive-transforms/node_modules/style-dictionary/lib/utils/references/getReferences.js:45
  value.replace(regex, function(match, variable) {
        ^

TypeError: value.replace is not a function
    at Object.getReferences (/Users/c0h04ba/git/style-dictionary/examples/advanced/transitive-transforms/node_modules/style-dictionary/lib/utils/references/getReferences.js:45:9)
    at sorter (/Users/c0h04ba/git/style-dictionary/examples/advanced/transitive-transforms/node_modules/style-dictionary/lib/common/formatHelpers/sortByReference.js:38:34)
    at Array.sort (<anonymous>)
    at formattedVariables (/Users/c0h04ba/git/style-dictionary/examples/advanced/transitive-transforms/node_modules/style-dictionary/lib/common/formatHelpers/formattedVariables.js:55:40)
    at Object.css/variables [as format] (/Users/c0h04ba/git/style-dictionary/examples/advanced/transitive-transforms/node_modules/style-dictionary/lib/common/formats.js:47:7)
    at buildFile (/Users/c0h04ba/git/style-dictionary/examples/advanced/transitive-transforms/node_modules/style-dictionary/lib/buildFile.js:101:37)
    at /Users/c0h04ba/git/style-dictionary/examples/advanced/transitive-transforms/node_modules/style-dictionary/lib/buildFiles.js:33:7
[...]

PR #623, opened 2021-05-11
fix(references): value object references now work

dbanksdesign:
*Issue #, if available:* #615

*Description of changes:* Token values that are objects as opposed to simple strings were not working with `outputReferences` and throwing an error. These changes fix some behavior and make sure certain scenarios don't fail. This makes explicit what `outputReferences` can and can't do:
* The `value` must be a string at time of formatting. This should be the case anyways, if not you will get `[object Object]` in output files. With `outputReferences` this was failing because an object does not have the `.replace` function. 
* The `value` must have the referenced values in it for `outputReferences` to replace the resolved values with the reference's name. For example if you define a color using an HSL object and transform it to a hex value, we cannot undo that transform to understand which parts of the value to use the reference's name. 

Here are the changes:
* In `createPropertyFormatter`, rather than looking at the *original* value and replacing references with the referenced token's name, we are now looking at the *transformed* value and replacing any referenced *values* with the reference's name. This allows us to still use `outputReferences` with a value object. 
    * If the transform that converts the value object to a string mutates the reference so that it is not in the transformed value, it will not work with `outputReferences`. It will build fine, but just won't output the reference and instead output the transformed and resolved value. An example of this is converting an HSL object to a Hex string.
    * You can reference a non-token as seen in the transitive-transforms example. References to non-token's will not work with `outputReferences` because the references will not have a `name` attribute. Just like the previous point, it will build fine, but just won't output the reference and instead output the transformed and resolved value. 
* Built-in formats expect a token's value to be a string at the time of formatting. This means if the token's value is an object it needs to be transformed to a string
* Updated `getReferences` to mirror `usesReference` with respect to object values. `usesReference` checks the object values, whereas `getReferences` did not. 

Added some unit tests to `getReferences` to verify behavior.
Added integration tests to verify different behaviors:
* Using an object value with references, transformed to a string where reference is present (css border short-hand). With and without `outputReferences`
* Using an object value with references to non-tokens
* Using an object value with references, transformed to a string where reference is not present (hsl -> hex)

### DIFF
--- a/__integration__/__snapshots__/objectValues.test.js.snap
+++ b/__integration__/__snapshots__/objectValues.test.js.snap
@@ -1,0 +1,93 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`integration object values css/variables border should match snapshot 1`] = `
+"/**
+ * Do not edit directly
+ * Generated on Sat, 01 Jan 2000 00:00:00 GMT
+ */
+
+:root {
+  --border-primary: 0.125rem solid #ff0000;
+}
+"
+`;
+
+exports[`integration object values css/variables border with references should match snapshot 1`] = `
+"/**
+ * Do not edit directly
+ * Generated on Sat, 01 Jan 2000 00:00:00 GMT
+ */
+
+:root {
+  --border-primary: var(--size-border) solid var(--color-red);
+}
+"
+`;
+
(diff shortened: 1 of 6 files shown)


chazzmoney (approved):
Nice work. For being some of the most complex code in all of SD, this is almost readable!

PR #635, opened 2021-05-23
docs(examples): fix source reference and properties function

jdelight:
Found a few small typos prevent the custom transforms examples from working.

### DIFF
--- a/examples/advanced/custom-transforms/config.json
+++ b/examples/advanced/custom-transforms/config.json
@@ -1,5 +1,5 @@
 {
-  "source": ["properties/**/*.json"],
+  "source": ["tokens/**/*.json"],
   "platforms": {
     "web": {
       "transformGroup": "custom/web",


examples/advanced/custom-transforms/build.js
```diff
@@ -96,7 +96,7 @@ StyleDictionary.registerTransformGroup({
 StyleDictionary.registerFormat({
   name: 'custom/android/xml',
   formatter: function(dictionary) {
-    return dictionary.allTokens.map(function(token) {
+    return dictionary.allProperties.map(function(token) {
```
dbanksdesign, line 99: This one is my bad. We are moving to using "design tokens" and "tokens" rather than "properties" in 3.0. This change will be backwards compatible (you can still use `allProperties` here), but we wanted to update the examples to reflect this change. The reason this doesn't work is I didn't push a pre-release version after making that change. If you rebase (I pushed a pre-release version) and change this back to `allTokens` and update line 98 to use a de-structured argument: `formatter: function({ dictionary }) {`

examples/advanced/custom-transforms/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "source": ["properties/**/*.json"],
+  "source": ["tokens/**/*.json"],
```
dbanksdesign, line 2: 🤦 thanks for catching this!

dbanksdesign (commented):
One small comment that was my bad, but if you can make that change super quick we can get this in today.

examples/advanced/custom-transforms/build.js
jdelight: ah ok - will revert this 👍

dbanksdesign (approved):
LGTM! :shipit:

PR #639, opened 2021-05-25
chore(docs): updating docs for 3.0 launch

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Updating some of the docs to reflect the 3.0 launch

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #640, opened 2021-05-25
Make "Style Properties → Design Tokens" a heading

nhoizey:
*Description of changes:*

"Style Properties → Design Tokens" was not a heading, I think it should be.


chazzmoney:
Thank you - good catch! It is being resolved in #639

PR #653, opened 2021-06-17
docs: Fix "getReferences" typos

kylegach:
*Issue #, if available:*

*Description of changes:*

- Change "getReference" to "getReferences"
- Update documented return value to array of original tokens
- Update code snippets to access array of tokens instead of single token

The [way `refs` are iterated over here](https://github.com/kylegach/style-dictionary/blob/b12c4b1c6a94c62c757cd1675d216d9638f8d6e0/lib/common/formatHelpers/createPropertyFormatter.js#L95-L116), implies that simply accessing the first token in the array may not be a good strategy. All tests & builds pass, though (including those in the "variables-in-outputs" example, where I changed actual code, not just docs).

### DIFF
--- a/docs/version_3.md
+++ b/docs/version_3.md
@@ -164,8 +164,12 @@ StyleDictionary.registerFormat({
       // an array of references to the whole tokens so that you can access
       // their names or any other attributes.
       if (dictionary.usesReference(token.original.value)) {
-        const references = dictionary.getReferences(token.original.value);
-        value = references[0].name;
+        const refs = dictionary.getReferences(token.original.value);
+        refs.forEach(ref => {
+          value = value.replace(ref.value, function() {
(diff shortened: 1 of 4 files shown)


dbanksdesign:
Thank you for catching this and fixing it! I totally agree with your comment about the example code accessing the first item in the array directly rather than iterating over the array like we do in `createPropertyFormatter`. If you can, could you update those examples to iterate over the references? It probably doesn't need to be that entire block, but something more simple like

```javascript
refs.forEach(ref => {
  value = value.replace(ref.value, function() {
    return `${ref.name}`;
  });
});
```

kylegach:
Thanks for the guidance, Danny. Just pushed suggested changes.

chazzmoney (approved):
LGTM! :shipit:

Issue #652, opened 2021-06-17
v3, registerFormat formatter: `tokens` & `allTokens` are not available on `dictionary` positional argument


kylegach:
This works:

```js
StyleDictionary.registerFormat({
  name: 'javascript/es6DefaultExport',
  formatter: ({ dictionary }) => {
    return [
      'const tokens = {',
      ...dictionary.allTokens.map(
        // ...format logic
      ),
      '};',
      'export default tokens;',
    ].join('\n');
  },
});
```

But this does not:

<details>
<summary>Full error</summary>

```
/Users/.../packages/tokens/scripts/build-tokens.js:19
      ...dictionary.allTokens.map(function (prop) {
                              ^

TypeError: Cannot read property 'map' of undefined
    at Object.formatter [as format] (/Users/.../packages/tokens/scripts/build-tokens.js:19:31)
    at buildFile (/Users/.../node_modules/style-dictionary/lib/buildFile.js:103:37)
    at /Users/.../node_modules/style-dictionary/lib/buildFiles.js:33:7
    at Array.forEach (<anonymous>)
    at buildFiles (/Users/.../node_modules/style-dictionary/lib/buildFiles.js:31:18)
    at Object.buildPlatform (/Users/.../node_modules/style-dictionary/lib/buildPlatform.js:62:3)
    at /Users/.../node_modules/style-dictionary/lib/buildAllPlatforms.js:31:10
    at Array.forEach (<anonymous>)
    at Object.buildAllPlatforms (/Users/.../node_modules/style-dictionary/lib/buildAllPlatforms.js:30:39)
    at Object.<anonymous> (/Users/.../packages/tokens/scripts/build-tokens.js:51:17)
```

</details>

```js
StyleDictionary.registerFormat({
  name: 'javascript/es6DefaultExport',
  formatter: (dictionary) => {
    return [
      'const tokens = {',
      ...dictionary.allTokens.map(
        // ...formatting logic
      ),
      '};',
      'export default tokens;',
    ].join('\n');
  },
});
```

-----

[...]

PR #655, opened 2021-06-22
fix(lib): Fix `createFormatArgs` positional args

kylegach:
*Issue #, if available:* #652

*Description of changes:*

- Update `createFormatArgs` to return `allTokens` and `tokens` properties along with `properties` and `allProperties`
- Update snapshots
    - It's hard to confirm in the diff, but I manually confirmed that the newly-added values for `allTokens` are the same as `allProperties`, and `tokens` are the same as `properties`

*Note:*

Correcting the types will be done in a future PR.

### DIFF
--- a/__integration__/__snapshots__/customFormats.test.js.snap
+++ b/__integration__/__snapshots__/customFormats.test.js.snap
@@ -566,6 +566,166 @@ exports[`integration custom formats inline custom with new args should match sna
       }
     }
   },
+  \\"allTokens\\": [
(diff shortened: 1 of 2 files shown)


chazzmoney (approved):
Thanks for catching this - this is how we should have done it. Double thanks for understanding our intent and fixing it to work that way!

LGTM! :shipit:

dbanksdesign (approved):
LGTM! :shipit:

PR #657, opened 2021-06-22
fix(types): Add overloaded formatter fn definiton

kylegach:
*Issue #, if available:* #652

*Description of changes:*

- `formatter` type now accepts either positional arguments or a single object argument

*Note:*

I didn't see an easy way to typecheck usages of `formatter` within the project, but I did make these changes in my local `node_modules` of another project and confirmed that the types seemed to be correct there.

I'm not familiar with `tsd`, so I'm unsure why the type tests are failing or how to fix them.

### DIFF
--- a/types/Format.d.ts
+++ b/types/Format.d.ts
@@ -36,10 +36,17 @@ export interface FormatterArguments {
 }
 
 /**
- * The formatter function receives an overloaded object as its arguments and
+ * The formatter function receives positional arguments and it should return a
+ * string, which will be written to a file.
+ */
+declare function formatter(dictionary: Dictionary, platform: Platform, file: File): string;
+/**
+ * The formatter function receives an overloaded object as its argument and
  * it should return a string, which will be written to a file.
  */
-export type Formatter = (arguments: FormatterArguments) => string;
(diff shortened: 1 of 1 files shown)


chazzmoney:
Hmmm.... I see the need for this. However, while the old style usage is not officially deprecated, I'd like to avoid encouraging people to use it in new code... which I feel like adding a formal type for the old style usage might do... or maybe I just need to rethink that we should have two function signatures... we definitely believe in not breaking old work... hmmm...

🤔

kylegach:
> However, while the old style usage is not officially deprecated, I'd like to avoid encouraging people to use it in new code... which I feel like adding a formal type for the old style usage might do... or maybe I just need to rethink that we should have two function signatures... we definitely believe in not breaking old work...

I certainly get not wanting to promote even unofficially deprecated features. I think only documenting `tokens`/`allTokens` sufficiently serves that purpose. Especially because 3.0 is _not_ currently backwardly-compatible for TypeScript users. It's a very straightforward migration, but still.

That said, unless someone can unblock this PR by fixing the tests, perhaps that settles the dilemma for you. 🙃

PR #681, opened 2021-08-14
fix(format): 'typescript/es6-declarations' return type

Dru89:
This change fixes an issue where all variables exported from 'typescript/es6-declarations' are assumed to be of type `string`.

Before this, if you had an input type like:
```json
{
  "color": {
    "blue": { "value": "#0000ff" },
  },
  "size": {
    "padding": { "value": 2 }
  }
}
```

Your output ES6 module would be:
```js
export const ColorBlue = "#0000ff";
export const SizePadding = 2;
```

But your output TypeScript module would be:
```ts
export const ColorBlue : string;
export const SizePadding : string;
```

This adds some *basic* type checking for a few known values, and will fall back to `any` if the type can't be easily determined. It will look for:

- strings
- numbers
- booleans
- or arrays of any of those things

If it can't determine the type, it will fall back to `any` (or `any[]` if we know it's an array).

### DIFF
--- a/__tests__/common/formatHelpers/getTypeScriptType.test.js
+++ b/__tests__/common/formatHelpers/getTypeScriptType.test.js
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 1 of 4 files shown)


chazzmoney (approved):

dbanksdesign (approved):
LGTM! :shipit:

Issue #678, opened 2021-08-10
getReferences should check if object values are strings first


dbanksdesign:
This is because the values could also be numbers, but this would throw an error.

dbanksdesign:
This is now fixed in v3.0.2

PR #682, opened 2021-08-19
fix(references): check if object value is a string before replacement

dbanksdesign:
*Issue #, if available:* #678

*Description of changes:* In the `getReferences` function first check if the values within an object are strings before trying to run `.replace` on them. It is valid to have a non-string value within an object, just like it is valid to have a non-string as a regular value (like a number for a dimension).


chazzmoney (approved):
LGTM! :shipit:

PR #690, opened 2021-08-31
fixed typo

lukasoppermann:


chazzmoney (approved):

dbanksdesign (changes requested):
Thanks for catching this! This file is actually being generated from JSDoc, so if you could update the JSDoc comment here: https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L644 first and then run `npm run generate-docs` that should make sure this change doesn't get thrown out when we generate the docs in the future. Thanks!

lukasoppermann:
Hey @dbanksdesign I changed it according to your comment, hope this is correct. Cheers

dbanksdesign (approved):
LGTM! :shipit:

PR #691, opened 2021-09-01
Fixed doc anchor

CharlyTorregrosa:
`Pre-defined transforms` anchor was pointing to `defining-custom-transforms`


chazzmoney (changes requested):
Thanks for catching this! This file is actually being generated from a handlebar template, so if you could update the anchor here: https://github.com/amzn/style-dictionary/blob/main/scripts/handlebars/templates/transforms.hbs#L3 first and then run npm run generate-docs that should make sure this change doesn't get thrown out when we generate the docs in the future. Thanks!

CharlyTorregrosa:
Closed in favor of https://github.com/amzn/style-dictionary/pull/699

PR #505, opened 2020-12-12
refactor(transforms): add platform option 'basePxFontSize'

jbarreiros:
*Issue #, if available:*
NA

*Description of changes:*
Per [this conversation](https://github.com/amzn/style-dictionary/pull/491#discussion_r533583145), this PR adds a `basePxFontSize` platform option, and updates each `remTo*` transform to use it.

Also added missing unit tests for `size/swift/remToCGFloat`.

I realize that either this PR or  #491 will need a quick update, depending which is merged first. Let me know what you want me to do.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #715, opened 2021-10-13
Add base px font size to platform type definition

didoo:
This is a follow-up of #505. 

I have added the proper type definition for basePxFontSize to the Platform type.

/cc @dbanksdesign I am not sure if the branch to merge into is 3.0, in case I think you can update it.


dbanksdesign:
I changed the branch to main, thanks for this change!

Issue #721, opened 2021-10-19
Allow references without .value


markacianfrani:
Currently, for outputReferences to work, you need to define your reference using .value like:
```
      "base"  : { "value": "{size.font.medium.value}" }
```

If your reference doesn't include .value it won't work:
```
    "base"  : { "value": "{size.font.medium}" }
```

The [design tokens working group first draft](https://design-tokens.github.io/community-group/format/#aliases-references) actually has the reference defined without .value
```
{
  "group name": {
    "token name": {
      "value": 1234
    }
  },
  "alias name": {
    "value": "{group name.token name}"
  }
}
```

Would you be open to supporting references without .value to better align with the spec's working draft? 

From what I can tell, we'd just need to modify around line 91 of https://github.com/amzn/style-dictionary/blob/main/lib/common/formatHelpers/createPropertyFormatter.js to maybe check if it's object and if the object has a value property, use that instead. 

Would be happy to put in a PR if you think that's a good approach.

phantasmagoriadigital:
+1 This would be really helpful to synchronise our tokens between figma tokens and style dictionary. That way we can keep all our tokens as per W3C spec.

esthercheran:
+1 Would be really helpful for us too.

zetareticoli:
+1 Would be really useful to stay aligned with Figma plugin and Style Dictionary output

chazzmoney:
Closed via #746

PR #723, opened 2021-10-24
Support for references without .value

markacianfrani:
https://github.com/amzn/style-dictionary/issues/721 

The issue above does a good job explaining the problem. If you omit the `.value` in your token format (as is the case when using the Figma Tokens plugin and also the W3 spec), SD fails the reference stage and results in [object object]. Looking at the object returned if you have this omission, there should be a .value property that you can just reference.

However, this seems way too simple so I'm wondering what edge cases or scenarios I might be overlooking here? 


markacianfrani:
Well, the good news is the test suites are really good and the bad news is....the test suites are really good. 

I see now that this might be way more complicated than I originally thought. It's the formatters that are giving me trouble now (compose/object for starters). 

Back to the drawing board!

lib/utils/flattenProperties.js
```diff
@@ -29,6 +29,15 @@ function flattenProperties(properties, to_ret) {
     if (properties.hasOwnProperty(name)) {
       // TODO: this is a bit fragile and arbitrary to stop when we get to a 'value' property.
       if (_.isPlainObject(properties[name]) && ('value' in properties[name])) {
+
+        /**
+         * If .value is omitted from the input
+         * ie value: "{color.core.purple}" instead of "{color.core.purple.value}"
+         */
+        if (properties[name].value.hasOwnProperty('value')) {
```
markacianfrani, line 37: I don't love putting this here, since this is flattenProperties function is supposed to be as generic as possible.

markacianfrani:
Had to go way farther up the chain for this. Doing it at the format level caused all sorts of headaches. 

I did find another edge case--in the current test suite, if you were to change, color.font.interactive.active to `color.brand.secondary`, it causes all sorts of side effects. Mixing and matching using .value and not using .value is going to be difficult I think. Although, I really don't see a legit use case where you would want to be able to use both at the same time. 

But is this even something we want to solve at the Style Dictionary level? I'm totally okay with closing this and solving it at the Figma Tokens Plugin level instead.

JackHowa:
this looks like it'd be really helpful!

crypt096 (approved):

lukasoppermann:
Hey, can this be merged? I am working on references for my [figma plugin](https://github.com/lukasoppermann/design-tokens/pull/146) which exports mainly for style dictionary. However I would love to keep it in line with the w3c spec. If thus could be merged I wouldn't need to include some hacky solution that adds `.value`.

@dbanksdesign ?

dbanksdesign:
Hey everyone! Sorry I haven't been responsive on this PR, my team just had a big launch at re:invent last week that took all of my attention. I am taking a closer look at this PR today and I would like to get this merged in this week and hopefully do a release.

dbanksdesign:
I took a look at this and figured it would be easier to make the change because it requires change code in a more core place. @markacianfrani I'll list you as the co-author though to give you credit. Take a look at #746 and thanks for your patience!

dbanksdesign:
Closing this in favor of #746 but we will give you credit in the release notes @markacianfrani!

PR #730, opened 2021-11-05
chore(docs): adddocs reference to playground

jorenbroekema:
As discussed with @dbanksdesign  I would open a PR to add a reference to the playground in the docs.

I took the liberty to add a CTA button for it on the coverpage, let me know if that's okay with you guys, I'm obviously a bit biased to saying it's important enough to be there ;)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,6 +18,9 @@ When you are managing user experiences, it can be quite challenging to keep styl
 ## Watch the Demo on Youtube
 [![Watch the video](assets/fake_player.png)](http://youtu.be/1HREvonfqhY)
 
+## Experiment in the playground
+Try the browser-based Style Dictionary playground: [https://www.style-dictionary-play.dev/](https://www.style-dictionary-play.dev/), built by the folks at [\<div\>RIOTS](https://divriots.com/).
+
 ## Examples
 [See examples of Style Dictionary here](examples.md)
 
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ For detailed usage head to https://amzn.github.io/style-dictionary
 ## Watch the Demo on Youtube
 [![Watch the video](/docs/assets/fake_player.png)](http://youtu.be/1HREvonfqhY)
 
+## Experiment in the playground
+Try the browser-based Style Dictionary playground: [https://www.style-dictionary-play.dev/](https://www.style-dictionary-play.dev/), built by the folks at [\<div\>RIOTS](https://divriots.com/).
+
 ## Contents
 * [Installation](#installation)
 * [Usage](#usage)


docs/_coverpage.md
```diff
@@ -10,5 +10,6 @@
 [Get Started](README.md)
 
 [See what's new in 3.0!](version_3.md)
+[Try in Playground](https://www.style-dictionary-play.dev/)
```
dbanksdesign, line 13: Could we remove this from the cover page? I don't think we are quite ready for that placement.

docs/README.md
```diff
@@ -18,6 +18,9 @@ When you are managing user experiences, it can be quite challenging to keep styl
 ## Watch the Demo on Youtube
 [![Watch the video](assets/fake_player.png)](http://youtu.be/1HREvonfqhY)
 
+## Experiment in the playground
+The folks at [\<div\>RIOTS](https://divriots.com/) created a [browser playground](https://www.style-dictionary-play.dev/) that's great if you want to experiment or test something out.
```
dbanksdesign, line 22: Could we flip this sentence around a bit? Maybe:

```
Try the browser-based Style Dictionary playground: [https://www.style-dictionary-play.dev/](https://www.style-dictionary-play.dev/), built by the folks at [\<div\>RIOTS](https://divriots.com/)
```

dbanksdesign (commented):
Just some minor content changes, thanks for submitting this!

docs/_coverpage.md
jorenbroekema: Can imagine ;) worth a try! I removed it

docs/README.md
jorenbroekema: Yeah that reads better, thanks, adjusted!

chazzmoney:
@jorenbroekema This is pretty cool, thanks for the work the team did on putting it together!

If I may, I'd like to suggest some potential work that could be done on it to make it even better (and SD coverpage worthy):
1. Add some visual rendering utilizing the SD components so people can see that when they change the dictionary, the css and the html styles update.
2. Need some way to show the multi-platform output. So web and mobile-web maybe? Like a fake webpage view and a mobile phone view.

These additions would show to new users what style dictionary is intended to be used for, while still enabling more pro users to copy/paste in content and use it more like a testing/dev tool.

Whether you decide to take on the work or not, I want to acknowledge your appreciation of SD as a tool worthy of spending time on. We really appreciate our community and love what you guys have built. So thanks, for being you!

jorenbroekema:
@chazzmoney  great suggestions! Definitely willing to do those, I have a good idea for a simple component which would have a mobile variant (<600px viewport) that looks slightly different, would be a good showcase 👍  

Users can just delete the token file for the "showcase" component (also deleting the component view) if they don't want to play with that and just dev/test something else.

jorenbroekema:
https://deploy-preview-13--style-dictionary-playground.netlify.app/  here's a preview.

It's especially nice to pop it out and then live edit the tokens and see the changes as you make them. Hopefully this comes close to what you guys had in mind, let me know if this makes it coverpage worthy and I can re-add the button :)

chazzmoney:
Thanks for building the previewer to show off the changes occurring - this is getting close! A few recommendations:

1. The requirement to scroll down to see the preview makes it too disjoint. Even though I **knew** it was supposed to be a preview, I was still confused. Is there a way to make it more directly connected? Can we get it within the same horizontal screen by default?
2. When it initially loads, all the files on the left are hidden, but tokens/card/card.json is being viewed. Maybe this is just cookies on my machine, but I'm confused about what's going on, or how to edit / play with it. It might be good to have all the toggles open by default so people know which file they are viewing (and what the file options are). Maybe the same goes for the build files - so people can see what is being built?
3. The tokens/color/font.json file isn't referenced anywhere. This is ok, but also feels confusing to me. Maybe remove it to simplify? Or use those tokens in the tokens/card/card.json instead of the base colors tokens?

Thanks again for your work here!

jorenbroekema:
2 and 3 are indeed good ones, easy to fix. 

For 1, would it suffice if I popped it out by default so it's overlaying the editor on the right side on initial load? A friend of mine mentioned that he'd like it to be popped out by default as well.

I think it makes sense that the editor has the center and is the primary element, I tried putting the preview next to it but that didn't look particularly nice.

jorenbroekema:
Updated the playground 👍 thanks again for the feedback, it's looking better and better!

chazzmoney (approved):
I think this is good enough to be added. Thanks for your hard work!

Issue #719, opened 2021-10-19
[Question] Can I generate extensions in iOS Swift?


antoniogamiz:
Hi, I have been using Style Dictionary and it's great! Thanks a lot for the hard work.

I have a question. I have the following configuration file to generate styles sheets on iOS Swift from design tokens hosted on Invision App:

~~~json
{
  "source": [
    "tokens/**/*.json"
  ],
  "platforms": {
    "ios-swift-separate-enums": {
      "transformGroup": "ios-swift-separate",
      "buildPath": "build/ios-swift/",
      "files": [
        {
          "destination": "StyleDictionaryColor.swift",
          "format": "ios-swift/enum.swift",
          "className": "StyleDictionaryColor",
          "filter": {
            "attributes": {
              "category": "color"
            }
          }
        },
        {
          "destination": "StyleDictionarySize.swift",
          "format": "ios-swift/enum.swift",
          "className": "StyleDictionarySize",
          "type": "float",
          "filter": {
            "attributes": {
              "category": "size"
            }
          }
        }
      ]
    },
  }
}
~~~

The result is something like this:

~~~swift
import UIKit

public enum StyleDictionaryColor {
    public static let basicGrayDark = UIColor(red: 0.345, green: 0.345, blue: 0.345, alpha: 1)
    public static let basicGrayDark60 = UIColor(red: 0.000, green: 0.000, blue: 0.000, alpha: 0.6)
}

~~~

But, I would like to generate something like this: 

~~~swift
extension Color {
    static let oldPrimaryColor = Color(UIColor.systemIndigo)
    static let darkGreyColor = Color("darkGrey")
}
~~~

Is that currently possible? Or would I need to create my own transformer? Thanks in advance (if this is not the place for this type of question, please redirect to the correct place if you which is, thanks! :))

antoniogamiz:
Okay, I will try to implement it in the near future! Thanks for the fast response :smiley: .

antoniogamizbadger:
PR made for the SwiftUI Color transformer, let me now if I need change something.

tfmart:
In case anyone wants an example on how to create a Swift extension file, on your iOS swift file object in `config.json`, use `ios-swift/any.swift` as the format and add a `options` object, in which you can provide a `objectType` with a `extension` value. Here's an example:
```json
{
  "source": [
    "tokens/**/*.json"
  ],
  "platforms": {
    "ios-swift": {
      "transforms": [
        "..."
      ],
[...]

PR #733, opened 2021-11-12
feat(ios-swift): add transformer for Color class from SwiftUI

antoniogamizbadger:
*Issue #719*

*Description of changes:*

I have added a new color formatter to correctly render `Color` from SwiftUI. See the [documentation](https://developer.apple.com/documentation/swiftui/color).

It would be used like this:

~~~json
  "ios-swift-separate-enums": {
    "transforms": [
      "attribute/cti",
      "name/ti/camel",
      "color/ColorSwiftUI",
      "content/swift/literal",
      "asset/swift/literal",
      "size/swift/remToCGFloat",
      "font/swift/literal"
    ],
    "buildPath": "build/ios-swift/",
    "files": [
      {
        "destination": "StyleDictionaryColor.swift",
        "format": "ios-swift/extension.swift",
        "className": "Color",
        "filter": {
          "attributes": {
            "category": "color"
          }
        }
      },
      {
        "destination": "StyleDictionarySize.swift",
        "format": "ios-swift/extension.swift",
        "className": "CGFloat",
        "filter": {
          "attributes": {
            "category": "size"
          }
        }
      }
    ]
  },
~~~

We need to specify the tranformers using `transforms` because there is not a transform group defined using this new transformer.

### DIFF
--- a/__tests__/common/transforms.test.js
+++ b/__tests__/common/transforms.test.js
@@ -416,6 +416,29 @@ describe('common', () => {
       });
     });
 
+    describe('color/ColorSwiftUI', () => {
+      it('should handle normal colors', () => {
+        var value = transforms["color/ColorSwiftUI"].transformer({
+          value: "#aaaaaa"
+        });
+        expect(value).toBe("Color(red: 0.667, green: 0.667, blue: 0.667, opacity: 1)");
+      });
+
+      it('should retain enough precision when converting to decimal', () => {
+        var value = transforms["color/ColorSwiftUI"].transformer({
+          value: "#1d1d1d"
+        });
+        expect(value).toBe("Color(red: 0.114, green: 0.114, blue: 0.114, opacity: 1)");
+      });
+
+      it('should handle colors with transparency', () => {
+        var value = transforms["color/ColorSwiftUI"].transformer({
+          value: "#aaaaaa99"
+        });
+        expect(value).toBe("Color(red: 0.667, green: 0.667, blue: 0.667, opacity: 0.6)");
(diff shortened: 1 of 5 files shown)


antoniogamizbadger:
I do not know why the linter has deleted all those blank spaces, maybe those markdown files are autogenerated and I should not have updated them. Should I delete those changes?

chazzmoney:
Very dumb question because I am not familiar enough with swift UI...

Does having 3-digit precision matter significantly? Or would it be more future-proof to enable a longer precision? If so, then maybe a higher value for toFixed() could be worthwhile?

If not, then I think it looks good.

__tests__/common/transforms.test.js
```diff
@@ -416,6 +416,29 @@ describe('common', () => {
       });
     });
 
+    describe('color/ColorSwiftUI', () => {
+      it('should handle normal colors', () => {
+        var value = transforms["color/ColorSwiftUI"].transformer({
+          value: "#aaaaaa"
+        });
+        expect(value).toBe("Color(red: 0.667, green: 0.667, blue: 0.667, opacity: 1)");
```
dbanksdesign, line 424: Thanks for adding unit tests!

dbanksdesign (commented):
LGTM! Thank you for this contribution! I think just answering @chazzmoney's question and we should be good to go! (I don't know about those whitespace changes, but its fine)

antoniogamizbadger:
@chazzmoney in my compony we never use more than 3 digits, so I think there's not a problem with that.

Issue #688, opened 2021-08-30
Error on token collision


goodmorninggoaway:
I've seen a few tickets where users want to silence collision warnings. I'd like to do the opposite: Turn them into errors and fail the build. Is this on the roadmap, or would you accept a PR?

chazzmoney:
Hi!

We had not considered the vast array of different error handling situations; we would love a suggestion on how best to support a range of desired behaviors on error states - a PR would be most welcome!

goodmorninggoaway:
I wouldn’t want to mess with the generally unopinionated architecture, but maybe a strict mode where some of the cations are enforced or general opt in. The two that come to mind are collision warnings and remembering to suffix references with `.value`. But both of those are at time desirable, so how about platform-scopes options to error?

If this approach works, I can make a PR

chazzmoney:
Sorry for the delay here, but I've been pondering this a bit with @dbanksdesign .

There are three "scopes" that SD is in during processing: everything, platform, and file.
There are three "settings" that any error should be able to have: ignore, warn, fail.
There are several different kinds of errors in SD, most of them listed in https://github.com/amzn/style-dictionary/blob/main/lib/utils/groupMessages.js

It seems like maybe we should have a specific configuration for each error type... (and maybe an "all errors")? This would be able to be set at each scope in the config, with each more specific level overwriting the choices of the more general level. Each error would then be handled as configured. Defaults would be "warn" for all errors.

Do you think this would enable you to meet your needs? Do you think it remains unopinionated?

Thanks for your thoughts!

goodmorninggoaway:
Yes it would

chazzmoney:
@goodmorninggoaway Are you thinking about a PR here still? We would definitely love to have it - even a really bad v0.0.1 would be appreciated!

goodmorninggoaway:
I’m sure I can put together a really bad PR, possibly next week

chazzmoney:
yay! Looking forward to it!

MildTomato:
Would also like to filter out warning messages.

<img width="1706" alt="Screenshot 2023-10-10 at 2 34 18 AM" src="https://github.com/amzn/style-dictionary/assets/8291514/3301681a-8494-46eb-892f-80f23bc13b62">

We output multiple files with these in, so the reference not being available is fine.

PR #735, opened 2021-11-18
Error on output name collisions based on file-level option

goodmorninggoaway:
*Issue #, if available:*
#688 

*Description of changes:*
After looking through tests, I found that the core fucntionality of #688 already appears to be implemented if `.extend` is used. This PR implements it at the file level, so that setting `options: { log: 'error' }` will error if there are collisions. Setting `options: { log: 'warn' }` or excluding that option retains existing behavior. 

**Question**
Where should this be documented?


chazzmoney:
Thanks @goodmorninggoaway !

This is good thinking material, and I need to think on it a bit more. I think we might need deeper configuration options, but I need to spend more time on it before I can say something more solid.

goodmorninggoaway:
Yeah, it's more basic than I anticipated. I think most of the top level apis would need to change to accept new configs, esp since options can't be too prescriptive. Which seems like a breaking change that I didn't want to introduce.

Issue #727, opened 2021-11-02
[Question] - Is there a way to not include default metadata?


delisma:
## Issue description
I was trying to build a JSON file for the Figma Tokens plugin written by @six7.  

When a `name` property is declared for a token in the JSON file, I've noticed that the plugin use it for organizing the tokens with the type property. But when no `name` property is included, the keys of the token is used instead. 

## Current behaviour
When I use the pre-defined JSON transform I get this verbose data structure
```json
{
  "colour": {
    "utils": {
      "blue": {
        "value": "#0535D2",
        "type": "color",
        "filePath": "tokens/color/base.js",
        "isSource": true,
        "original": {
          "value": "#0535D2",
          "type": "color"
        },
        "name": "gcds-colour-utils-blue",
        "attributes": {},
        "path": [
          "colour",
          "utils",
          "blue"
        ]
      },
      "purple": {
        "value": "#7834BC",
        "type": "color",
        "filePath": "tokens/color/base.js",
        "isSource": true,
        "original": {
          "value": "#7834BC",
          "type": "color"
        },
        "name": "gcds-colour-utils-purple",
        "attributes": {},
        "path": [
          "colour",
          "utils",
          "purple"
        ]
      },
      "white": {
        "value": "#ffffff",
        "type": "color",
        "filePath": "tokens/color/base.js",
        "isSource": true,
        "original": {
          "value": "#ffffff",
          "type": "color"
        },
        "name": "gcds-colour-utils-white",
        "attributes": {},
        "path": [
          "colour",
          "utils",
          "white"
        ]
      }
    },
    "base": {
      "red": {
        "100": {
          "value": "#ecb6b5",
          "type": "color",
          "filePath": "tokens/color/base.js",
          "isSource": true,
          "original": {
            "value": "#ecb6b5",
            "type": "color"
          },
          "name": "gcds-colour-base-red-100",
          "attributes": {},
          "path": [
            "colour",
            "base",
            "red",
            "100"
          ]
        },
        "300": {
          "value": "#c85c5a",
          "type": "color",
          "filePath": "tokens/color/base.js",
          "isSource": true,
          "original": {
            "value": "#c85c5a",
            "type": "color"
          },
          "name": "gcds-colour-base-red-300",
          "attributes": {},
          "path": [
            "colour",
            "base",
            "red",
            "300"
          ]
        },
        "500": {
          "value": "#bf3230",
          "type": "color",
          "filePath": "tokens/color/base.js",
[...]

PR #737, opened 2021-11-21
feat(json/original): Add json/original format

theKosh:
Issue #727

*Description of changes:*
- The json/original format allows you to leave only those keys that were specified in the original object without additional metadata. More info #727.
- Added dictionary filtered function:
  In:
   - obj - The object to filter. You will most likely pass `dictionary.tokens.color` to it.
   - exclude - The keys that need to be excluded from the filtered object
   - target - The keys that must be included in the result object.

  Out: the object which contains only given keys.

Tested with the Figma Tokens plugin. 


crypt096 (approved):

chazzmoney:
Hi @theKosh , thanks for your contribution!

This is a good idea, but needs some changes.

I'll add all of the comments below as a review, but I wanted to try to capture all the thinking in one spot:

- Adding a function to filter token metadata is a great idea that may have utility for a wide variety of SD users. Can we get another function that does the same exact thing but uses an "include" list instead of an "exclude" list?
- Adding a format for this is.... confusing. Formats are intended to be for platform style usage elsewhere. I'm unclear how this would be widely beneficial - or rather, how this would be useful to many/most StyleDictionary users
- There are a couple things that need to be cleaned, like changing the version number / package.lock

Thanks again for your contribution! Looking forward to adding in this functionality.

lib/common/formatHelpers/filterDictionary.js
```diff
@@ -0,0 +1,48 @@
...
+  if (target == null || target.length === 0) return obj;
+  if (obj == null || _.isEmpty(obj)) return {};
+  target = target.flat(1);
+  return Object.fromEntries(
+    Object
+      .entries(obj)
+      .map(([k, v]) => target.includes(k) ? [k, v] : typeof v === 'object' ? [k, filterDictionary(v, exclude, target)] : [k, {}])
+      .filter(([k, v]) => v == null ? true : false || _.isEmpty(v) || exclude.includes(k) ? false : true)
+  )
+}
```
chazzmoney, line 46: This is great. The name might be confusing though. Lets rename it something focused... maybe 'filterExcludeTokenMetadata'?

Also, can you add a 'filterIncludeTokenMetadata' function as well?

__tests__/common/formatHelpers/filterDictionary.test.js
```diff
@@ -0,0 +1,237 @@
+/*
```
chazzmoney, line 1: Nice work on the tests

chazzmoney (changes requested):
Nice work. A few changes to get it all in line for merge. Thanks!

theKosh:
Hi @chazzmoney.  

> Can we get another function that does the same exact thing but uses an "include" list instead of an "exclude" list?

Yes, okay, I'll do it.

> Adding a format for this is.... confusing. Formats are intended to be for platform style usage elsewhere. I'm unclear how this would be widely beneficial - or rather, how this would be useful to many/most StyleDictionary users

For example:
1. When you need to change tokens via transform, but you need to leave the json format as the original one, since you don't need extra metadata. In my case, for the web, I applied the rule `"transforms['color/css2']"`.
2. When tokens are generated for many platforms, but no changes are required for one of them, at the same time, it is more convenient for developers to have everything in one place in the build folder along with the processed assets. In this case, we can do it through the `fs.copyfilesync` function, but doing it through the format is more flexible, since it allows you to use transforms, manage assets, etc.

There is a discussion of the problem in [this topic](https://github.com/amzn/style-dictionary/issues/727).

> There are a couple things that need to be cleaned, like changing the version number / package.lock

I'll fix it.


Thanks for the detailed code review. I'll try to do everything in my spare time.

theKosh:
Hi, @chazzmoney.

I fixed all the errors and added the `excludeTokensMetadata` method as you asked.

Sorry it took so long — I got tied up at work.

PR #753, opened 2021-12-25
Update README.md

bobychaudhary:
Minor typo mistake
```diff
-stared
+started
```

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ float const SizeFontBase = 16.00f;
 
 ## Quick Start
 
-The style dictionary framework comes with some example code to get you stared. Install the node module globally, create a directory and `cd` into it.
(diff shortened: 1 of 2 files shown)


dbanksdesign (changes requested):
Thanks for making this change! The docs/formats.md file actually gets generated in a `generate-docs` script. To make sure this change doesn't get overridden you will need to make it in this file: https://github.com/amzn/style-dictionary/blob/main/scripts/handlebars/templates/formats.hbs and then run the `generate-docs` script.

crypt096 (changes requested):
Change as mentioned above.

chazzmoney:
Stale.

PR #759, opened 2022-01-12
Asynchronous parsers

jorenbroekema:
Hi, I think it would be valuable if parsers could be asynchronous.

Essentially parsers allow you to write tokens in any format, as long as you can write a parser for it that parses to the format that style-dictionary understands.

Multiple use cases have come up internally in our product where we actually need to compile or parse tokens but we can't get around the fact that this has to be done asynchronously, which is currently not possible in style-dictionary.

In our fork `browser-style-dictionary` we already introduced this change and it's working alright, but it's important to note that if style-dictionary were to adopt this change, it may need to do a breaking change to the main API `.extend()`. If parsers become async, probably so do `combineJSON` and `extend` by extension. I'm not sure if this is within your breaking change budget for 4.0.0 but it may go together well with other refactors such as ESM-only and modern ES syntax?

### DIFF
--- a/lib/utils/combineJSON.js
+++ b/lib/utils/combineJSON.js
@@ -37,9 +37,9 @@ function traverseObj(obj, fn) {
  * @param {Function} collision - A function to be called when a name collision happens that isn't a normal deep merge of objects
  * @param {Boolean} [source=true] - If json files are "sources", tag properties
  * @param {Object[]} [parsers=[]] - Custom file parsers
- * @returns {Object}
+ * @returns {Promise<Object>}
  */
-function combineJSON(arr, deep, collision, source, parsers=[]) {
+async function combineJSON(arr, deep, collision, source, parsers=[]) {
   var i, files = [],
     to_ret = {};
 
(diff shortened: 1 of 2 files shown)


lib/utils/combineJSON.js
```diff
@@ -61,14 +61,14 @@ function combineJSON(arr, deep, collision, source, parsers=[]) {
 
       // Iterate over custom parsers, if the file path matches the parser's
       // pattern regex, use it's parse function to generate the object
-      parsers.forEach(({pattern, parse}) => {
+      for (const {pattern, parse} of parsers) {
         if (resolvedPath.match(pattern)) {
-          file_content = parse({
+          file_content = await parse({
             contents: fs.readFileSync(resolvedPath, {encoding:'UTF-8'}),
```
tanvirstreame, line 67: As you are converting these in async await, I think you need to use `readFile()` instead of  `readFileSync`

jorenbroekema:
Done in v4 branch

PR #761, opened 2022-01-20
Add support for css variable fallback values when using outputReferences option

gillennium:
*Issue #, if available:* N/A

*Description of changes:* Add support for css variable fallback values when using outputReferences option.


chazzmoney:
Thanks for the contribution. My only concern is whether someone might not want this in their output. Should we make it a boolean configurable parameter? And, if so, should it be on or off by default?

gillennium:
Yeah, I think it makes sense to create a new configurable option parameter. I don't have a strong preference for whether it should be on or off by default. I'd think that most users would prefer it be on (if they're using the `outputReferences` option). The only advantage I could think off for turning it off is if you want the prop value to default back to `inherit` if the token is missing?

Would `outputFallbacks` be an intuitive param name to use? We'd just need to clarify in the documentation that this parameter is only pertinent when `outputReferences` is set to true.

ModulesUnraveled:
I think this is a great idea! I'd vote to have it on by default. And maybe `outputReferenceFallbacks` as the parameter name to help reinforce that it applies to the `outputReferences` parameter?

dbanksdesign:
I agree having it on by default is the way to go. @gillennium do you think you could add that configuration option? It would basically look like the `outputReferences` option here: https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L46 but you could default to `true`. If you don't have time I can make the change too.

dbanksdesign:
Looks like you might need to update some more snapshot tests too: https://github.com/amzn/style-dictionary/runs/4944101980?check_suite_focus=true#step:6:146

gillennium:
I'll make some updates to my PR today.

> I think this is a great idea! I'd vote to have it on by default. And maybe `outputReferenceFallbacks` as the parameter name to help reinforce that it applies to the `outputReferences` parameter?

I'll add a new option to control fallbacks, and agree with the default being true. I believe this would only apply to css variable references, correct? If so, should that be reflected in the option param name?

gillennium:
Ok, added `outputReferenceFallbacks` option to support css variable fallback values when using `outputReferences`; Fixed tests and snapshots.

dbanksdesign (approved):
LGTM!

kaelig:
I tried adding `outputReferenceFallbacks` to my sd.config.js file using the new release (3.7.0) and it didn't output the fallbacks.

Considering no snapshots were updated in this PR, is it possible that the functionality doesn't work?

sampotts:
The issue is that `outputReferenceFallbacks` doesn't get passed down to `createPropertyFormatter`. I'll open a PR but essentially:

`outputReferenceFallbacks` is not captured and passed to `formattedVariables` here:
https://github.com/amzn/style-dictionary/blob/6182d63765b86b19a83d27eeee227d5b0b9ea4ee/lib/common/formats.js#L46

Then `formattedVariables` needs to accept `outputReferenceFallbacks` in it's parameters, here:
https://github.com/amzn/style-dictionary/blob/6182d63765b86b19a83d27eeee227d5b0b9ea4ee/lib/common/formatHelpers/formattedVariables.js#L46
...and pass it to `createPropertyFormatter` here:
https://github.com/amzn/style-dictionary/blob/6182d63765b86b19a83d27eeee227d5b0b9ea4ee/lib/common/formatHelpers/formattedVariables.js#L64


If I modify my files in `node_modules` manually, I get the expected output.

sampotts:
Here's a patched version of the formatter for folks looking:

```js
import StyleDictionary from 'style-dictionary';
import { createPropertyFormatter, sortByReference, fileHeader } from 'style-dictionary/lib/common/formatHelpers';

type SortFn = (a: any, b: any) => number;

function formattedVariables({
  format,
  dictionary,
  outputReferences = false,
  outputReferenceFallbacks = false,
  formatting = {},
  themeable = false,
}) {
  let { allTokens } = dictionary;

  let { lineSeparator } = Object.assign(
    {},
    {
      lineSeparator: '\n',
    },
    formatting,
  );

  // Some languages are imperative, meaning a variable has to be defined
  // before it is used. If `outputReferences` is true, check if the token
  // has a reference, and if it does send it to the end of the array.
  // We also need to account for nested references, a -> b -> c. They
  // need to be defined in reverse order: c, b, a so that the reference always
  // comes after the definition
  if (outputReferences) {
    // note: using the spread operator here so we get a new array rather than
    // mutating the original
    allTokens = [...allTokens].sort(sortByReference(dictionary) as SortFn);
  }

  return allTokens
    .map(
      createPropertyFormatter({
        outputReferences,
        outputReferenceFallbacks,
        dictionary,
        format,
        formatting,
        themeable,
      }),
    )
    .filter(Boolean)
    .join(lineSeparator);
}

  name: 'css/variables-patched',
  formatter: ({ dictionary, options, file }) => {
    const selector = options.selector ? options.selector : `:root`;
    const { outputReferences, outputReferenceFallbacks } = options;
    return (
      fileHeader({ file }) +
      `${selector} {\n` +
      formattedVariables({ format: 'css', dictionary, outputReferences, outputReferenceFallbacks }) +
      `\n}\n`
    );
  },
});
```

PR #770, opened 2022-02-01
Feat(dictionary) Added new filter `removePrivate`

silversonicaxel:
*Description of changes:*
This PR is intended to provide a new builtin filter functionality, called `removePrivate`, to allow to set a design token to `private: true` in order to avoid to be outputted in the distribution folder after the build process.

For example, having an initial source of design tokens like the following:
```
{
  'black': {
    'value': '#000000',
    'private': true
  },
  'aqua': {
    'value': '#00FFFF'
  },
  'white': {
    'value': '#FFFFFF',
    'private': false
  }
}
```

a potential `scss` output would be something like this:
```
$aqua: '#00FFFF';
$white: '#FFFFFF';
``` 

or a potential `json` output would be something like this:
```
{
  'aqua': '#00FFFF',
  'white': '#FFFFFF'
}
``` 

This PR is created as follow up of the discussion hold in this other PR https://github.com/amzn/style-dictionary/pull/704 where the same feature was tried to be achieved with a different approach


silversonicaxel:
For visibility I tag @chazzmoney

silversonicaxel:
Since you were interested, I tag also @crypt096, @caseybaggz and @didoo

silversonicaxel:
I still need to improve the documentation. I'll try by myself tomorrow, but any tips are more than welcome.

silversonicaxel:
Ready for review

examples/advanced/filters/build.js
```diff
@@ -0,0 +1,32 @@
+const StyleDictionary = require('../../../index');
```
dbanksdesign, line 1: Can we remove this file completely since it has nothing really to do with filters?

examples/advanced/filters/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "style-dictionary-filters",
+  "version": "1.0.0",
+  "description": "",
+  "main": "build/index.js",
+  "files": [
+    "build",
+    "properties"
+  ],
+  "scripts": {
+    "build": "node ./build.js",
```
dbanksdesign, line 11: If we remove the `build.js` file we can change this to `"build": "style-dictionary build"`

examples/advanced/filters/build.js
dbanksdesign: Or if we want to keep this and show adding custom filters and make the file like this example config/build file: https://github.com/amzn/style-dictionary/blob/main/examples/advanced/custom-parser/sd.config.js

lib/common/filters.js
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * Remove a token from the ditribution output if it contains a key `private` set to true
+   *
+   * @memberof Filters
+   * 
+   * @param {Object} token
+   * @returns {Boolean} 
+   */
+  removePrivate: function(token) {
+    return (token && token.private) ? false : true
```
dbanksdesign, line 28: ah so clean ✨

scripts/version.js
```diff
@@ -1,6 +1,5 @@
 const fs = require('fs-extra');
 const glob = require('glob');
-const path = require('path');
```
dbanksdesign, line 3: Thanks!

examples/advanced/filters/README.md
```diff
@@ -0,0 +1,99 @@
+## Filters
```
dbanksdesign, line 1: Thank you for adding an example!

__tests__/common/filters.test.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+var filters = require('../../lib/common/filters');
+
+describe('common', () => {
+  describe('filters', () => {
```
dbanksdesign, line 17: Thank you for adding unit tests!

dbanksdesign (commented):
Thank you for all your work on this! Just one minor thing with the example and we are good to go!

chazzmoney (approved):
Thanks for doing this! It is really nicely done.

After @dbanksdesign changes are complete, I'm ready to merge this!

Nice work.

examples/advanced/filters/build.js
silversonicaxel: done the second option -> adjust custom filters

examples/advanced/filters/package.json
silversonicaxel: note necessary

silversonicaxel:
FYI https://github.com/design-tokens/community-group/issues/110#issuecomment-1029865540

chazzmoney:
@silversonicaxel I don't this the filter approach is in opposition; in fact, it is a simple change to make the filter run by default if the standards body decides that is a good idea and then publish it as a breaking change.

However, the final point in that discussion (that "private" may be an overloaded term with multiple meanings), is maybe a good one, which [I wrote a response to here](https://github.com/design-tokens/community-group/issues/110#issuecomment-1033123734).

Do we really want to use "private"? Or should we try and use a more explicit property name?

silversonicaxel:
@chazzmoney I see your point, line is subtle. there are some options on the table I think.

in favour (and with a personal order preference) (I also believe that them all might be also overloaded with extra meanings)
- `private` sounds the best choice, clear in its meaning
- `reserved` could be enough clear, understandable
- `secret` particular word, but also probably clear
- `hidden` semantically I feel it slightly more confusing than the rest (hiding from what?)
- `filtered` or `filteredOut` also possible


not in favour (since it tackles the other set of design tokens to keep in, and not filter out)
- `exported` semantically good
- `exposed` semantically good
- `visible` semantically is not super clear

What do you think? @dbanksdesign also, your opinion?

dbanksdesign (approved):
LGTM! :shipit: 

I think because this is a filter and not changing core functionality and the filter name clearly says 'removePrivate' I am good with this change even if the spec goes a different route.

PR #780, opened 2022-02-16
fix(types): Correct type of  `Core.format`

### LINKED ISSUE #779: Type definition of `Core.format` is not correct
As can be seen in [format.js](../blob/main/lib/common/formats.js) the exported dictionary is of type `Record<string, function>` (specifically `Record<string, Formatter>`), but in [index.d.ts](../blob/main/types/index.d.ts) it's defined as `Record<string, Format>`.

When accessing any entry (e.g. `StyleDictionary.format["css/variables"]`) the return type is then `Format` where it should be the `Formatter` function.

jakobe:
*Issue #, if available:* closes #779

*Description of changes:*
`Core.format` now has correct type (`Formatter` instead of `Format`).

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -20,7 +20,7 @@ import {Dictionary as _Dictionary} from './Dictionary';
 import {File as _File} from './File';
 import {FileHeader as _FileHeader} from './FileHeader';
 import {Filter as _Filter} from './Filter';
-import {Format as _Format} from './Format';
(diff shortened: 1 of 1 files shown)


dbanksdesign (approved):
LGTM! :shipit:

jakobe:
> LGTM! :shipit:

Great :)

So - who can :shipit: ?

Issue #794, opened 2022-03-13
Transitive transforms do not work well with references that don't have `.value` at the end


six7:
When you use references that do not contain .value at the end transitive transforms do not work as expected.

In the [transitive transforms example](https://github.com/amzn/style-dictionary/blob/main/examples/advanced/transitive-transforms/tokens/color/font.json5), when you swap out token references from `{color.font.primary.value}` to `{color.font.primary}` output changes so that it does not include the resolved previous reference.

**Expected output:**
```
  --color-font-primary: #000000;
  --color-font-secondary: #2c2c2c;
  --color-font-tertiary: #555555;
```

**Actual output:**
```
  --color-font-primary: #000000;
  --color-font-secondary: #2c2c2c;
  --color-font-tertiary: #2c2c2c;
```

Sebazzz:
Another example:

```json
    "fontFamilies": {
      "open-sans": {
        "value": "Open Sans",
        "type": "fontFamilies"
      }
    },
...
  "H1": {
      "fontFamily": {
        "value": "$fontFamilies.open-sans",
        "type": "fontFamily"
      },
...
```

Output:

```scss
$font-families-open-sans: Open Sans;
$h1-font-family: $fontFamilies.open-sans;
$h1-font-weight: $fontWeights.open-sans-0;
$h1-line-height: $lineHeights.0;
...

dbanksdesign:
Looks like this issue has been fixed for a bit. Closing for now, please re-open if the issue is still occurring.

PR #808, opened 2022-04-05
fix(transforms): transitive transforms now work without .value in refs

dbanksdesign:
*Issue #, if available:*

#794 

*Description of changes:*

Updated `resolveObject` function to do the ignorePaths check based on if the reference has a value or not. This allows transitive transforms to work on references without `.value`

### DIFF
--- a/__tests__/exportPlatform.test.js
+++ b/__tests__/exportPlatform.test.js
@@ -89,8 +89,14 @@ describe('exportPlatform', () => {
     const dictionary = StyleDictionary.extend({
       tokens: {
         one: { value: 'foo' },
-        two: { value: '{one}' },
-        three: { value: '{two}' }
+        two: { value: '{one.value}' },
+        three: { value: '{two}' },
+        four: { value: '{one}' },
+        five: { value: '{four.value}' },
+        six: { value: '{one}' },
+        seven: { value: '{six}'},
+        eight: { value: '{one.value}' },
+        nine: { value: '{eight.value}' }
       },
       transform: {
         transitive: {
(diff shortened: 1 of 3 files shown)


__tests__/exportPlatform.test.js
```diff
@@ -85,6 +85,32 @@ describe('exportPlatform', () => {
     expect(dictionary.color.button.hover.value).toEqual('#0077CC-darker-darker');
   });
 
+  it('should have transitive transforms applied without .value in references', () => {
+    const dictionary = StyleDictionary.extend({
+      tokens: {
+        one: { value: 'foo' },
+        two: { value: '{one}' },
+        three: { value: '{two}' }
```
chazzmoney, line 93: Can we add a couple tests that check different combinations of references with and without .value?

examples/advanced/transitive-transforms/tokens/color/font.json5
```diff
@@ -13,7 +13,7 @@
       },
       tertiary: {
         // transitive transforms allow you to modify a modified reference
-        value: "{color.font.secondary.value}",
+        value: "{color.font.secondary}",
```
chazzmoney, line 16: If we believe not having ".value" is the preferred method that we would like people to follow, this is great!

If not, it could be worthwhile to leave one of these with the .value appended so people know they can use it either way. Though, maybe that would be confusing and it is better to have a suggested method.

chazzmoney (changes requested):
Looks good - just a couple minor thoughts.

chazzmoney (approved):
LGTM! :shipit:

Issue #795, opened 2022-03-17
Crash when mixing refs and complex values in transitive transform


michaelurban:
I'm trying to do something that I haven't seen document, but, should(?) work. It almost works, at any rate.

Example project here: [box-shadow-example](https://github.com/michaelurban/box-shadow-example)

I'm specifying box-shadows as a list:
```
{
  "box-shadow": {
    "level1": {
      "value": [
        {
          "blurRadius": "2px",
          "color": "{color.shadow.key.value}",
          "offsetX": "0",
          "offsetY": "1px",
          "spreadradius": "0"
        },
        {
          "blurRadius": "3px",
          "color": "{color.shadow.ambient.value}",
          "offsetX": "0",
          "offsetY": "1px",
          "spreadradius": "1px"
        }
      ]
    }
}
```
And, I'm using a [transform](https://github.com/michaelurban/box-shadow-example/blob/main/src/transforms/box-shadow.transform.js) to convert that into a valid CSS `box-shadow` property value.

Expected Output:
```
:root {
  --bse-color-shadow-ambient: rgba(0, 0, 0, 0.15);
  --bse-color-shadow-key: rgba(0, 0, 0, 0.3);
  --bse-box-shadow-level2: 0 1px 2px var(--bse-color-shadow-key),
    0 2px 6px var(--bse-color-shadow-ambient);
  --bse-box-shadow-level1: 0 1px 2px var(--bse-color-shadow-key),
    0 1px 3px var(--bse-color-shadow-ambient);
  --bse-component-rect-height: 90px;
  --bse-component-rect-width: 160px;
  --bse-component-rect-box-shadow: var(--bse-box-shadow-level1);
}
```

[tokens/box-shadows.refs.json](https://github.com/michaelurban/box-shadow-example/blob/main/tokens/box-shadows.refs.json)

Actual Output:

```
if (a.original && dictionary.usesReference(a.original.value)) {
          ^

TypeError: Cannot read properties of undefined (reading 'original')
    at sorter (/<project path>/box-shadow-example/node_modules/style-dictionary/lib/common/formatHelpers/sortByReference.js:35:11)
    at sorter (/<project path>/box-shadow-example/node_modules/style-dictionary/lib/common/formatHelpers/sortByReference.js:57:16)
    at Array.sort (<anonymous>)
    at formattedVariables (/<project path>/box-shadow-example/node_modules/style-dictionary/lib/common/formatHelpers/formattedVariables.js:56:32)
[...]

PR #812, opened 2022-04-19
fix(references): getReferences now searches the entire object

dbanksdesign:
*Issue #, if available:* #795

*Description of changes:* 
* `getReferences` searches the entire object now. Previously `usesReference` searches the whole object so it would return `true`, but then `getReferences` would return an empty array if the token value was an array of objects
* Made `sortByReference` a bit more resilient to potential issues. 


VincentN:
This is great! We have an issue with nested references at the moment and this fix would be _very_ useful for us 🙌  I'm going to patch our style-dictionary node module with these changes through `patch-package` for the time being. 😁

chazzmoney (approved):
LGTM!

PR #829, opened 2022-06-14
fix: added packageName as optional property on interface File

aborgan:
*Issue #, if available: N/A*

*Description of changes:*

The `packageName` property is not present on the `File` interface, but is listed in [the docs](https://amzn.github.io/style-dictionary/#/formats?id=composeobject) and is used in testing [here](https://github.com/amzn/style-dictionary/blob/main/__integration__/compose.test.js#L30).

### DIFF
--- a/types/File.d.ts
+++ b/types/File.d.ts
@@ -16,6 +16,7 @@ import { TransformedToken } from './TransformedToken';
 
 export interface File {
   className?: string;
+  packageName?: string;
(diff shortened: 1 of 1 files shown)


dbanksdesign (approved):
LGTM!

alexiscordova:
@dbanksdesign can we get this merged? I think only repo admins have write access. Thanks!

PR #869, opened 2022-09-08
fix(docs): Correct custom-file-header example link

mattleff:
Fixes the link to the `custom-file-header` example


dbanksdesign (approved):
LGTM!

PR #874, opened 2022-09-19
Compose object imports

giovannimartusciello:
*Issue #, if available:*  Selectively import on compose class was not available

*Description of changes:*
Added a formatHelper for import option + added map function in the compose template


cgal-capco:
This is amazing! It's exactly what I miss to convert to Android Compose objects

lib/common/templates/compose/object.kt.template
```diff
@@ -17,9 +17,10 @@
...
 
-<%  // Ideally we would selectively import the correct classes & functions based on whether the applied transforms actually need them. %>
-import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.unit.*
+<%= options.import.map(function(item) {
+    return 'import ' + item
+  }).join('\n')
+%>
 
 object <%= file.className %> {
```
danieldelcore, line 25: Unrelated to this pr, but shouldn't `file.className` be `options.className`?

giovannimartusciello: I don't think so, because in the config .json the className is directly on the object file and not under options

lib/common/formats.js
```diff
@@ -740,6 +741,8 @@ declare const ${moduleName}: ${JSON.stringify(treeWalker(dictionary.tokens), nul
       allProperties = [...dictionary.allProperties].sort(sortByName);
     }
 
+    options = setComposeObjectProperties(options);
```
bherbst, line 744: Is it possible to choose the imports based on which transforms are being used instead of accepting arbitrary inputs in the configuration?

My concern is that by putting the inputs in the configuration we make it relatively easy to accidentally generate a Kotlin file that won't compile, whereas if we do the detection of which imports we actually need we can guarantee that the Kotlin file will compile correctly.

giovannimartusciello: I actually followed the same line as was already done with swift classes, sometimes custom import are needed

martymiller:
Really need this.  Is there a reason it was never merged?

cjgaliana:
+1 @martymiller 

Ideally the imports should be automatically imported from the actual used classes and functions based on whether the applied transforms actually need them.

However, at the moment the current code hardcodes 2 imports 
```
import androidx.compose.ui.graphics.Color
import androidx.compose.ui.unit.*
```

I think this change, although it won't reach the desired goal, it will improve the current code allowing to import more than the 2 hardcoded imports.

giovannimartusciello:
@dbanksdesign can we request an approve for this? in order to integrate this feature?

giovannimartusciello:
> Really need this. Is there a reason it was never merged?

Not really, we are waiting for an approve

dbanksdesign (approved):
LGTM!

Apologies for the lateness of this review, been very busy, but trying to dig through PRs and issues now. Thank you for understanding.

PR #878, opened 2022-09-27
Add matcher type to export

### LINKED ISSUE #875: [Typescript] Export Matcher type
The [Matcher type](https://github.com/amzn/style-dictionary/blob/main/types/Matcher.d.ts) is missing from the [SD export](https://github.com/amzn/style-dictionary/blob/main/types/index.d.ts#L36-L55).

However when defining matchers as functions and adding them in an object to the SD config without the register function, the type would be helpful.

I am happy to send a PR if that makes sense @chazzmoney?

lukasoppermann:
This fixes #875

Exports matcher type.


dbanksdesign (approved):
LGTM!

PR #883, opened 2022-10-20
fix(types): fix filter config key expected matcher value

pascalduez:
*Description of changes:*

The `filter` config key expect an object of type `{ name: string, matcher: Matcher }` and not `{ name: string, matcher: Filter }`.   
Which is different from `registerFilter` that accept a `Filter` object.


dbanksdesign (approved):
Great catch! Thanks for fixing this!

Issue #914, opened 2023-01-10
Remove examples folder from published package


brettimus:
Currently, when you `npm i style-dictionary`, you end up with all the examples in the `examples` folder. This is roughly 19mb, which is quite significant!

```
% mkdir style-dictionary-examples-test && cd style-dictionary-examples-test
% npm i style-dictionary
% du -h node_modules/style-dictionary/examples
...
 19M	node_modules/style-dictionary/examples
```

Adding `examples` to the `.npmignore` would solve this issue.

dbanksdesign:
Thanks for the feedback! We currently use the examples folder if you use `style-dictionary init [simple|complete]`, but that was before "create" packages became a thing so it would be good to move to a "create-style-dictionary" package and remove the "examples" from the npm artifacts.

adidde1073:
Is there any way we can revisit this? Currently, we're having a security issue because some of the dependencies ([FasterXML/jackson-core](https://github.com/FasterXML/jackson-core)) need to be updated

donaldpipowitch:
Another use case: tools like https://dependencytrack.org/ might track irrelevant dependency manifests that way.

<img width="994" height="244" alt="Image" src="https://github.com/user-attachments/assets/1bf8fa02-b709-4327-a93f-af22d47ea434" />

jorenbroekema:
This is planned to be revisited for v6 milestone. Examples, if we choose to publish/maintain them at all, will probably be in its own package e.g. `@style-dictionary/example-foo`

PR #915, opened 2023-01-10
Ignore docs and examples from the published npm package

brettimus:
*Issue #, if available:* Closes https://github.com/amzn/style-dictionary/issues/914

*Description of changes:*

This PR changes the `.npmignore` to exclude files related to examples and documentation from the published NPM package. 

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,3 +3,6 @@
 .editorconfig
 images/
 test/
+examples/
+docs/
+.jsdoc.json


tripu (approved):
Static analysis tools looking for vulnerabilities are finding a file `style-dictionary/examples/advanced/custom-file-header/node_modules/json5/package.json` and screaming [CVE-2022-46175](https://nvd.nist.gov/vuln/detail/CVE-2022-46175).

Yet another reason to exclude examples from the npm tarball.

So, +10 from my and my colleagues.

coliff (approved):

dbanksdesign (approved):
LGTM! Thank you for this!

chazzmoney:
Unfortunately, the SD executable expects the examples folder to be present. However, you could remove / hide the advanced folder. Or, you could suggest a PR to remove the init functionality.

dbanksdesign:
Ah good catch @chazzmoney! You are right, this would break `style-dictionary init` .

chazzmoney:
For clarity, this is the code that would be broken:

https://github.com/amzn/style-dictionary/blob/main/bin/style-dictionary#L60-L67

so if you retain examples/basic and examples/complete and npmignore examples/advanced, it would remove that vulnerability.

Or, maybe we could be convinced to remove the init...

.npmignore
```diff
@@ -3,3 +3,6 @@
 .editorconfig
 images/
 test/
+examples/
```
chazzmoney, line 6: See the comments on the PR. Can't exclude this folder directly because it breaks functionality. Other options would work though.

chazzmoney (changes requested):

jorenbroekema:
Revamping docs, examples in the near future, will try to ensure if we still put them on NPM, they at least don't have security issues, but will close this particular PR as it would break init command

brettimus:
Sorry I never followed up on this PR! But yeah, I think for teams who need to be meticulous about every dependency and sub-dependency they use, having minimal deps for the examples would be much appreciated. Will be happy to see #914 closed 🙏

PR #926, opened 2023-02-01
fix(types): loosening the Platform type to allow for extension

dbanksdesign:
*Issue #, if available:*

*Description of changes:*

The Platform type can really be any object so users can have custom defined attributes on it which they can use in their custom transforms. This loosens up the Platform type a bit and uses that type with generic in the transformer function. 


chazzmoney (approved):
LGTM! :shipit:

PR #931, opened 2023-02-06
Add missing className file format documentation for ios-swift classes

gsouquet:
*Issue #, if available:* N/A

*Description of changes:*

Adding the missing file property `className`. 
This template (see https://github.com/amzn/style-dictionary/blob/main/lib/common/templates/ios-swift/any.swift.template) confirms that this property can be used, but is undocumented, 

### DIFF
--- a/docs/formats.md
+++ b/docs/formats.md
@@ -608,6 +608,27 @@ StyleDictionary.registerFormat({
 
 * * *
 
+### setComposeObjectProperties 
+> formatHelpers.setComposeObjectProperties(options) ⇒ <code>Object</code>
+
+Outputs an object for compose format configurations. Sets import.
+
+<table>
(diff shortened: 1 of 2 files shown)


dbanksdesign (changes requested):
Thank you for catching this! That props table is generated from a JSDoc comment, here: https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L915

And then run `npm run generate-docs` 

Just to make sure this change doesn't get wiped out.

gsouquet:
Thank you for the review @dbanksdesign 
I've made those changes. It looks like `npm run generate-docs` was lagging a bit behind, so there's a few unrelated changes in the diff.

dbanksdesign (approved):

Issue #952, opened 2023-03-28
Comments with newlines are not formatted nicely in createPropertyFormatter


jorenbroekema:
This was raised in our repo https://github.com/tokens-studio/sd-transforms/issues/62 but I believe it to be best fixed in style-dictionary `createPropertyFormatter`.

## Current

```css
:root {
  --foo-bar: #FFFFFF; // Some comment with a 
<-- new line here that creates invalid CSS
}
```

^ this part I would call a bug, because it actually results in broken CSS, so in that sense this issue is not just a feature request but moreso a bug report.

or if you set it to commentStyle `"long"`:

```css
:root {
  --foo-bar: #FFFFFF; /* Some comment with a 
<-- new line here that creates invalid CSS */
}
```

## Expected

Newlines to be filtered out in short:

```css
:root {
  --foo-bar: #FFFFFF; // Some comment with a <-- new line here that creates invalid CSS
}
```

or with commentStyle `"long"`:

```css
:root {
  /** 
   * Some comment with a 
   * <-- new line here that creates invalid CSS
   */
  --foo-bar: #FFFFFF;
}
```

I'm happy to create a PR for this, but only if you guys have the time to review & release it at some point :)

chazzmoney:
@jorenbroekema We will have time to review and release - we are checking once a week or so right now. If you have a PR submitted, we will take a look.

PR #953, opened 2023-04-05
fix: account for multi-line comments in tokens, improve comment style

jorenbroekema:
*Issue #, if available:*

fixes https://github.com/amzn/style-dictionary/issues/952

*Description of changes:*

- Account for multi-line comments, by splitting them up in the final comment string.
- Put multi-line comments above the token, making it more readable.


jorenbroekema:
I added a second commit with a small change that allows formatting prop to be user-passed, even if the format is css/sass/less/stylus, as well as a small change to allow passing commentStyle "short-above" or "long-above" to ensure the comment is put 1 line above the outputted property, rather then inline next to it.

Let me know if those two things are okay or you prefer it some other way or via a separate PR

__tests__/common/formatHelpers/createPropertyFormatter.test.js
```diff
@@ -0,0 +1,106 @@
...
+
+      // red = single-line comment, blue = multi-line comment
+      const cssRed = cssFormatter(dictionary.tokens.color.red);
+      const cssBlue = cssFormatter(dictionary.tokens.color.blue);
+      const sassRed = sassFormatter(dictionary.tokens.color.red);
+      const sassBlue = sassFormatter(dictionary.tokens.color.blue);
+
+      // Note that since CSS puts it inside a selector, there is an indentation of 2 spaces as well
+      // CSS also has commentStyle long, whereas sass uses short
+      expect(cssRed).toEqual('  --color-red: #FF0000; /* Foo bar qux */');
```
dbanksdesign, line 78: Could we make these `toMatchInlineShanpshot` just so if things do change intentionally it is easy to update.

lib/common/formatHelpers/createPropertyFormatter.js
```diff
@@ -19,6 +19,59 @@ const defaultFormatting = {
   suffix: ';'
 }
 
+/**
+ * Split a string comment by newlines and
+ * convert to multi-line comment if necessary
+ * @param {string} to_ret_prop
+ * @param {{comment: string; style: 'short' | 'long'; indentation: string}} options
+ * @param {'short' | 'long'} commentStyle
```
dbanksdesign, line 27: I think this param is unnecessary I think?

lib/common/formatHelpers/createPropertyFormatter.js
```diff
@@ -19,6 +19,59 @@ const defaultFormatting = {
...
+ * @param {string} indentation
+ * @returns {{hasNewLines: boolean, comment: string}}
+ */
+function addComment(to_ret_prop, options) {
+  const { comment, style, indentation } = options;
+
+  const [commentStyle,placeAbove] = style.split('-above');
```
dbanksdesign, line 34: Should we just have a separate option for `commentPosition`? Feels a bit weird to split a string for configuration

__tests__/common/formatHelpers/createPropertyFormatter.test.js
```diff
@@ -0,0 +1,106 @@
...
+        "color",
+        "green",
+      ]
+    }
+  }
+};
+const dictionary = createDictionary({ properties });
+
+describe('common', () => {
+  describe('createPropertyFormatter', () => {
```
dbanksdesign, line 63: Thank you for adding unit tests!

dbanksdesign (commented):
Overall looks good, just a few minor things! Thank you!

lib/common/formatHelpers/createPropertyFormatter.js
jorenbroekema: Indentation param also incorrect, and return type was incorrect too, fixed them

lib/common/formatHelpers/createPropertyFormatter.js
jorenbroekema: Agree, was a bit hacky, I improved the code to use commentPosition, also improved the switch statement

__tests__/common/formatHelpers/createPropertyFormatter.test.js
jorenbroekema: Done!

dbanksdesign (approved):

chris-dura:
> I added a second commit with a small change that allows formatting prop to be user-passed, even if the format is css/sass/less/stylus, as well as a small change to allow passing commentStyle "short-above" or "long-above" to ensure the comment is put 1 line above the outputted property, rather then inline next to it.
> 
> Let me know if those two things are okay or you prefer it some other way or via a separate PR

Should this have documentation updated as well around the additional options for `commentStyle`?

https://amzn.github.io/style-dictionary/#/formats?id=fileheader

---

Related, are these types incorrect? The types listed here say the options for `commentStyle` are `"short" | "long" | "none"`...

https://github.com/amzn/style-dictionary/blob/0272b8f946bcaef50452af4dc7393b1f559bfdf3/types/FormatHelpers.d.ts#L21

PR #961, opened 2023-04-12
Ability to now run Tests in Parallel

gillianyost:
Now no longer need --runInBand for tests

While using this repo for a flaky test detection project, I have found that when your tests are run in parallel they have test dependencies which creates flaky tests. I have modified the test dependencies between test suites where the regular clearing of the __tests__/__output directory means the test suites cannot be run in parallel. Running in parallel is more efficient and thus in the package.json file the "--runInBand" command has been removed from the test command. There are now separate config files for the tests and the helper function that clears the output now takes a parameter to select which directory it clears. 


jorenbroekema:
In v4 branch we use web test runner which runs each test file in a separate browser tab, number depends on concurrency.

Node tests were migrated to mocha, not currently in parallel due to some issues with Mocha's ESM support, but either way I think we can close this PR since we moved away from Jest, I explained some of that here https://github.com/amzn/style-dictionary/pull/1113#issuecomment-1988280585

PR #972, opened 2023-05-12
Add static option to any.swift.template

germain-gg:
*Issue #, if available:*

N/A

*Description of changes:*

Sometimes it is desirable to reference design tokens via a [KeyPath](https://developer.apple.com/documentation/swift/keypath) when using them from Swift. To support this, the generated token set would need to be instance properties and not static. This PR adds a `static` options for Swift classes, it defaults to `true` so all current deployment will continue working as is.


jorenbroekema:
Hey, this seems like a good addition, if still needed, could you rebase it onto the v4 branch and run npx changeset to add a patch changelog entry for it?

PR #979, opened 2023-05-24
Fix grammar

ic-768:
Fixed minor grammar issues in transforms section of documentation

### DIFF
--- a/docs/transforms.md
+++ b/docs/transforms.md
@@ -33,7 +33,7 @@ There are 3 types of transforms: attribute, name, and value.
 
 **Attribute:** An attribute transform adds to the attributes object on a design token. This is for including any meta-data about a design token such as it's CTI or other information.
 
-**Name:** A name transform transform the name of a design token. You should really only be apply one name transformer because they will override each other if you use more than one.
+**Name:** A name transform transforms the name of a design token. You should really only be applying one name transformer because they will override each other if you use more than one.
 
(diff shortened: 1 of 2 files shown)


alanbsmith (approved):
👍🏼

dbanksdesign (approved):

docs/transforms.md
```diff
@@ -33,7 +33,7 @@ There are 3 types of transforms: attribute, name, and value.
 
 **Attribute:** An attribute transform adds to the attributes object on a design token. This is for including any meta-data about a design token such as it's CTI or other information.
 
-**Name:** A name transform transform the name of a design token. You should really only be apply one name transformer because they will override each other if you use more than one.
+**Name:** A name transform transforms the name of a design token. You should really only be applying one name transformer because they will override each other if you use more than one.
```
dbanksdesign, line 36: Can you update this file: https://github.com/amzn/style-dictionary/blob/main/scripts/handlebars/templates/transforms.hbs  the docs/transforms.md file is generated via a script based on that handlebars file

dbanksdesign (changes requested):
Need to update another file so this change doesn't get lost when we release

docs/transforms.md
ic-768: Done!

jorenbroekema:
@dbanksdesign  can you change your review to approve?

I also went ahead and forward ported this to v4 already :)

dbanksdesign (approved):

PR #1000, opened 2023-07-21
fix(formats): added nested property to json/asset

jduthon:
*Issue #, if available:*

*Description of changes:*

We are getting a lot of warnings during tokens' generation mentioning name collisions.

I started looking into this and discovered the `nested` format idea, which works great for the formats where it's set up (and we had to add it to a few of our formats), but then I realized I was still getting errors during asset files generation with the `json/format` format.
Looking at the code + at its output, it seems like this format is a nested format too, but I am guessing it was missed during the time where nested formats where introduced.

So this PR is just making the `json/format` officially a `nested` format, so there won't be anymore wrong warnings for it.

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -1263,3 +1263,4 @@ declare const ${moduleName}: ${JSON.stringify(treeWalker(dictionary.tokens), nul
 module.exports['json/nested'].nested = true;
 module.exports['javascript/module'].nested = true;
 module.exports['javascript/object'].nested = true;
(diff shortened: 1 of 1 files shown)


jorenbroekema:
Hey, sorry for the late response, I think this PR is still valid and hasn't been fixed yet in main branch, could you rebase your PR on the latest branch? https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L1540

And when you're done, running `npx changeset` will help you writing a patch note for the change (a patch is fine for this)

Issue #974, opened 2023-05-15
Conflicts with outputReferences when custom transform changes token value


germain-gg:
I am using [tokens.studio](https://tokens.studio/) and their set of style-dictionary transforms https://github.com/tokens-studio/sd-transforms

Specifically a transform called `ts/resolveMath`. It appears that the transform is applied before the formatting phase. And one part of the system gets confused and the output is invalid CSS. See a more in-depth [investigation by the package maintainer](https://github.com/tokens-studio/sd-transforms/issues/13#issuecomment-1486981956)

## Reproduction steps

```jsonc
{
  "space": {
    "scale": {
      "value": "4",
      "type": "spacing",
      "description": "Base unit for spacing scale."
    },
    "6x": {
      "value": "{space.scale} * 6",
      "type": "spacing"
    },
    "7x": {
      "value": "{space.scale} * 7",
      "type": "spacing"
    }
  }
}
```

And a style-dictionary config of

```js
{
  "transformGroup": "css-tokens",
  "transforms": [
    "ts/resolveMath",
    "ts/size/px",
    "name/cti/kebab"
  ],
  "files": {
    "destination": "style.css",
    "format": "css/variables",
    "options": {
      "outputReferences": true
    }
  }
}
```

### Expected output

```css
:root {
  --space-scale: 4px;
  --space-7x: 28px;
  --space-6x: 24px;
}
```

### Actual output

```css
:root {
  --space-scale: 4px;
  --space-7x: 28px;
  --space-6x: 2var(--space-scale);
}
```

chazzmoney:
Thanks for your note.

After looking into it, there is indeed a conflict between ts/resolveMath and "outputReferences": true. Using both of these will cause problems because they both work functionally on strings.

For the output you wish to see, I would suggest removing the "outputReferences": true. This should create the output that you want. If, however, you wanted something more like:

```
:root {
  --space-scale: 4px;
  --space-7x: var(--space-scale) * 7;
  --space-6x: var(--space-scale) * 6;
}
```

Then you can leave in "outputReferences": true and remove the ts/resolveMath transform.

In general these two systems both intend to process the strings to create different outputs, so should not be used together.

chazzmoney:
@jorenbroekema if you have additional thoughts, happy to hear them

jorenbroekema:
Another related issue here https://github.com/tokens-studio/sd-transforms/issues/121 where I share my thoughts.

[...]

PR #1002, opened 2023-07-28
fix: outputReferences to use original.value when possible

jorenbroekema:
*Issue #, if available:*
https://github.com/amzn/style-dictionary/issues/974

*Description of changes:*
As suggested in the issue, change the replacing of values with references to use original.value instead of ref.value (which could have gotten transformed in the meantime), whenever possible, as this will create a more accurate output when the ref.value has been changed to a point where using that as a replacer reference is no longer possible.

### DIFF
--- a/lib/common/formatHelpers/createPropertyFormatter.js
+++ b/lib/common/formatHelpers/createPropertyFormatter.js
@@ -106,13 +106,22 @@ function createPropertyFormatter({
       // or else you will get '[object Object]' in the output
       if (typeof value === 'string') {
         const refs = dictionary.getReferences(prop.original.value);
+
+        // original can either be string value or an object value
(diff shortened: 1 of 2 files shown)


lib/common/formatHelpers/createPropertyFormatter.js
```diff
@@ -122,7 +131,11 @@ function createPropertyFormatter({
               } else {
                 return `${prefix}${ref.name}`;
               }
-            });
+            }
+            // when original is object value, we replace value by matching ref.value and putting a var instead
```
dbanksdesign, line 135: Thank you for this comment!

dbanksdesign (approved):
LGTM! Thank you!

PR #1014, opened 2023-09-20
Make ESM-only and browser-compatible out of the box

jorenbroekema:
*Description of changes:*
Converts the entire repository to:
- ESM-only (modern JS format)
- Browser-compatible out of the box, by shimming path, glob, fs, modules

Apart from the ESM format changes, the changes you will encounter:

- Filesystem can be customized, see: https://github.com/amzn/style-dictionary/pull/1014/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R43-R51
- Any reliances on "require()" to import JSON files are now using `JSON.parse(fs.readFileSync())` instead. For JS files, only ESM is supported, through dynamic imports. Due to dynamic imports being async in nature, some of the APIs have already changed to becoming async and returning promises.
- Some deprecated stuff, I already removed in this PR, because they tended to be features that were a bit annoying to convert to browser-compatible ESM.
- Some formatting changes still, sadly, didn't catch everything in the previous PR
- Changing `var` to `const` or `let`
- Changing the templates files to be ES Modules, so they can be more easily imported in both NodeJS and the Browser too.

TODOs for next PR, since at Tokens Studio we could really use an early, unstable, prerelease of this change, and the PR is big enough as is:
- Lots of manual smoke testing in both browser and node environment to find issues early
- Make the tests work again, to spot regressions, but we should test in a headless browser to spot browser incompatibilities, so we'll have to switch test runner
- Remove leftover references to properties/allProperties, now is tokens/allTokens

### Breaking changes list

- CommonJS format Configs or Tokens are no longer supported
- StyleDictionary.extend() is now async, so returns `Promise<StyleDictionary.Core>` instead of `StyleDictionary.Core`
- allProperties/properties was deprecated, is now removed from StyleDictionary.Core in some places (although lots of leftovers still remain in the codebase, to be removed).
- Templates/registerTemplate were deprecated in favor of formats, now templates is removed.


index.js
```diff
@@ -10,12 +10,45 @@
...
+// Allow to be overridden by setter, set default to memfs for browser env, node:fs for node env
+export let fs =
+  typeof window === 'object'
+    ? (await import('@bundled-es-modules/memfs')).default
+    : (await import('node:fs')).default;
+
+// since ES modules exports are read-only, use a setter
+export const setFs = (_fs) => {
+  fs = _fs;
+};
```
jorenbroekema, line 51: important change, to allow customizing the filesystem that's used.

- If browser -> use memfs by default
- If node -> use node:fs by default
- Allow setting it to a different FS shim/equivalent with setFs()

lib/buildFile.js
```diff
@@ -11,13 +11,13 @@
...
  */
 
-var path = require('path'),
-  fs = require('fs-extra'),
-  chalk = require('chalk'),
-  filterProperties = require('./filterProperties'),
-  GroupMessages = require('./utils/groupMessages');
+import path from '@bundled-es-modules/path-browserify';
+import chalk from 'chalk';
+import { fs } from '../index.js';
```
dbanksdesign, line 16: Does this create a circular dependency because index.js imports this file too?

dbanksdesign (approved):
LGTM! Just one minor comment/question about importing `fs`

lib/buildFile.js
jorenbroekema: Yes, there's a few actually in the library that I spotted in the configurator project when bundling stuff for prod:

```
node_modules/style-dictionary/index.js -> node_modules/style-dictionary/lib/extend.js -> node_modules/style-dictionary/index.js
node_modules/style-dictionary/index.js -> node_modules/style-dictionary/lib/extend.js -> node_modules/style-dictionary/lib/utils/combineJSON.js -> node_modules/style-dictionary/index.js
node_modules/style-dictionary/index.js -> node_modules/style-dictionary/lib/common/transforms.js -> node_modules/style-dictionary/lib/utils/convertToBase64.js -> node_modules/style-dictionary/index.js
...and 6 more
```

Maybe this PR is a good time to deal with them all :) even though most bundlers can deal with circular dependencies, the order of imported modules might matter, so circular deps are still an anti-pattern that should preferably be dealt with

jorenbroekema: Fixed the circular deps in the most recent commit :) tested with Rollup bundle -> no warnings whatsoever now.

PR #1020, opened 2023-10-02
fix: escape double quotes for ts outputStringLiterals

andnorda:
Escaping double quotes when generating string literal types is useful when a token value is referencing a font family name.

### DIFF
--- a/__tests__/common/formatHelpers/getTypeScriptType.test.js
+++ b/__tests__/common/formatHelpers/getTypeScriptType.test.js
@@ -47,9 +47,10 @@ describe('common', () => {
       });
 
       it('should handle outputStringLiterals', () => {
-        const stringValue = 'I am a string';
+        const stringValue = 'I "am" a string';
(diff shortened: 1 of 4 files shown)


jorenbroekema:
Hi, would you be down to create this change on the v4 branch? the v3/main branch is so far behind at this point that I'm only backporting really important bug or security fixes.

Could you also explain a bit more in detail why there's a triple backslash instead of a double? Double is what I'd expect, one to escape the escape character and one for the escape character itself, but the third one isn't immediately obvious to me why it's required or why that's the output of the test.

andnorda:
Opened a new PR to v4 here: https://github.com/amzn/style-dictionary/pull/1097

Not sure what the triple backslash was about. It seems to work with double backslashes in the other PR.

andnorda:
As this rule says, some backslashes have no effect: https://eslint.org/docs/latest/rules/no-useless-escape

I think that was the case here, so functionally three was the same as two.

PR #1043, opened 2023-11-02
feat: add preprocessors

jorenbroekema:
Sometimes it makes sense to process the dictionary after it has been created from merging the parsed input files together into one JS object.
For example, when more complicated and broad transformations have to be done in order for things like references to work, metadata to be deleted, etc.

In the past, I've used custom parsers to parse design token files from Tokens Studio, but we ran into issues with cross-file references, which at the parser lifecycle cannot be resolved and transformed, because the parsers work on single input files one by one. I think that totally makes sense for parsers, I wouldn't change that, but instead I think it makes sense to introduce this preprocessor lifecycle hook to deal with use cases like this.

TODO:
- [x] tests
- [x] example (one under advanced to delete metadata or something, should do)


types/index.d.ts
```diff
@@ -208,6 +211,25 @@ declare namespace StyleDictionary {
...
+     * @example
+     * ```js
+     * StyleDictionary.registerParser({
+     *   pattern: /\.json$/,
+     *   parse: ({contents, filePath}) => {
+     *     return JSON.parse(contents);
+     *   }
+     * })
+     * ```
+     */
```
dbanksdesign, line 230: Update this JSDOC comment please :)

dbanksdesign (commented):
Just one small thing, but overall looks good!

types/index.d.ts
jorenbroekema: Done. I also rebased this branch to `fix-tests` because it had quite some changes and refactors. I'll also add some tests now.

jorenbroekema:
@dbanksdesign this needs another review, I've rebased to the branch that has all tests fixed (and also contains some refactors and API changes that affect this PR). I also changed the API of preprocessors to be aligned with formats, transforms, fileheader etc. which just means you register it with a name property, I think this is better API design, helps with indexing and debugging preprocessors if multiple are used.

dbanksdesign (approved):

PR #1047, opened 2023-11-23
fix: types for FormatHelpers.LineFormatting to add commentPosition

jorenbroekema:
*Issue #, if available:*
https://github.com/amzn/style-dictionary/pull/953#issuecomment-1823532951

*Description of changes:*
Adds commentPosition prop to FormatHelpers.LineFormatting type

We should release 3.9.1 as well, for a fix in outputReferences (falsy values not replaced atm)

### DIFF
--- a/types/FormatHelpers.d.ts
+++ b/types/FormatHelpers.d.ts
@@ -19,6 +19,7 @@ import { File } from "./File";
 export interface LineFormatting {
   prefix?: string;
   commentStyle?: "short" | "long" | "none";
+  commentPosition?: "inline" | "above";
   indentation?: string;
   separator?: string;
   suffix?: string;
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "style-dictionary",
-  "version": "3.9.0",
+  "version": "3.9.1",
```
dbanksdesign, line 3: Can we remove this change? `standard-version` takes care of updating the version number

dbanksdesign (commented):
Just one small change

package.json
jorenbroekema: Oh true, forgot about that

dbanksdesign (approved):

Issue #1040, opened 2023-10-30
Drop _es6 helper utility


jorenbroekema:
https://github.com/amzn/style-dictionary/blob/v4/lib/utils/es6_.js

Most of these utils can be dropped, e.g. modern JS supports it.

change-case is still useful. is-plain-object npm dependency might be useful as well.

@tahul perhaps?

PR #1048, opened 2023-11-23
chore: remove es6 helpers and use actual ES6

jorenbroekema:
*Issue #, if available:*
fixes https://github.com/amzn/style-dictionary/issues/1040

*Description of changes:*
Remove es6 helpers, no longer needed with how far JS / ES6+ has gotten


lib/utils/resolveObject.js
```diff
@@ -31,7 +30,7 @@ let updated_object, regex, options;
 export default function resolveObject(object, opts) {
   options = Object.assign({}, defaults, opts);
 
-  updated_object = cloneDeep(object); // This object will be edited
+  updated_object = structuredClone(object); // This object will be edited
```
lb-, line 33: Just a heads up `structuredClone` may only be supported for Node 17 & up.
https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

I think this project still expects to be run on Node 12. Maybe the `engines.node` in package.json will need to be updated to 18 or 20.

dbanksdesign (approved):

PR #1056, opened 2023-12-04
feat: refactor reference utils and expose as standalone functions

jorenbroekema:
*Description of changes:*

Makes reference utility functions not bound to the dictionary instance.
This makes it easier to expose them as publicly usable utilities, makes them more straightforward to use by users (no need to do `const boundGetReferences = getReferences.bind({ properties: ... });` in order to use)

### DIFF
--- a/__tests__/utils/reference/getValueByPath.test.js
+++ b/__tests__/utils/reference/getValueByPath.test.js
@@ -11,7 +11,7 @@
  * and limitations under the License.
  */
 import { expect } from 'chai';
-import resolveReference from '../../../lib/utils/references/resolveReference.js';
+import getValueByPath from '../../../lib/utils/references/getValueByPath.js';
 
 const dictionary = {
   color: {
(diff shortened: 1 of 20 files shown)


__tests__/utils/reference/resolveReferences.test.js
```diff
@@ -0,0 +1,223 @@
+/*
```
jorenbroekema, line 1: Mostly copied from the resolveObject tests but I refactored every single test to make it a unit-test for the resolveReference function, which is now more granularly tested

__tests__/utils/reference/getValueByPath.test.js
```diff
@@ -11,7 +11,7 @@
  * and limitations under the License.
  */
 import { expect } from 'chai';
-import resolveReference from '../../../lib/utils/references/resolveReference.js';
+import getValueByPath from '../../../lib/utils/references/getValueByPath.js';
```
jorenbroekema, line 14: This needed a rename imo, bit confusing, it doesn't actually resolve the reference, just gives you the value of the token using the token "path"

lib/common/templates/android/resources.template.js
```diff
@@ -1,50 +1,61 @@
-export default `<?xml version="1.0" encoding="UTF-8"?>
```
jorenbroekema, line 1: Since this file now imports from `style-dictionary/utils` I was getting errors about "import not allowed outside module" error or whatever. I couldn't find the culprit, I imagine it has something to do with lodash "template" util not allowing ESM imports or something, so I just went ahead and refactored this thing to use native ES6 Template Literals instead, as I'm planning to do with all templates (so we can ditch lodash)

PR #1081, opened 2024-01-11
feat: support w3c value property, even if alongside regular value

jorenbroekema:
*Description of changes:*

Summary:
- Support $value by replacing every instance that refers to `value` with `$value ?? value`, so that `$value` takes precedence while allowing `value` next to it still as a token group name, to be compliant with the spec
- Support $type by adding a preprocessing step so that design tokens inherit the $type from its ancestor tree if available. Other than this, type wasn't used in general in Style-Dictionary, but uses CTI structure instead to determine token types. This will be fixed in another PR.
- Support $description property by supporting it in createPropertyFormatter comment utility that adds it as a comment in the output, this was a very small change only
- add the above 3 props to the DesignToken type interface
- clean up eslint config and enforce some extra rules to prevent human errors (like leaving in an `it.only`, debugging related `console.log`, leftover `var`s potentially getting hoisted and causing bugs, etc.)

### DIFF
--- a/__integration__/w3c-forward-compat.test.js
+++ b/__integration__/w3c-forward-compat.test.js
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+ * the License. A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+import { expect } from 'chai';
+import StyleDictionary from 'style-dictionary';
+import { fs } from 'style-dictionary/fs';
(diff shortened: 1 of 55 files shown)


__integration__/w3c-forward-compat.test.js
```diff
@@ -0,0 +1,68 @@
...
+describe('integration', () => {
+  afterEach(() => {
+    clearOutput(buildPath);
+  });
+
+  describe('W3C DTCG draft spec forward compatibility', async () => {
+    const sd = new StyleDictionary({
+      tokens: {
+        colors: {
+          $type: 'color', // $type should be inherited by the children
```
jorenbroekema, line 30: In this PR, `$type` is not yet supported, nor its inheritance down the token tree.

Therefore you will notice that the custom/css/color transform doesn't apply, and in the snapshot, the color output is still in hex. In a future PR we should adjust this snapshot to be rgb output, and ensure $type and $type inheritance is supported.

__tests__/exportPlatform.test.js
```diff
@@ -234,156 +234,210 @@ describe('exportPlatform', () => {
     expect(actual).to.eql(expected);
   });
 
-  // describe('token references without .value', async () => {
```
jorenbroekema, line 237: These tests were disabled for some reason, so I re-enabled them, since they do work fine and provide coverage

six7 (approved):
looks good to me!

PR #1097, opened 2024-01-26
fix: escape double quotes for ts outputStringLiterals

andnorda:
Escaping double quotes when generating string literal types is useful when a token value is referencing a font family name.


jorenbroekema:
Thanks again for contributing :)! this will be included in the next prerelease

PR #1100, opened 2024-02-05
W3C Syntax refactor to allow value & type token names/groups

jorenbroekema:
*Description of changes:*

Allows the use of "value" and "type" as token names or groups.
The sacrifice is that you cannot combine W3C with non-W3C syntax, but that use-case seems less important to support.

### DIFF
--- a/.changeset/ninety-geckos-build.md
+++ b/.changeset/ninety-geckos-build.md
@@ -2,4 +2,4 @@
 'style-dictionary': patch
 ---
 
-Expose typeW3CDelegate utility. Don't take "value" into account anymore to determine that it's a design token, use $value.
+Expose `typeDtcgDelegate` utility. Don't take `value` into account anymore to determine that it's a design token, use `$value`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,14 +22,14 @@
 
 ### Patch Changes
 
-- 0c1a36f: Fix small issue in type w3c delegate utility type tracking.
-- 0c1a36f: Expose typeW3CDelegate utility. Don't take "value" into account anymore to determine that it's a design token, use $value.
+- 0c1a36f: Fix small issue in type DTCG delegate utility type tracking.
+- 0c1a36f: Expose `typeDtcgDelegate` utility. Don't take `value` into account anymore to determine that it's a design token, use `$value`.
 
 ## 4.0.0-prerelease.9
 
 ### Minor Changes
 
-- 294fd0e: Support W3C Draft specification for Design Tokens, by adding support for $value, $type and $description properties.
+- 294fd0e: Support Design Token Community Group Draft specification for Design Tokens, by adding support for $value, $type and $description properties.
 
 ### Patch Changes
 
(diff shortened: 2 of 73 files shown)


__integration__/__snapshots__/customFormats.test.snap.js
```diff
@@ -962,7 +962,8 @@ snapshots["inline custom with new args should match snapshot"] =
   },
   "options": {
     "otherOption": "Test",
-    "showFileHeader": true
+    "showFileHeader": true,
+    "usesW3C": false
```
kachkaev, line 966: I wonder if this should be `usesDtcg` instead. The spec at is not a W3C one. A working draft is not a W3C standard (and there is a small chance that it’ll never be one). If I understand correctly, usage of ‘W3C‘ in code or comments is misleading and discouraged. Related discussion: https://github.com/design-tokens/community-group/issues/164

jorenbroekema: Thanks, that makes sense! I've adjusted the PR and also got rid of any other W3C term occurrences.

In the debate of "userId" versus "userID", I'm on the side of "userId" since fairly recently, so using all-caps for such abbreviations is still a bit engrained in my habits sometimes 😅

kachkaev: Yeah `strictCamelCase` vs `looseCAMELCase` is a simpler rule to follow. Even though it may produce vars that are odd-looking at first glance (e.g. `myUiConfig`), using the rule everywhere frees up some cognitive resources. There is no need to think through each individual case.

.changeset/ninety-geckos-build.md
```diff
@@ -2,4 +2,4 @@
 'style-dictionary': patch
 ---
 
-Expose typeW3CDelegate utility. Don't take "value" into account anymore to determine that it's a design token, use $value.
+Expose typeDTCGDelegate utility. Don't take "value" into account anymore to determine that it's a design token, use $value.
```
kachkaev, line 5:
```suggestion
Expose typeDtcgDelegate utility. Don't take "value" into account anymore to determine that it's a design token, use $value.
```

```suggestion
Expose `typeDtcgDelegate` utility. Don't take `value` into account any more to determine that it's a design token, use `$value`.
```

CHANGELOG.md
```diff
@@ -22,14 +22,14 @@
 
 ### Patch Changes
 
-- 0c1a36f: Fix small issue in type w3c delegate utility type tracking.
-- 0c1a36f: Expose typeW3CDelegate utility. Don't take "value" into account anymore to determine that it's a design token, use $value.
+- 0c1a36f: Fix small issue in type DTCG delegate utility type tracking.
+- 0c1a36f: Expose typeDTCGDelegate utility. Don't take "value" into account anymore to determine that it's a design token, use $value.
```
kachkaev, line 26:
```suggestion
- 0c1a36f: Expose typeDtcgDelegate utility. Don't take "value" into account anymore to determine that it's a design token, use $value.
```

```suggestion
- 0c1a36f: Expose `typeDtcgDelegate` utility. Don't take `value` into account any more to determine that it's a design token, use `$value`.
```

.changeset/ninety-geckos-build.md
jorenbroekema: good catch

PR #1112, opened 2024-03-04
fix: outputReferenceFallbacks for css/variables formatter

tkgroot:
*Issue #, if available:*

If options were provided with outputReferences and outputReferenceFallbacks set to `true` did not add the fallbacks when running with the css/variables formatter.

*Description of changes:*

- outputReferenceFallbacks options are passed through to up until the createPropertyFormatter function, which was previously not the case
- add test to verify outputReferenceFallbacks are set
- update test snapshot


tkgroot:
created a PR for the new v4 version as well. it can be found here #1113

jorenbroekema:
Since v3 is essentially in maintenance mode now that v4 is coming soon, I'd prefer to only merge high prio bugfixes/security vulnerabilities, is it okay to only do this fix to the v4 branch?

tkgroot:
TL;DR: I would appreciate if this bugfix would make it into the current version.

tbh, I find that a bit weird to put v3 in maintenance mode before v4 is actually released. Do not understand the reasoning behind. I will not upgrade my current setup to a running beta because v4 is coming soon (especially because I dislike the term "coming soon"). 
My current setup does require that bugfix and I'm, like I said, not keep to adapt a beta version.

jorenbroekema (approved):

tkgroot:
Oh my, I didn't know that this was such a hassle. appreciate the effort you put in to get this released. 💯

jorenbroekema:
v4 is stable and merged to main by now, so closing this as fixed in v4

PR #1137, opened 2024-03-26
Adding all types to index.d.ts

### LINKED ISSUE #1135: How to access FormatterArguments types, etc.
Hey @jorenbroekema,

I have no idea how to access some types, like `FormatterArguments`.

They are not exported from `style-dictionary/types` and also not available through something like `Format['FormatterArguments']`.

The way you use it in here, seems to not be typical typescript, so can you help me find out how I can access those types? Thanks.

One of the places where I have this issue is here: https://github.com/lukasoppermann/style-dictionary-utils/pull/65/files#diff-4c8b203eb592cb0973b6227b4be99616520e7b0351981b54dfbd320fd0aa117aR3

lukasoppermann:
Fixes #1135

Adds all types to index.d.ts to be used by others e.g. when building custom formatters.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,6 +1,6 @@
 export type { Action } from './Action.d.ts';
 
-export type { PlatformConfig, Config } from './Config.d.ts';
+export type { PlatformConfig, Config, LocalOptions, LogConfig } from './Config.d.ts';
 
 export type {
   DesignToken,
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -1,6 +1,6 @@
 export type { Action } from './Action.d.ts';
 
-export type { PlatformConfig, Config } from './Config.d.ts';
+export type { PlatformConfig, Config, LocalOptions, LogConfig } from './Config.d.ts';
```
jorenbroekema, line 3:
```suggestion
export * from './Config.d.ts';
```

Let's go all out and export everything, easier to maintain as well.

jorenbroekema: Actually.... that makes them public API and I can't break them then :\ some of them are more meant as internal stuff,  so your way is better maybe

types/index.d.ts
```diff
@@ -9,13 +9,13 @@ export type {
   TransformedTokens,
```
jorenbroekema, line 9: Let's add Dictionary as well here, pretty commonly used especially in format hook

jorenbroekema:
Can you run `npx changeset` locally to add a changelog entry for this? patch is fine

lukasoppermann:
Hey, hope this is fine. Good catch on Dictionary.

jorenbroekema (approved):

Issue #882, opened 2022-10-17
[Question] is it possible to selectively output references?


KenjiCrosland:
I'm looking to use 3.0 outputReferences but there are several tokens we filter out of our main output file. Is it possible to only output the references for non-filtered tokens?

For example, lets say we have an `_options` folder whose values get used in our final output file but the values of those options are not in the output file. When I use outputReferences these option token names get referenced even though they aren't in the final outputted file. Could we have the values of these instead?

woylie:
We're having the same situation. In our case, we define the base colors in one file, and then define one additional color file for each theme. The theme color files reference the base colors, and the base color file is filtered out from the final output.

The ideal behavior for this situation would be:

- referenced value in final output -> use reference
- referenced value not in final output -> use literal value

nahiyankhan:
Very much doable! On our end, we have a custom CTI that has a layer field which could be `base`, `semantic`, or `component`. In our build config we then have a filter that essentially makes sure `token.attributes.layer != base`.

woylie:
> Very much doable! On our end, we have a custom CTI that has a layer field which could be `base`, `semantic`, or `component`. In our build config we then have a filter that essentially makes sure `token.attributes.layer != base`.

That's not the issue, we're doing that as well. The issue is that references to the filtered tokens are still included in the output.

For example:

`base.json`

```json
{
  "color": {
    "base": {
      "white": { "value": "#ffffff" }
    }
  }
}
```

`some.theme.json`

```json
{
  "color": {
    "background": { "value": "{color.base.white}" },
    "button": {
      "background": { "value": "{color.background.white}" }
    }
  }
}
```

Now if you filter `base.json` from the output _and_ set `outputReferences` to `true`:

```js
{
  platforms: {
    css: {
      transformGroup: "css",
      files: [{
        destination: "variables.css",
        format: "css/variables",
        filter: "baseColorFilter",
        options: {
          outputReferences: true
        }
      }]
    }
  }
}
```

The desired output would be:

```css
--color-background: #ffffff;
--color-button-background: --color-background;
```

So references would only be used if the referenced value is part of the output. Right now, this doesn't work, though, because Styledict tries to use references even if the referenced value is not included in the final output, and since it cannot resolve the references, the build will fail.

nahiyankhan:
@woylie You are right, I misinterpreted the problem. My bad! I appreciate your example. I get the issue now. Just thought more about it and have a likely solution.

[...]

Issue #859, opened 2022-08-31
[QUESTION] When `filteredReferenceCount > 0` and `outputReferences` fallback action?


fredboyle:
When `filteredReferenceCount > 0` and `outputReferences: true` is there a mechanism that can be used to force that specific token or all tokens to "fall back" to the referenced value instead of outputting a reference to another CSS custom property/Sass variable/etc?

We're attempting to create either a value transform or a new format but it's quite complex since there isn't a direct hook for overriding or extending the `css/variables` formatter output when the above condition is true.

Hopefully this makes sense. Any help and guidance is greatly appreciated.

fredboyle:
Doing some digging and it appears `GroupMessages.GROUP.FilteredOutputReferences` has what might be needed.

If something like this was added:
```javascript
if (options.valuesForFilteredOutReferences === true && GroupMessages.GROUP.FilteredOutputReferences.contains(ref.name)) {
    return token.value;
}
```

At this location:
https://github.com/amzn/style-dictionary/blob/28787befb1a2c7173c64a2a4bef5b029d0799cce/lib/common/formatHelpers/createPropertyFormatter.js#L116

Would that do the trick? 
I haven't fully dug into the `GroupMessages` object so I may be off on that piece.

The above basically short circuits the reference injection since it wouldn't make sense. The `GroupMessages.GROUP.FilteredOutputReferences` would need some cleanup as well to prevent unnecessary warning output during the build process.

Would be great to get some thoughts from the maintainers and the community.

jakobe:
@fredboyle for the `css/variables` format and your custom formats this can be achieved by using the `outputReferenceFallbacks` option: https://amzn.github.io/style-dictionary/#/formats?id=custom-format-helpers

PR #1145, opened 2024-04-03
feat: allow outputReferences function, publish filter util for it

jorenbroekema:
*Issue #, if available:*
fixes https://github.com/amzn/style-dictionary/issues/859 (with this PR feature, you output the resolved value instead of a broken ref with a fallback to make it work)
fixes https://github.com/amzn/style-dictionary/issues/882

- [x] bit more docs on outputReferences not only allowing boolean but also an outputReferences function


lukasoppermann:
Hey, just to make sure, this allows any custom function as a filter, correct? But you also provide one that removes references that are not in the output, right?

jorenbroekema:
> Hey, just to make sure, this allows any custom function as a filter, correct? But you also provide one that removes references that are not in the output, right?

Yes correct, I've just added a bunch more docs to illustrate this.

PR #1176, opened 2024-05-02
fix: changing FormatHelpers.setSwiftFileProperties type function to return object

vsjolander:
Docs specify that return type should be `Object` but it's `string` in [FormatHelpers.d.ts](https://github.com/amzn/style-dictionary/blob/17f4cb2f30bd002dfd55d6ef8c5bee4138de8d64/types/FormatHelpers.d.ts#L76) file.

*Description of changes:*
changing FormatHelpers.setSwiftFileProperties type function to return object

### DIFF
--- a/types/FormatHelpers.d.ts
+++ b/types/FormatHelpers.d.ts
@@ -73,5 +73,9 @@ export interface FormatHelpers {
     options: object,
     objectType: string,
     transformGroup: string,
-  ) => object;
+  ) =>  {
+    objectType?: string;
+    import?: string[];
+    accessControl?: string;
+};
 }
--- a/types/FormatHelpers.d.ts
+++ b/types/FormatHelpers.d.ts
@@ -73,5 +73,5 @@ export interface FormatHelpers {
     options: object,
     objectType: string,
     transformGroup: string,
-  ) => string;
+  ) => object;
 }


types/FormatHelpers.d.ts
```diff
@@ -73,5 +73,5 @@ export interface FormatHelpers {
     options: object,
     objectType: string,
     transformGroup: string,
-  ) => string;
+  ) => object;
```
jorenbroekema, line 76:
```suggestion
  ) =>  {
    objectType?: string;
    import?: string[];
    accessControl?: string;
};
```
To be even more specific

jorenbroekema (changes requested):

jorenbroekema:
Sidenote: in v4 branch this is fixed already

jorenbroekema (approved):

jorenbroekema:
In v4 the type interfaces aren't separated from the JS code itself anymore, typescript support is now first-class.

If you really need the interface separated you could do this:
```js
import { setSwiftFileProperties } from 'style-dictionary/utils';

const foo = setSwiftFileProperties(...);
//      ^^^   foo is typed implicitly because the return type of setSwiftFileProperties is known, no more need to typecast it

declare type fooType = typeof foo;
```

jorenbroekema:
Since v4 is stable and merged to main, no need to merge this

PR #1240, opened 2024-06-13
feat: allow not throwing on broken refs

jorenbroekema:
Style Dictionary is being used now as a Tokens parser & resolver and not just for building outputs, which means that it makes more sense now to allow not throwing on broken references, which is a token author error rather than a misconfiguration of Style Dictionary.

This change will allow tool makers to use Style Dictionary to for example show a list of tokens with their resolved and original values and highlighting broken references.

TODO: 
- [x] docs
- [x] changeset

### DIFF
--- a/.changeset/chilly-numbers-work.md
+++ b/.changeset/chilly-numbers-work.md
@@ -1,0 +1,21 @@
+---
+'style-dictionary': minor
+---
+
+Allow not throwing fatal errors on broken token references/aliases, but `console.error` instead.
+`resolveReferences` and `getReferences` both allow passing `throwOnBrokenReferences` option, setting this to false will prevent a fatal error.
+You can also configure this on global/platform `log` property:
+
+```json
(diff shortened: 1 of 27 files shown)


floscr:
Nice :+1: 

Would it make sense to add resolvable tokens to the test case, to verify that other tokens resolve with a missing one?

E.g.:

```js
        tokens: {
          bar: {
            value: '12px',
            type: 'other',
          },
          foo: {
            value: '{missing}',
            type: 'other',
          },
          baz: {
            value: '{bar}',
            type: 'other',
          },
        },
```

jorenbroekema:
Yeah good idea, I'll add a test for that!

floscr:
Can we add `silent` as well here, I want to handle errors in the UI and not show them in the console in production :pray:

floscr:
Works as expected, thank you :raised_hands:

PR #1241, opened 2024-06-14
chore(docs): starting to style the docs a bit

dbanksdesign:
*Issue #, if available:*

*Description of changes:*

This pull request is mainly to do some styling and branding of the docs a bit. There will be other PRs to update some of the content of the docs site as well. 

Some things this includes:
* Using Style Dictionary _in_ the docs for the design tokens. This helps because the docs needs color information for monaco editor instances and expressive code, as well as starlight components and shoelace components. 

Todo:

- [ ] Standardize "grey" and "gray"
- [x] Remove commented out code
- [ ] Accessibility testing to make sure all the color choices are accessible


docs/theme/build.ts
```diff
@@ -0,0 +1,245 @@
...
+    },
+    filters: {
+      css: (token) => {
+        return token.path[0] === 'color';
+      },
+    },
+  },
+});
+
+['dark', 'light'].forEach(async (mode) => {
```
jorenbroekema, line 185: async within forEach is technically not supported and may result in buggy behavior, better to use a for of loop here.

Or, if you want to run things in parallel, use this pattern:

```js
// wrapper async function for putting stuff in sequence
const runSD = async (mode) => {
  const sdExt = await sd.extend({ ... });
  await sdExt.buildAllPlatforms();
};

// to run stuff in parallel
await Promise.all(['dark', 'light'].map(runSD));
```

docs/theme/build.ts
```diff
@@ -0,0 +1,245 @@
+import StyleDictionary from 'style-dictionary';
```
jorenbroekema, line 1: Hah this is super cool, using design tokens and "eating our own dogfood" to create the Monaco themes, I like it :)!

package.json
```diff
@@ -44,8 +44,9 @@
   ],
   "scripts": {
     "build": "node scripts/inject-version.js && tsc --emitDeclarationOnly",
-    "docs:start": "astro dev --root ./docs",
-    "docs:build": "astro check  --root ./docs && astro build --root ./docs",
+    "docs:token-build": "npx tsx docs/theme/build.ts",
```
jorenbroekema, line 47: `tsx` should also be installed as a devDependency imo, I think it works for you locally because you've probably got it installed globally?

docs/theme/tokens/application/editor.ts
```diff
@@ -0,0 +1,164 @@
+export const editor = {
+  background: { value: '{color.background.primary.value}' },
```
jorenbroekema, line 2: I wonder if we'd be setting a better example by using the DTCG format and using $value and then just putting the $type at the very very top group level, so it only needs to be set once (since all of the tokens here are color tokens it seems?).

docs/theme/build.ts
```diff
@@ -0,0 +1,245 @@
...
+          prefix: 'sl',
+          files: [
+            {
+              destination: `${mode}.variables.css`,
+              format: 'css/variables',
+              filter: (token) => {
+                // light mode don't include the core colors
+                // because default is dark mode which will have the core colors
+                if (mode === 'light') {
+                  return token.path[0] === 'color' && token.path[1] !== 'core';
```
jorenbroekema, line 228: Generally as a best practice I'd personally advise against relying on the `token.path` property.

I think `token.$type` is more reliable and explicit for determining type. For whether it's a core token or not, maybe we should consider putting all core tokens in a folder so we can use `token.filePath` to know whether some token is a core token or not.

I see the `token.path` being relied on in quite a few places here and it reminds me a bit of relying on CTI structure (which is just syntactic sugar on top of the token path), which I kind of wanted to get away from in v4 and in light of DTCG spec. I do see the convenience and making lots of token "attributes" implicit through its token taxonomy/hierarchy, but in my experience it makes things a bit brittle when restructuring tokens.

docs/src/styles.css
```diff
@@ -1,18 +1,119 @@
...
+html {
+  font-feature-settings: 'ss01', 'ss02', 'case', 'salt';
 }
 
 /* Light mode colors. */
 :root[data-theme='light'] {
-  --sl-color-accent-low: #55b3af;
-  --sl-color-accent: #047771;
-  --sl-color-accent-high: #07524f;
+  /* --accent-hue: 180;
```
jorenbroekema, line 16: I think this file needs a bit of cleanup? I'm assuming you puts the CSS vars in comments here to make it easy to lookup without having to tab in/out of browser vs editor, but correct me if I'm wrong in that assumption ;)

Does it get removed by Astro/Vite when building the docs site, e.g. CSS minify step?

docs/src/components/colors.ts
```diff
@@ -0,0 +1,78 @@
+import { LitElement, css, html } from 'lit';
+import { darkTokens } from '../../theme/tokens/dark';
+
```
jorenbroekema, line 3: image

I think the style for important marked file in FileTree component is a bit off here on dark mode

.prettierignore
```diff
@@ -1,3 +1,4 @@
 .changeset/
 **/*.snap.js
 /docs/dist
+/docs/**/*.mdx
```
jorenbroekema, line 4: Was prettier messing it up or?

docs/src/components/sd-playground.ts
```diff
@@ -201,67 +222,69 @@ class SdPlayground extends LitElement {
...
-              stroke="currentColor"
-              stroke-width="1.5"
-              stroke-miterlimit="10"
-              stroke-linecap="round"
-            />
-          </svg>
-        </sl-radio-button>
-      </sl-radio-group>
-      <slot name="monaco-editor"></slot>
+      <div class="foo">
```
jorenbroekema, line 225: I think the div can be removed right? Seemed okay at least when I did this locally. Not sure if the diff here is just that wrapper div or if you've changed other things in the template.

docs/theme/build.ts
```diff
@@ -0,0 +1,245 @@
...
+        transitive: true,
+        filter: (token) => {
+          return token.path[0] === 'color';
+        },
+        transform: (token) => {
+          const color = parse(token.value);
+
+          if (!color) {
+            console.log(token);
+          }
```
jorenbroekema, line 50: cleanup

docs/theme/build.ts
```diff
@@ -0,0 +1,245 @@
...
+  settings: {
+    foreground?: string;
+    fontStyle?: string;
+  };
+}
+
+const sd = new StyleDictionary({
+  // log: {
+  //   verbosity: 'silent',
+  // },
```
jorenbroekema, line 36: cleanup

docs/theme/build.ts
```diff
@@ -0,0 +1,245 @@
...
+        transitive: true,
+        filter: (token) => {
+          return token.path[0] === 'color';
+        },
+        transform: (token) => {
+          const color = parse(token.value);
+
+          if (!color) {
+            console.log(token);
+          }
```
jorenbroekema, line 74: cleanup

jorenbroekema (changes requested):
Awesome work, docs site looking much shinier now :)! And I'm a big fan of tokenizing some of the stuff and using SD to generate the CSS / monaco themes, cool stuff.

Most of my comments are cleanup/nitpick related, but a couple do require changes I think, let me know what you think

docs/src/content/docs/style/index.mdx
```diff
@@ -0,0 +1,246 @@
+---
```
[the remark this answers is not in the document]
dbanksdesign: Yup that is correct and good idea!

.prettierignore
dbanksdesign: Yes, because it removes empty lines and for some features of expressive code need empty lines: https://expressive-code.com/key-features/text-markers/#adding-labels-to-line-markers

docs/src/components/sd-playground.ts
dbanksdesign: haha yup, I was playing around and forgot to remove it

docs/theme/build.ts
dbanksdesign: I'm actually using path specifically, and not trying to use the token type (because all the tokens are actually color type). So how would you target tokens based on the dictionary object structure if not by `token.path` ? Or would you re-organize the dictionary object?

jorenbroekema: > how would you target tokens based on the dictionary object structure if not by `token.path` ?

Well, I wouldn't, because imo it's a bad practice to rely upon that, since the structure can be changed and it wouldn't be obvious that that would break your transforms. I would use `token.$type` and `token.filePath` instead, or some metadata prop that lives in the tokenset at the top but does not have a value prop so SD wouldn't classify it as a token, it would just be some metadata living there.

dbanksdesign: Couldn't the same be said about `token.filePath` ? They are both paths, if I change a file name that would break my transforms too. I still don't really see why using the path is bad practice. Also, for this specifically filePath can't be used because the tokens are added via `tokens` and not `source` or `include`. Do you have a specific example you can point to for the best practice?

jorenbroekema: I see your point, they're similar in that the connection between your tokens structure/categorization and your tokens transforms logic isn't explicit and easy to forget about, causing something to break that you didn't expect.

It would be good to think outside the box and see if there's a way we can do it that improves on this problem ("token grouping" that's not tied to token structure/taxonomy), but I agree with you that `token.filePath` isn't that much better than using `token.path`. Feel free to resolve this!

jorenbroekema:
I just realized this PR is gonna have the same merge conflict hell that I had to deal with when rebasing v4 on main last Friday. Just give me a heads up when you consider this PR "done" and I can help with this, I think I got the most context on the code atm to be able to resolve all these pesky conflicts.

PR #1244, opened 2024-06-17
feat: recursively expand nested composite tokens

Imable:
*Description of changes*

My organization structures design tokens such that design tokens are dynamically linked with a color from our color palette and its foreground color. This requires recursive expansion of composite tokens, which until now was marked `// TODO` in code.

I reused the `expandTokensRecurse` function to recursively expand the expanded tokens of a composite token. If the composite token only contains leafs (no composite token), no work is done.

### DIFF
--- a/.changeset/lucky-buckets-wave.md
+++ b/.changeset/lucky-buckets-wave.md
@@ -1,0 +1,10 @@
+---
+'style-dictionary': minor
+---
+
+Some fixes for Expand utility:
+
+- Array values such as `dashArray` property of `strokeStyle` tokens no longer get expanded unintentionally, `typeof 'object'` check changed to `isPlainObject` check.
+- Nested object-value tokens (such as `style` property inside `border` tokens) will now also be expanded.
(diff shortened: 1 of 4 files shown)


jorenbroekema:
Hey, thanks for raising this, I made some minor adjustments to simplify it a little and I also adjusted your test to test with a DTCG use case of nested object-values (border -> strokeStyle), your test example is a bit more custom and technically not compliant with DTCG because color values aren't supposed to be object-values (as of now).

That said, I did also run your test locally and it does pass with my adjustments, with one small change:
```js
            Background: {
              Surface: {
                Background: {
                  value: '#ffffff', // <-- {Gray.0.Background} in my test because it can remain a ref, due to resolved value being a string value
```

I found an edge case as well with the strokeStyle test because it has a multi-value dimension (Array type), and I was doing `typeof 'object'` checks so it would erroneously expand array values, now it checks whether the value is a plain object which fixes that bug.

I also added a changeset ensuring it will be in next release.

jorenbroekema:
Thanks for contributing :)!

PR #1265, opened 2024-07-03
fix: missing filepath in falsy token values

tim-richter:
_Issue #, if available:_

_Description of changes:_

The `if (prop)` check in `traverseObj` checks, if the current value is 'truthy'. In the case of a token with a `0` value, this check will evaluate to false, and the filePath will not be part of the resulting obj, because we never traverse into the containing object.

In our case, we sometimes opt in to filter tokens by filePath, so with the recent 4.0 version, we lost a bunch of tokens in the output, because they didn't contain `filePath` and were filtered out. This is a regression to 3.x.

Reproduction: https://stackblitz.com/edit/vitejs-vite-h8axgu?file=config.js (the console should output two tokens: one with a positive integer and one with the 0 value)

---

Probably debatable if empty strings should be part of the tokens, but I opted to only check for null/undefined for now. I don't have strong feelings about this, tho.

I don't have a lot of experience in this repo, so I didn't know where to put tests for this. There is `combineJSON.test.js`, but I didn't see tests for the filePath in there. If someone gives me pointers, I can probably also add tests for this bug.

---

### DIFF
--- a/lib/utils/combineJSON.js
+++ b/lib/utils/combineJSON.js
@@ -33,7 +33,7 @@ import { detectDtcgSyntax } from './detectDtcgSyntax.js';
 function traverseObj(obj, fn) {
   for (let key in obj) {
     const prop = obj[key];
-    if (prop !== null && prop !== undefined) {
+    if (prop != null) {
       fn.apply(null, [obj, key, prop]);
       if (typeof prop === 'object') {
         traverseObj(prop, fn);
--- a/lib/utils/combineJSON.js
+++ b/lib/utils/combineJSON.js
@@ -33,7 +33,7 @@ import { detectDtcgSyntax } from './detectDtcgSyntax.js';
 function traverseObj(obj, fn) {
   for (let key in obj) {
     const prop = obj[key];
-    if (prop) {
+    if (prop !== null || prop !== undefined) {
       fn.apply(null, [obj, key, prop]);
       if (typeof prop === 'object') {
         traverseObj(prop, fn);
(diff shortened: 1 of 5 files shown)


lib/utils/combineJSON.js
```diff
@@ -33,7 +33,7 @@ import { detectDtcgSyntax } from './detectDtcgSyntax.js';
 function traverseObj(obj, fn) {
   for (let key in obj) {
     const prop = obj[key];
-    if (prop) {
+    if (prop !== null && prop !== undefined) {
```
jorenbroekema, line 36:
```suggestion
    if (prop != null) {
```
Single equals null matches both `null` and `undefined`, so it's a shorthand for this.
ESLint also allows this specific instance of doing single equals vs double equals.

jorenbroekema (changes requested):
Could you run `npx changeset` locally and create a patch changeset for this change?
~~I also wonder whether it would be a good idea to add a regression test, let me know if you need help with that though.~~
Just read your comment about the test, feel free to just put it anywhere in the combine JSON test file, I can clean it up a bit if needed. You're right in that we don't have tests validating the presence of filePath props in there currently

lib/utils/combineJSON.js
tim-richter: Yup thanks, added this in latest commit.

tim-richter:
Hey @jorenbroekema, I've added tests for this in combineJSON.test.js. I just added the falsy values cases in "not_circular.json". If you want a seperate test case let me know.

Also added the changset :slightly_smiling_face:

jorenbroekema (approved):

dbanksdesign (approved):

PR #1269, opened 2024-07-08
feat: add preExpand meta info to expanded tokens

jorenbroekema:
_Issue #, if available:_
A user has told me that it's kinda difficult for him to apply custom transforms to typography token properties such as letterSpacing or lineHeight, because after expanding them they are typed as `dimension`/`number` respectively, and therefore it's hard to set a filter function that matches only letterSpacing/lineHeight tokens when the types have been aligned / reduced to DTCG types.

_Description of changes:_

Add metadata under `$extensions.['com.styledictionary']` (where metadata should go from now on, following DTCG guidelines) so you can set a transform filter that targets expanded typography -> lineHeight tokens

### DIFF
--- a/lib/utils/expandObjectTokens.js
+++ b/lib/utils/expandObjectTokens.js
@@ -178,11 +178,22 @@ export function expandToken(token, opts, platform) {
       expandedTokenObjRef = expandedTokenObjRef[index + 1];
     }
     Object.entries(objectVal).forEach(([key, value]) => {
-      expandedTokenObjRef[key] = {
-        ...copyMeta,
+      expandedTokenObjRef[key] = deepmerge(copyMeta, {
         [`${uses$ ? '$' : ''}value`]: value,
         [`${uses$ ? '$' : ''}type`]: getTypeFromMap(key, compositionType, typesMap),
-      };
+        // store the original composite token type as well as which prop it was
+        // so that for example a lineHeight property of a typography token, when expanded
+        // can still be matched as a lineHeight token, even though its type is now "number" after the expansion
+        $extensions: {
+          'com.styledictionary': {
+            preExpand: {
+              type: compositionType,
+              prop: key,
+            },
+          },
+        },
+      });
+
       if (Array.isArray(expandedTokenObjRef[key].path)) {
         // if we're expanding on platform level, we already have a path property
         // we will need to adjust this by adding the new keys of th expanded tokens to the path array
--- a/.changeset/swift-scissors-cover.md
+++ b/.changeset/swift-scissors-cover.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': minor
+---
+
+When expanding object-value tokens, add some metadata about what type the token belonged to before expanding, and which property. This is useful e.g. for transforms to target specific object-value properties like lineHeights. When expanding, the type is aligned to the `number` type (DTCG), but you may want to transform `lineHeight` (not a known DTCG type) tokens but not all `number` tokens.
--- a/__tests__/StyleDictionary.test.js
+++ b/__tests__/StyleDictionary.test.js
@@ -532,14 +532,38 @@ Use log.verbosity "verbose" or use CLI option --verbose for more details.
           color: {
             type: 'color',
             value: '#000',
+            $extensions: {
+              'com.styledictionary': {
+                preExpand: {
+                  prop: 'color',
+                  type: 'border',
+                },
+              },
+            },
           },
           style: {
             type: 'strokeStyle',
             value: 'solid',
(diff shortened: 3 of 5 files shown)


lib/utils/expandObjectTokens.js
```diff
@@ -178,11 +178,22 @@ export function expandToken(token, opts, platform) {
...
         [`${uses$ ? '$' : ''}value`]: value,
         [`${uses$ ? '$' : ''}type`]: getTypeFromMap(key, compositionType, typesMap),
-      };
+        // store the original composite token type as well as which prop it was
+        // so that for example a lineHeight property of a typography token, when expanded
+        // can still be matched as a lineHeight token, even though its type is now "number" after the expansion
+        $extensions: {
+          'com.styledictionary': {
+            preExpand: {
+              type: compositionType,
```
dbanksdesign, line 190: Should this be `compositionType` to be more explicit?

lib/utils/expandObjectTokens.js
```diff
@@ -178,11 +178,22 @@ export function expandToken(token, opts, platform) {
...
-      expandedTokenObjRef[key] = {
-        ...copyMeta,
+      expandedTokenObjRef[key] = deepmerge(copyMeta, {
         [`${uses$ ? '$' : ''}value`]: value,
         [`${uses$ ? '$' : ''}type`]: getTypeFromMap(key, compositionType, typesMap),
-      };
+        // store the original composite token type as well as which prop it was
+        // so that for example a lineHeight property of a typography token, when expanded
+        // can still be matched as a lineHeight token, even though its type is now "number" after the expansion
+        $extensions: {
```
dbanksdesign, line 187: This is the first time Style Dictionary is using `$extensions`. Do we expect to put more in here later?

dbanksdesign (changes requested):
While this change seems small, I think it warrants a larger discussion. This is the first time we are using `$extensions` part of the design token. Once we add this in, it will be difficult to remove. 

I would like to see some docs added in this PR as well. It would be helpful to see an example in docs that lays out the user problem this is solving.

lib/utils/expandObjectTokens.js
jorenbroekema: `compositeType` I guess? Although both is probably fine

lib/utils/expandObjectTokens.js
jorenbroekema: I'm thinking that perhaps it makes sense to move other properties that Style Dictionary adds to the token metadata under this namespace, as not to clash with metadata that the token author themselves may add. This could be planned for a future v5 I reckon

jorenbroekema:
I'll add some docs on this, I agree it makes sense to be clear about use cases 👍🏻

jorenbroekema:
Hey @dbanksdesign I added some docs, and while describing the use case of letterSpacing I started to realize that maybe we don't need to go through with this PR.

I'm arguing that e.g. for `typography.letterSpacing` you want to preserve the original type in metadata so you can apply a `letterSpacing`-only transform on it while still having it be a `dimension` token, making dimension-specific transforms also apply.

So essentially, this token has 2 types in a sense. But that's kinda stupid imo because a standalone token of type "letterSpacing" could never have 2 types, so expanded tokens would be unique in that capability; this doesn't make much sense.

So if a DTCG extension spec decides there are types that don't exist (yet) in DTCG, they just have to add them to the expandTypesMap:

```js
{
  expand: {
    typesMap: {
      typography: {
        letterSpacing: 'letterSpacing'
      }
    }
  }
}
```

And then the expanded `letterSpacing` properties of typography tokens will be of type `letterSpacing` and you can apply your transform on it. If you then need a StyleDictionary built-in transform to apply to those as well but they only apply to `dimension` tokens, StyleDictionary actually makes it pretty easy to do this:

```js
StyleDictionary.registerTransform({
  ...StyleDictionary.hooks.transforms['size/rem'],
  filter: (token, options) => {
    const type = options.usesDtcg ? token.$type : token.type;
    return ['dimension', 'fontSize', 'letterSpacing'];
  }
})
```
This overrides the built-in `'size/rem'` transform with a custom filter that includes `letterSpacing` tokens.
I think the above trick should be better documented though, it's useful for any platform/package that extends the DTCG types but also wants to have the built-ins apply to the extended types in some cases (e.g. `letterSpacing` is an extension of `dimension` in this example).

Conclusion: I'd like to close this PR since I don't see an urgent need for it anymore. I'll raise an issue to document the above thing better.

jorenbroekema:
https://github.com/amzn/style-dictionary/issues/1334 @dbanksdesign I found a use case for this PR to be reopened

Issue #1286, opened 2024-07-23
scss/map-deep formatting not applied to map values


Stun3R:
Hey 👋🏼 

I just encountered an issue related to formatting options.

It looks like the formatting options only format the SCSS variables' references, not the ones passed to the SCSS map.

**Config:**
```js
{
  sources: ["properties/**/*.json",
  platforms: {
    css: {
      transformGroup: ['attribute/cti', 'name/kebab', 'color/css', 'opacity/css', 'fontFamily/css'],
      buildPath: './dist',
      files: [
        {
          destination: 'style.css',
          format: 'css/variables',
          options: {
            showFileHeader: false,
            selector: '.common',
            formatting: {
              prefix: `--common-`,
            },
          },
        },
      ],
      expand: true,
    },
    scss: {
      transformGroup: ['attribute/cti', 'name/kebab', 'color/css', 'opacity/css', 'fontFamily/css'],
      buildPath: './dist',
      files: [
        {
          destination: 'tokens.scss',
          format: 'scss/map-deep',
          options: {
            showFileHeader: false,
            formatting: {
              prefix: `$common-`,
            },
          },
        },
      ],
    expand: true,
    },
  }
}
```

**CSS result:**
```css
.common {
  --common-duration-veryfast: 100ms;
  --common-duration-fast: 250ms;
  --common-duration-medium: 500ms;
  --common-duration-slow: 1000ms;
  --common-duration-veryslow: 2000ms;
  --common-opacity-highlight: 0.2;
  --common-opacity-disabled: 0.5;
  --common-opacity-loading: 0.8;
  --common-opacity-image-hover: 0.8;
  --common-breakpoint-mobile-small: 390px;
  --common-breakpoint-mobile: 600px;
  --common-breakpoint-tablet-small: 720px;
  --common-breakpoint-tablet: 960px;
  --common-breakpoint-desktop-small: 1280px;
  --common-breakpoint-desktop: 1920px;
  --common-effect-material-blur: 50px;
  --common-z-index-app-bar: 1100;
  --common-z-index-drawer: 1200;
  --common-z-index-modal: 1300;
  --common-z-index-snackbar: 1400;
  --common-z-index-tooltip: 1500;
}
```

**SCSS result**
```scss
$common-z-index-tooltip: 1500 !default;
$common-z-index-snackbar: 1400 !default;
$common-z-index-modal: 1300 !default;
$common-z-index-drawer: 1200 !default;
[...]

PR #1294, opened 2024-07-30
fix: respect formatting.prefix scss formats

jorenbroekema:
_Issue #, if available:_
https://github.com/amzn/style-dictionary/issues/1286


__tests__/formats/__snapshots__/scssMaps.test.snap.js
```diff
@@ -51,3 +51,53 @@ $tokens: (
...
+$foo-size-font-small: 12rem !default;
+$foo-size-font-large: 18rem !default;
+$foo-color-base-red: #ff0000 !default; // comment
+$foo-color-white: #ffffff !default;
+$foo-asset-icon-book: url("data:image/png;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0IiBmaWxsPSJub25lIiBzdHJva2U9ImN1cnJlbnRDb2xvciIgc3Ryb2tlLXdpZHRoPSIyIiBzdHJva2UtbGluZWNhcD0icm91bmQiIHN0cm9rZS1saW5lam9pbj0icm91bmQiIGNsYXNzPSJmZWF0aGVyIGZlYXRoZXItYm9vayI+PHBhdGggZD0iTTQgMTkuNUEyLjUgMi41IDAgMCAxIDYuNSAxN0gyMCI+PC9wYXRoPjxwYXRoIGQ9Ik02LjUgMkgyMHYyMEg2LjVBMi41IDIuNSAwIDAgMSA0IDE5LjV2LTE1QTIuNSAyLjUgMCAwIDEgNi41IDJ6Ij48L3BhdGg+PC9zdmc+") !default;
+
+$tokens: (
+  'size': (
+    'font': (
+      'small': $foo-size-font-small,
```
jorenbroekema, line 85: these were missing the `$foo-` prefix before this change. Meaning the SCSS result is broken (broken refs)

jorenbroekema:
Oops, looks like some unrelated minor docs improvements made it into this branch, if it's okay I guess we can also merge those while we're at it?

lib/common/templates/scss/map-deep.template.js
```diff
@@ -12,14 +12,15 @@
...
+  const { formatting } = options;
   let output = '';
   if (obj === null) {
     output += `''`;
   } else if (typeof obj === 'string') {
     output += `'${obj}'`;
   } else if (Object.hasOwn(obj, `${options.usesDtcg ? '$' : ''}value`)) {
     // if we have found a leaf (a property with a value) append the value
-    output += `$${obj.name}`;
+    output += `${formatting?.prefix ?? '$'}${obj.name}`;
```
dbanksdesign, line 23: It looks like we are missing other `formatting` options as well here like `suffix`, should we add those in too?

jorenbroekema:
Done, making sure the formatting options are all respected.

I tried reusing createPropertyFormatter but this wasn't possible due to the nested nature of these formats.
Suffix is already "respected" in the sense that for map-deep, the top part of it does reuse createPropertyFormatter, where suffix is respected. And in the deep-map below it, you wouldn't want the custom suffix to apply to the inner map syntax, both for flat/deep, that would break the SCSS syntax no matter what.

I think these formatting options are really awkward to be honest. Why would you ever want to change the separator if it breaks the output syntax? Same for prefix, this is better done as part of a token name transform, you're likely to break the output if you do it at the level of output property format. Same for suffix, this either breaks the output or you're better off adding a custom transform that adds a suffix to token values imo. Perhaps something to redesign for v5

dbanksdesign (approved):

PR #1295, opened 2024-08-01
fix: change fontWidth prop to fontStretch as per W3C spec

jorenbroekema:
_Description of changes:_
https://github.com/mdn/content/issues/35265 this was wrong in the MDN formal syntax heading and this mistake got adopted in the typography transform. This is a bug so even though technically breaking, acceptable to fix without a major bump imo.


jorenbroekema:
fontWidth apparently is the correct term, MDN just doesn't use it properly in all places. So no need for this PR

PR #1296, opened 2024-08-02
feat: cache SD methods, deprecate and replace ambiguous API

jorenbroekema:
- [x] update docs (since some of the methods have gained an `options` 2nd param, briefly explain that computing platformConfig / platform transformed tokens is cached now)
- [x] fix perf test which now fails (e.g. bump the threshold to 2000ms for now, although I do want to get it below 1 sec if possible later..)

I've started work on adding performance tests and defining budgets for some SD processes.
So far I've got a simple/basic test with a single token, as well as a set with 9000 tokens of which 6000 are references, with a ref depth of 3 (ref chain).
This is where it starts to slow down and using the Chrome DevTools performance tab I already managed to spot a significant bottleneck in the SD class methods. Many of these are called multiple times redundantly, and no caching happens for the expensive ones (mainly `exportPlatform`). This started a clean up process of the methods and my proposal is to deprecate (for removal in v5) the `getPlatform` and `exportPlatform` methods; they are ambiguous, not pure, and have unintuitive return types. Instead there are two new methods replacing them, more pure and clear about what they do.

I've also added in-memory caching to get rid of the redundant recomputes, with the added ability to turn this caching off for very advanced use cases (e.g. when doing tests with a single SD instance, recomputing token transforms/preprocessors while changing the config/tokens in between)

@dbanksdesign  I'm not sure if I should continue my perf hunt in this branch or create new ones, the 2nd perf test currently fails because it takes 1-2 seconds, before my changes this was 5+ seconds so I already sped things up by more than 2x and that's with unthrottled CPU (imagine how much it matters for shitty CPUs, e.g. in CI pipelines which is a majority use case for this library).

The main performance culprit some of the Tokens Studio users are reporting very likely has to do with resolving token references / expanding nested composites so that's what I'd be looking into next. Aggressive memoization of reference resolutions will probably make a huge difference, or making use of JS Map structure for cheaper accessing/searching of tokens (array is expensive for accessing, object is expensive for traverse, Map is way better for both)


__tests__/buildPlatform.test.js
```diff
@@ -24,7 +24,7 @@ describe('buildPlatform', () => {
 
   it("should throw if passed a platform that doesn't exist", async () => {
     await expect(StyleDictionaryExtended.buildPlatform('foobar')).to.eventually.rejectedWith(
-      'Platform "foobar" does not exist',
+      'Please supply a valid platform, "foobar" does not exist',
```
jorenbroekema, line 27: This "new" error message combines the two old messages, which both meant the exact same thing but were not deduped.

lib/StyleDictionary.js
```diff
@@ -352,6 +361,7 @@ export default class StyleDictionary extends Register {
       preprocessedTokens = expandTokens(preprocessedTokens, this.options);
     }
     this.tokens = preprocessedTokens;
+    this.allTokens = flattenTokens(/** @type {PreprocessedTokens} */ (this.tokens), this.usesDtcg);
```
jorenbroekema, line 364: moved, used to be the last ran platform specific transformed tokens (not intuitive at all), which doesn't make sense as a global SD prop. Now this prop is the parsed/preprocessed flattened tokens prior to running platforms.

lib/StyleDictionary.js
```diff
@@ -380,17 +390,51 @@ export default class StyleDictionary extends Register {
...
-   * @returns {Promise<TransformedTokens>}
+   * @param {{ cache?: boolean }} [opts]
    */
-  async exportPlatform(platform) {
-    await this.hasInitialized;
-
+  getPlatformConfig(platform, opts) {
     if (!platform || !this.platforms?.[platform]) {
-      throw new Error('Please supply a valid platform');
+      throw new Error(`Please supply a valid platform, "${platform}" does not exist`);
```
jorenbroekema, line 397: See earlier comment in test file about combining two messages doing the same thing into one

lib/StyleDictionary.js
```diff
@@ -541,35 +585,25 @@ export default class StyleDictionary extends Register {
         console.log(chalk.rgb(255, 140, 0).bold(err));
       }
     }
-
-    return exportableResult;
+    return { tokens: exportableResult, allTokens: flattenTokens(exportableResult, this.usesDtcg) };
```
jorenbroekema, line 588: We flatten these only once now after computing the transformed tokens for platform, instead of doing it like 3 times before this change, saves a lot of compute for large sets.

lib/StyleDictionary.js
```diff
@@ -541,35 +585,25 @@ export default class StyleDictionary extends Register {
...
+   * @deprecated use getPlatformConfig / getPlatformTokens instead
    * @param {string} platform
+   * @param {{ cache?: boolean }} [opts]
    * @returns
    */
-  async getPlatform(platform) {
+  async getPlatform(platform, opts) {
     await this.hasInitialized;
-    if (!this.platforms?.[platform]) {
-      throw new Error(`Platform "${platform}" does not exist`);
```
jorenbroekema, line 557: This is the redundant duplicate error, now combined with the other one

lib/StyleDictionary.js
```diff
@@ -380,17 +390,51 @@ export default class StyleDictionary extends Register {
...
+    return this._dictionaries[platform];
+  }
+
+  /**
+   * Public wrapper around _exportPlatform, returns only tokens object
+   * Here for backwards compatibility.
+   * @deprecated use getPlatformTokens instead
+   *
+   * @param {string} platform
+   * @param {{ cache?: boolean }} [opts]
```
jorenbroekema, line 423: Make it a reusable interface

dbanksdesign (approved):

dbanksdesign (approved):

Issue #1267, opened 2024-07-05
The size/rem tranform ads units (rem) to zero based values


bengie:
My setup is Styledictionary 4.0.0 with sd-transdorms 1.1.0 and now I get the `rem` unit appended to all `0` based values in CSS

```css
// before
---some-border-width: 0;
// now
--some-border-width: 0rem;
```
Expected result: no units appended to zero-based values.

The size/rem transform doesn't account for zero-based values, see https://github.com/amzn/style-dictionary/blob/main/lib/common/transforms.js#L824

An example from sd-transforms that does account for zero-based values, see: https://github.com/tokens-studio/sd-transforms/blob/main/src/transformDimension.ts#L3

jorenbroekema:
Yep would be good to follow the example from sd-transforms here and ignore 0 values, since 0 doesn't need a unit.

PR #1302, opened 2024-08-12
fix: zero value is returned without a unit at size/rem transform

yinm:
_Issue #, if available:_
#1267

_Description of changes:_
As the title says,  I fixed that zero value is returned without a unit at size/rem transform.

### DIFF
--- a/lib/common/transforms.js
+++ b/lib/common/transforms.js
@@ -832,6 +832,7 @@ export default {
       }
       const parsedVal = parseFloat(nonParsed);
       if (isNaN(parsedVal)) throwSizeError(token.name, nonParsed, 'rem');
+      if (parsedVal === 0) return '0';
       return parsedVal + 'rem';
     },
   },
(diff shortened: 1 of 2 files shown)


lib/common/transforms.js
```diff
@@ -832,6 +832,7 @@ export default {
       }
       const parsedVal = parseFloat(nonParsed);
       if (isNaN(parsedVal)) throwSizeError(token.name, nonParsed, 'rem');
+      if (parsedVal === 0) return '0';
```
jorenbroekema, line 835:
```suggestion
      if (parsedVal === 0) return Number.isInteger(nonParsed) ? 0 : '0';
```

Let's check that if the token value is the number 0, we leave it as a number. Please add a test for this as well :)

jorenbroekema (changes requested):

jorenbroekema (approved):

jorenbroekema:
Awesome, thanks for contributing a fix! Just waiting for @dbanksdesign to approve and look to release it

dbanksdesign (approved):

PR #1328, opened 2024-09-05
docs: fix `compose/object` parameters table

julien-deramond:
### Description

This PR fixes a rendering issue in the documentation at https://styledictionary.com/reference/hooks/formats/predefined/#composeobject. Indeed, the table containing the parameters is not rendered properly:

Screenshot 2024-09-05 at 10 23 22

The fix uses the same Markdown table formatting as for the other formats.

The new rendering can be seen at https://pr-1328.d16eby4ekpss5y.amplifyapp.com/reference/hooks/formats/predefined/#composeobject. Here's a screenshot:

Screenshot 2024-09-05 at 10 24 19

### DIFF
--- a/docs/src/content/docs/reference/Hooks/Formats/predefined.md
+++ b/docs/src/content/docs/reference/Hooks/Formats/predefined.md
@@ -559,15 +559,14 @@ Example:
 
 Creates a Kotlin file for Compose containing an object with a `val` for each property.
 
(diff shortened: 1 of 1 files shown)


jorenbroekema (approved):
Thanks, good catch :)!

dbanksdesign (approved):

PR #1330, opened 2024-09-05
docs: mention `options.className` for `ios-swift/*` formats

julien-deramond:
### Context

I was testing the `ios-swift/enum.swift` format and found out that there's a `options.className` available. However, it's not mentioned in the documentation (parameters table), and it's mentioned for Kotlin.

```json
{
  "source": [
    "tests/tokens.json"
  ],
  "platforms": {
    "ios-swift": {
      "transforms": ["attribute/cti", "name/camel", "content/swift/literal", "asset/swift/literal", "size/swift/remToCGFloat"],
      "files": [{
        "destination": "tests/build/TokensEnum.swift",
        "format": "ios-swift/enum.swift",
        "options": {
          "className": "StyleDictionaryEnum"
        }
      },{
        "destination": "tests/build/TokensClass.swift",
        "format": "ios-swift/class.swift",
        "options": {
          "className": "StyleDictionaryClass"
        }
      },{
        "destination": "tests/build/TokensAny.swift",
        "format": "ios-swift/any.swift",
        "options": {
          "className": "StyleDictionaryAny"
        }
      }]
    }
  }
}
```

### Description

So, this PR suggests adding `options.className` that isn't mentioned for `ios-swift/any.swift` and `ios-swift/enum.swift`:
- https://pr-1330.d16eby4ekpss5y.amplifyapp.com/reference/hooks/formats/predefined/#composeobject
- https://pr-1330.d16eby4ekpss5y.amplifyapp.com/reference/hooks/formats/predefined/#ios-swiftclassswift
- https://pr-1330.d16eby4ekpss5y.amplifyapp.com/reference/hooks/formats/predefined/#ios-swiftenumswift
- https://pr-1330.d16eby4ekpss5y.amplifyapp.com/reference/hooks/formats/predefined/#ios-swiftanyswift

On top of that, I suppose it must also be added in `lib/common/formats.js` (but not enough familiar with the repo yet to be 100% sure).

⚠️ Same thing here, only `ios-swift/any.swift` is mentioned in the migration Markdown file here. Maybe `ios-swift/enum.swift` and `ios-swift/class.swift` should be there too?

https://github.com/amzn/style-dictionary/blob/ed64beec9a1fffa5c6df5a89ab0e70123e1eca13/docs/src/content/docs/version-4/migration.md?plain=1#L528-L538

---

_Feel free to drop this PR if the quality is not enough, or if you had done it differently :) I'm not that familiar with the repository, so I might have modified the wrong files or missed some things_


jorenbroekema:
Hey, thanks for contributing, this looks good to me!

Can you add it to the migration docs as you mentioned? I think you're correct on that one.

Also, since you did a small type change in the source code, could you run `npx changeset` locally in the root of the project and create a changeset (patch) with a small description of the change, so it ends up nicely in the patch notes?

julien-deramond:
Thanks for the quick review, the required modifications are pushed.
Everything is in the same commit. If necessary, I can split it to separate the "doc" and the "lib" content.

jorenbroekema (approved):

.changeset/odd-icons-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+'style-dictionary': patch
```
dbanksdesign, line 2: Docs changes don't need a changeset, right @jorenbroekema ? We don't actually need to bump the npm package to change the docs

jorenbroekema: True but a small type adjustment was done inside the library code itself which does require a patch bump so users get the correct autocomplete for when they use those formats functions e.g. `StyleDictionary.hooks.formats['ios-swift/class.enum'].format({})` -> needs to autocomplete the options.className

dbanksdesign: Ah you are right, we should change the commit message on merge then to reflect that its not just a docs change

dbanksdesign (approved):

PR #1347, opened 2024-09-20
fix: empty tokens in filterTokenObject

mjmonline:
My tokens have a description key which often is empty which resulted in the following error

```javascript
const tokenValue = options.usesDtcg ? token.$value : token.value;
                                                               ^

TypeError: Cannot read properties of null (reading 'value')
    at file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:51:64
    at async file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:50:17
    at async filterTokenObject (file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:49:18)
    at async file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:65:24
    at async file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:50:17
    at async file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:50:17
    at async filterTokenObject (file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:49:18)
    at async file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:65:24
    at async filterTokenObject (file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:49:18)
    at async file:///project-dir/node_modules/style-dictionary/lib/filterTokens.js:65:24
```

### DIFF
--- a/lib/filterTokens.js
+++ b/lib/filterTokens.js
@@ -48,6 +48,9 @@ async function filterTokenObject(tokens, filter, options) {
   // out
   const result = await Object.entries(tokens ?? []).reduce(async (_acc, [key, token]) => {
     const acc = await _acc;
+    if (!token) {
+      return acc;
+    }
     const tokenValue = options.usesDtcg ? token.$value : token.value;
     // If the token is not an object, we don't know what it is. We return it as-is.
     if (!isPlainObject(token)) {


jorenbroekema:
Could you add a test for this so we can ensure not to create a regression for this bug in the future? 
Inside `__tests__/filterTokens.test.js`

mjmonline:
@jorenbroekema added one of my problematic tokens, the current tests fails without the fix so I didn't add any new test. I noticed that it only happens with my shadow tokens and not with any other type of token, I guess it's the structure of those tokens.

I recognise that [the plugin I use](https://github.com/lukasoppermann/design-tokens) could export them with a nicer format but nonetheless, no harm in this fix.

jorenbroekema:
closing this and proceeding with https://github.com/amzn/style-dictionary/pull/1354 to fix it

Issue #1305, opened 2024-08-14
`sortByReference` inverted (and incorrect) when using DTCG standard


chris-dura:
```javascript
// config.mjs

{
  "platforms": {
    "css": {
      "transformGroup": "css",
      "files": [
        {
          "destination": "vars.scss",
          "format": "scss/variables",
          "options": {
              "outputReferences": true
          }
        }
      ]
    }
  }
}
```

v3 format:

```json
{
  "colors": {
    "red": {
      "value": "#ff0000",
      "type": "color"
    },
    "primary": {
      "value": "{colors.red}",
      "type": "color"
    }
  }
}
```

Expected output, `$colors-red` is referenced after it is declared:

```scss
// Do not edit directly, this file was auto-generated.

$colors-red: #ff0000;
$colors-primary: $colors-red;
```

Change to DTCG format:

```json
{
  "colors": {
    "red": {
      "$value": "#ff0000",
      "$type": "color"
    },
    "primary": {
      "$value": "{colors.red}",
      "$type": "color"
    }
  }
}
```

Incorrect output, `$colors-red` is referenced before it is declared.

```scss
// tokens.scss output

// Do not edit directly, this file was auto-generated.

$colors-primary: $colors-red;
$colors-red: #ff0000;
```

jorenbroekema:
Oof that's not good, definitely seems like a bug, will take a look

lukasoppermann:
Hey, any news here? I may be running into the same issue.

Also this example suggest to use `dictionary`, but the type needs `TransformedTokens` so `dictionary.tokens`: https://github.com/amzn/style-dictionary/blob/main/examples/advanced/format-helpers/sd.config.js#L38

lukasoppermann:
@jorenbroekema do we just need to make sure to use `$value` if `usesDtcg` is `true` in link https://github.com/amzn/style-dictionary/blob/0cf8102afe49f51ba9e2b30e950a6fbc2dc11027/lib/common/formatHelpers/sortByReference.js#L57 and lines 59, 60, 64?

But it seems like the function does not have access to the `Config` object atm, right?

jorenbroekema:
@lukasoppermann yeah that definitely needs to use `original.$value` if `usesDtcg` is `true`.

[...]

PR #1349, opened 2024-09-24
fix(outputReferences): sorting of tokens in dtcg format

timges:
Fixes https://github.com/amzn/style-dictionary/issues/1305

**Issue Summary**:
The current implementation of the `sortByReference` function doesn't properly handle the DTCG token format, causing references in the output to be declared before their target. This occurs because the function assumes that token values can be accessed using the `value` attribute, which is not the case for the DTCG format, where the attribute is `$value`.

For example, this issue results in incorrect output like:
```scss
// Do not edit directly, this file was auto-generated.

$colors-primary: $colors-red;
$colors-red: #ff0000;
```
caused by:
```js
// -- sortByReference.js
function sortByReference(dictionary)
  function sorter(a, b) {
    // [...]
    // If token a uses a reference and token b doesn't, b might come before a
    // read on..
    if (a.original && dictionary.usesReference(a.original.value)) {
      // Both a and b have references, we need to see if the reference each other
      if (b.original && dictionary.usesReference(b.original.value)) {
        const aRefs = dictionary.getReferences(a.original.value);
        const bRefs = dictionary.getReferences(b.original.value);
    // [...]
  }
}
```

**Description of changes:**
This PR addresses the issue by modifying the `sortByReference` function to account for the `usesDtcg` options. It ensures that the correct token value is accessed based on the format (default or DTCG), preventing incorrect reference sorting.
```js
// -- sortByReference.js
// [...]
  const valueKey = usesDtcg ? '$value' : 'value';
// [...]
  const aUsesRefs = a.original && usesReferences(a.original[valueKey]);
  const bUsesRefs = b.original && usesReferences(b.original[valueKey]);
```

**Additional Changes:**
- Refactored the module to improve readability and follow clean code conventions.
- Inline comments where already used heavily, so I stuck to that practice and commented some lines.

**Checklist**
- [x] fix issue
- [x] add unit tests
- [x] add integration tests 
- [x] refactor code
 

### DIFF
--- a/__tests__/common/formatHelpers/sortByReference.test.js
+++ b/__tests__/common/formatHelpers/sortByReference.test.js
@@ -13,35 +13,24 @@
 import { expect } from 'chai';
 import sortByReference from '../../../lib/common/formatHelpers/sortByReference.js';
 
-const DEFAULT_TRANSFORMED_TOKENS = {
+const TRANSFORMED_TOKENS = (dtcg) => ({
   color: {
     primary: {
-      value: '#FF0000',
-      type: 'color',
-      original: { value: '{color.red}', type: 'color' },
+      [`${dtcg ? '$' : ''}value`]: '#FF0000',
+      [`${dtcg ? '$' : ''}type`]: 'color',
+      original: { 
+        [`${dtcg ? '$' : ''}value`]: '{color.red}',
+        [`${dtcg ? '$' : ''}type`]: 'color' },
     },
     red: {
-      value: '#FF0000',
-      type: 'color',
-      original: { value: '#FF0000', type: 'color' },
+      [`${dtcg ? '$' : ''}value`]: '#FF0000',
+      [`${dtcg ? '$' : ''}type`]: 'color',
+      original: { 
+        [`${dtcg ? '$' : ''}value`]: '#FF0000',
+        [`${dtcg ? '$' : ''}type`]: 'color' },
     },
   },
-};
-
-const DTCG_TRANSFORMED_TOKENS = {
-  color: {
-    primary: {
-      $value: '#FF0000',
-      $type: 'color',
-      original: { $value: '{color.red}', $type: 'color' },
-    },
-    red: {
-      $value: '#FF0000',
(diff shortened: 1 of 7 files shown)


jorenbroekema (approved):
Awesome, and thanks for cleaning the code up a bit as well, could you run `npx changeset` and follow the CLI instructions to add a human readable patch note for this change? I think a patch bump is fine since it's a bugfix

timges:
Done 👌

__tests__/common/formatHelpers/sortByReference.test.js
```diff
@@ -0,0 +1,112 @@
...
+      $type: 'color',
+      original: { $value: '{color.red}', $type: 'color' },
+    },
+    red: {
+      $value: '#FF0000',
+      $type: 'color',
+      original: { $value: '#FF0000', $type: 'color' },
+    },
+  },
+};
```
jorenbroekema, line 44:
```suggestion
const TRANSFORMED_TOKENS = (dtcg) => ({
  color: {
    primary: {
      [`${dtcg ? '$' : ''}value`]: '#FF0000',
      [`${dtcg ? '$' : ''}type`]: 'color',
      original: { 
        [`${dtcg ? '$' : ''}value`]: '{color.red}',
        [`${dtcg ? '$' : ''}type`]: 'color' },
    },
    red: {
      [`${dtcg ? '$' : ''}value`]: '#FF0000',
      [`${dtcg ? '$' : ''}type`]: 'color',
      original: { 
        [`${dtcg ? '$' : ''}value`]: '#FF0000',
        [`${dtcg ? '$' : ''}type`]: 'color' },
    },
  },
});
```

nitpick suggestion, could simplify this, up to you

jorenbroekema (dismissed):

__tests__/common/formatHelpers/sortByReference.test.js
timges: Sure, looks fine 👍

dbanksdesign (approved):

PR #1353, opened 2024-09-25
feat: add `markdown` and `markdown/table`

### LINKED ISSUE #1340: New Markdown (tables) format
### Context

I’ve been exploring Style Dictionary but haven't come across a built-in feature to generate Markdown files, nor issues or PRs mentioning it. I'm not familiar with all the features, so I've maybe missed something :)

While this might be a niche use case, I am currently in a situation where I need to generate Markdown tables from the JSON files. In addition to generating Android, iOS, and web files, I also need Markdown tables to reference in documentation.

After some experimentation, I was able to create a basic solution using the `registerFormat` method. Although it's a rough implementation and only tested for simple cases, it works as a proof of concept. Here's the snippet I used:

```mjs
  name: 'markdown',
  format: async ({ dictionary, options }) => {
    return (
      '| Name | Value | Type |\n' +
      '| ---- | ----- | ---- |\n' +
      dictionary.allTokens
        .map(function (token) {
          return `| ${token.name} | ${
            options.usesDtcg ? JSON.stringify(token.original.$value) : token.original.value
          } | ${token.$type} |`
        })
        .join('\n')
    )
  },
})
```

### Proposal

If there's currently no feature for generating Markdown tables output, would you be open to adding this as a new format? I’d be happy to refine the code further and submit a proper PR if there’s interest in supporting Markdown tables export as part of the project.

Let me know your thoughts, and I'd be glad to contribute!

julien-deramond:
> [!WARNING]  
> This PR is heavily draft.
> Feel free to:
> - close this PR if too heavily draft and that it bothers the repo to have it opened :)
> - directly modify it if needed

Closes #1340 

## Description

This PR proposes to add a new [`markdown/table` format](https://pr-1353.d16eby4ekpss5y.amplifyapp.com/reference/hooks/formats/predefined/#markdowntable) that would render a table in Markdown format. This would be useful for documentation purposes, as it would allow for a more structured and readable output.

For instance, it would generate the following Markdown:

```md
| Token | Description | Type | Value |
| --- | --- | --- | --- |
| aqua 0 |   | color | `#d9fcfb` |
| aqua 100 |   | color | `#c5f9f9` |
| aqua 200 |   | color | `#a5f2f2` |
| aqua 300 |   | color | `#76e5e2` |
| aqua 400 |   | color | `#33d6e2` |
| aqua 500 |   | color | `#17b8ce` |
| aqua 600 |   | color | `#0797ae` |
| aqua 700 |   | color | `#0b8599` |
| aqua 800 |   | color | `#0f6e84` |
| aqua 900 |   | color | `#035e73` |
| aqua 1000 |   | color | `#083d4f` |
| aqua 1100 |   | color | `#002838` |
```

that would be rendered this way:

| Token | Type | Value |
| --- | --- | --- |
| aqua 0 | color | `#d9fcfb` |
| aqua 100 | color | `#c5f9f9` |
| aqua 200 | color | `#a5f2f2` |
| aqua 300 | color | `#76e5e2` |
| aqua 400 | color | `#33d6e2` |
| aqua 500 | color | `#17b8ce` |
| aqua 600 | color | `#0797ae` |
| aqua 700 | color | `#0b8599` |
| aqua 800 | color | `#0f6e84` |
| aqua 900 | color | `#035e73` |
| aqua 1000 | color | `#083d4f` |
| aqua 1100 | color | `#002838` |

In terms of usage, even if I'm not that familiar with all usages of Style Dictionary, I can imagine that it could be useful to also have a new [`markdown` transform group](https://pr-1353.d16eby4ekpss5y.amplifyapp.com/reference/hooks/transform-groups/predefined/#markdown) that would allow rendering the tokens in Markdown format.

### `original` property

Mentioned in https://github.com/amzn/style-dictionary/issues/1340#issuecomment-2373798882:

> "original (reads from `token.original.value`/`token.original.$value`) → can be left empty if it's the same as `$value/value`"

- [ ] TODO: check how to implement it. Render a new dedicated column. If so, automatic or with a property named `showOriginalColumn` (or something like that).

### `description` property

Mentioned in https://github.com/amzn/style-dictionary/issues/1340#issuecomment-2373798882:

> "description (reads from `token.$description` or `token.comment` (legacy) property)"

This first version adds a `showDescriptionColumn` which makes it a voluntary gesture from the user. There could be another approach as described in https://github.com/amzn/style-dictionary/pull/1353#discussion_r1780921496 where there's nothing to configure as a user, and the description is automatically displayed if there's at least one description.

We could add some specific unit tests too.

### Reference chain (nice to have)

Mentioned in https://github.com/amzn/style-dictionary/issues/1340#issuecomment-2373798882:

> (nice to have) "reference chain, shows the entire chain of references that lead to the resolved value"

- [ ] TODO: Let's prototype! Maybe optional for this first version?

### Value preview (nice to have)

Mentioned in https://github.com/amzn/style-dictionary/issues/1340#issuecomment-2373798882:

> (nice to have) "value preview e.g. for colors putting a small color box div prefix, maybe some other token types can also show value previews"

Unfortunately, Markdown doesn't inherently support colored text (or areas). For example, it's not possible to include colored text in GitHub comments or descriptions, or even READMEs, even when using something like `<span style="color:orange;">Word up</span>`. However, some Markdown processors do allow for HTML.

However, as mentioned in https://github.com/amzn/style-dictionary/issues/1340#issuecomment-2376168925, GitHub and GitLab render colors in a specific way when the value is surrounded by backticks. This could be used to render a color preview in the table.

For instance: `#ffdd00`

Based on this, I've surrounded all the values (not necessarily colors) by backticks in the table. This would render the color values in a colored box in GitHub and GitLab, but not in other Markdown processors. And the other values will also have a code-like appearance. It would avoid complexity and specificities for each type of token.

**If we'd like to go further,** we might add a new option/property to provide the rendering as a parameter, or to enable a rich preview, where we could use a colored rounded `<div>` for instance. This would be useful for all Markdown preprocessors that don't display nicely the colors; for instance, the Markdown documentation of GitLab.

---

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -45,6 +45,7 @@ import scssMapDeep from './templates/scss/map-deep.template.js';
 import scssMapFlat from './templates/scss/map-flat.template.js';
 import macrosTemplate from './templates/ios/macros.template.js';
 import plistTemplate from './templates/ios/plist.template.js';
+import markdownTable from './templates/markdown/table.md.template.js';
 
 /**
  * @typedef {import('../../types/Format.ts').Format} Format
@@ -1531,6 +1532,50 @@ declare const ${moduleName}: ${JSON.stringify(treeWalker(dictionary.tokens), nul
       options,
     });
     return flutterClassDart({ allTokens: sortedTokens, file, options, formatProperty, header });
+  },
+
+  // Markdown templates
+  /**
+   * Creates a Markdown file containing a table with a row for each property.
+   *
+   * @memberof Formats
+   * @kind member
+   * @typedef {Object} markdownTableOpts
+   * @property {boolean} [markdownTableOpts.showFileHeader=true] - Whether or not to include a comment that has the build date
+   * @property {boolean} [markdownTableOpts.showDescriptionColumn=false] - Whether or not to show the description column in the table
+   * @property {OutputReferences} [markdownTableOpts.outputReferences=false] - Whether or not to keep [references](/#/formats?id=references-in-output-files) (a -> b -> c) in the output.
+   * @param {FormatArgs & { options?: markdownTableOpts }} options
+   * @example
+   * ```md
+   * | Token | Type | Value |
+   * | --- | --- | --- |
+   * | red 5 | color | `#fffaf3f2` |
+   * ```
+   */
+  'markdown/table': async function ({ dictionary, options, file }) {
+    const { allTokens, tokens, unfilteredTokens } = dictionary;
+    const { outputReferences, formatting, usesDtcg } = options;
+    const formatProperty = createPropertyFormatter({
+      outputReferences,
+      dictionary,
+      formatting,
+      usesDtcg,
+    });
+
+    let sortedTokens;
+    if (outputReferences) {
+      sortedTokens = [...allTokens].sort(sortByReference(tokens, { unfilteredTokens }));
+    } else {
+      sortedTokens = [...allTokens].sort(sortByName);
+    }
+
+    const header = await fileHeader({
+      file,
+      commentStyle: 'xml',
+      formatting: getFormattingCloneWithoutPrefix(formatting),
+      options,
+    });
+    return markdownTable({ allTokens: sortedTokens, options, formatProperty, header });
   },
 };
 
--- a/lib/common/templates/markdown/table.md.template.js
+++ b/lib/common/templates/markdown/table.md.template.js
@@ -1,0 +1,30 @@
+/**
+ * @typedef {import('../../../../types/DesignToken.ts').TransformedToken} TransformedToken
+ * @typedef {import('../../../../types/Config.ts').Config} Config
+ * @typedef {import('../../../../types/Config.ts').LocalOptions} LocalOptions
+ */
+
+/**
+ * @param {{
+ *   allTokens: TransformedToken[]
+ *   formatProperty: (token: TransformedToken) => string
+ *   options: Config & LocalOptions
+ *   header: string
+ * }} opts
+ */
+export default ({ allTokens, formatProperty, options, header }) => {
+  const hasDescription = options.showDescriptionColumn;
+
+  return `
+${header}
+
+| Token | ${hasDescription ? 'Description  | ' : ''}Type | Value |
+| --- | ${hasDescription ? '---  | ' : ''}--- | --- |
+${allTokens
+  .map(
+    (token) =>
+      `| ${token.name.replace(/ $/, '')} | ${hasDescription ? (token.$description ? token.$description : token.comment ? token.comment : '') + ' | ' : ''}${token.original.type} | \u0060${options.usesDtcg ? JSON.stringify(token.original.$value) : token.original.value}\u0060 |`,
+  )
+  .join('\n')}
+`;
+};
--- a/lib/common/transformGroups.js
+++ b/lib/common/transformGroups.js
@@ -322,4 +322,14 @@ export default {
    * @memberof TransformGroups
    */
   'react-native': ['name/camel', 'color/css', 'size/object'],
+
+  /**
+   * Transforms:
+   *
+   * [attribute/cti](/reference/hooks/transforms/predefined#attributecti)
+   * [name/human](/reference/hooks/transforms/predefined#namehuman)
+   *
+   * @memberof TransformGroups
+   */
+  markdown: ['attribute/cti', 'name/human'],
 };
--- a/__integration__/__snapshots__/customFormats.test.snap.js
+++ b/__integration__/__snapshots__/customFormats.test.snap.js
@@ -717,6 +717,10 @@ snapshots["integration custom formats inline custom with new args should match s
           "name/camel",
           "color/css",
           "size/object"
+        ],
+        "markdown": [
+          "attribute/cti",
+          "name/human"
         ]
       },
       "transforms": {
@@ -1671,6 +1675,10 @@ snapshots["integration custom formats register custom format with new args shoul
           "name/camel",
           "color/css",
           "size/object"
+        ],
+        "markdown": [
+          "attribute/cti",
+          "name/human"
         ]
       },
       "transforms": {
--- a/__integration__/__snapshots__/markdown.test.snap.js
+++ b/__integration__/__snapshots__/markdown.test.snap.js
@@ -1,0 +1,349 @@
+/* @web/test-runner snapshot v1 */
+export const snapshots = {};
+
+snapshots["integration markdown markdown/table should match snapshot"] = 
+`
+<!--
+  Do not edit directly, this file was auto-generated.
+-->
+
+| Token | Type | Value |
+| --- | --- | --- |
+| aqua 0 | color | \u0060#d9fcfb\u0060 |
+| aqua 100 | color | \u0060#c5f9f9\u0060 |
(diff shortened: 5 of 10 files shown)


julien-deramond:
Marking it as 'ready to review' to get initial feedback from the core team before proceeding further.

lib/common/templates/markdown/table.md.template.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @typedef {import('../../../../types/DesignToken.ts').TransformedToken} TransformedToken
+ * @typedef {import('../../../../types/Config.ts').Config} Config
+ * @typedef {import('../../../../types/Config.ts').LocalOptions} LocalOptions
+ */
+
+/**
+ * @param {{
+ *   allTokens: TransformedToken[]
+ *   formatProperty: (token: TransformedToken) => string
```
jorenbroekema, line 10: I don't think this is needed here, see other comment about not using `createPropertyFormatter`

lib/common/formats.js
```diff
@@ -1532,6 +1533,50 @@ declare const ${moduleName}: ${JSON.stringify(treeWalker(dictionary.tokens), nul
...
+   */
+  'markdown/table': async function ({ dictionary, options, file }) {
+    const { allTokens, tokens, unfilteredTokens } = dictionary;
+    const { outputReferences, formatting, usesDtcg } = options;
+    const formatProperty = createPropertyFormatter({
+      outputReferences,
+      dictionary,
+      formatting,
+      usesDtcg,
+    });
```
jorenbroekema, line 1563: I think this one is redundant here, the createPropertyFormatter is mostly useful as a utility to easily put the key/value pairs into a string property matching one of these formats: css, js, sass, styles, less etc.

In the case of the markdown template, there isn't much overlap between what that does and what this util does.

lib/common/templates/markdown/table.md.template.js
```diff
@@ -0,0 +1,30 @@
...
+ *   options: Config & LocalOptions
+ *   header: string
+ * }} opts
+ */
+export default ({ allTokens, formatProperty, options, header }) => {
+  const hasDescription = options.showDescriptionColumn;
```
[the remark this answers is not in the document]
jorenbroekema: Yeah this thought crossed my mind as well but I think it makes more sense to have the user explicitly choose whether or not they want the column, they can make that decision based on whether any tokens are using a description or not, they'll know whether it's useful as a column or redundant. I am mostly worried that removing the column because there are no descriptions while the user has explicitly configured that there should be a description column may be unexpected/counter-intuitive behaviour for them because that's a bit of invisible magic that goes against their configured thing.

lib/common/templates/markdown/table.md.template.js
```diff
@@ -0,0 +1,30 @@
...
+
+| Token | ${hasDescription ? 'Description  | ' : ''}Type | Value |
+| --- | ${hasDescription ? '---  | ' : ''}--- | --- |
+${allTokens
+  .map(
+    (token) =>
+      `| ${token.name.replace(/ $/, '')} | ${hasDescription ? (token.$description ? token.$description : token.comment ? token.comment : '') + ' | ' : ''}${token.original.type} | \u0060${options.usesDtcg ? JSON.stringify(token.original.$value) : token.original.value}\u0060 |`,
+  )
+  .join('\n')}
+`;
```
jorenbroekema, line 29: Disclaimer: the following code adjustment suggestion was written in Github PR WYSIWYG editor so I can't guarantee that it's valid 😅 

```suggestion
  return `${header.length > 0 ? `${header}\n\n` : ''}| Token | ${hasDescription ? 'Description  | ' : ''}Type | Value |
| --- | ${hasDescription ? '---  | ' : ''}--- | --- |
${allTokens
  .map(
    (token) =>
      `| ${token.name} | ${hasDescription ?
        (
          token.$description ?
            token.$description
            : token.comment ?
              token.comment 
              : ''
        ) + ' | '
      : ''}${options.outputReferences ?  : }${options.usesDtcg ? token.$type : token.type} | \`${JSON.stringify(options.outputReferences ? 
        (
          options.usesDtcg ?
            token.original.$value
            : token.original.value
        )
      : (
          options.usesDtcg ?
            token.$value
            : token.value
        )}\` |`,
  )
  .join('\n')}
`;
```

If we were to allow the columns option and granular control over which columns and what order, this format becomes a little bit more complex, code-wise, my code suggestion here is:
- we should not have leading newline
- we should not have redundant newlines when there is no file header
- we should support both DTCG and non DTCG formatted tokens
- fixed JSON.stringify to apply to both $value and value
- we should respect whatever token name convention the user configured rather than removing the first space (I'm not sure why it does that tbh)
- I think we should take the "type" rather than "original.type" unless we know a specific use case where the type gets changed (e.g. by a preprocessor) and we have a good reason to not use the transformed type over the original type. To be honest, I'm not even sure if the original.type gets set before the preprocessors lifecycle 😅 in that case it shouldn't be possible to have a different value for original.type vs type.
- we only output the original value and not the resolved value if outputReferences is false so I addressed this
- we should format that nested ternary a bit to make it more readable, or abstract it into a little function, on that note I'm curious if Prettier already released --experimental-nested-ternaries https://prettier.io/blog/2023/11/13/curious-ternaries, maybe with this flag it would auto-format the nested ternary. Feel free to add this flag in this PR if it works :).

For the no fileheader (empty string) use-case, I think we should add a test so we can verify by snapshot no redundant newlines are present. A(n integration) test for DTCG formatted tokens also makes sense imo.

When looking at my code suggestion and the fact that we may make it more complicated with the "columns" option and the ordering, I think I would prefer if we split up the formatting per token in functions (formatToken, formatValue, formatDescription), at least for the value and description this seems important for maintainability/readability. I'd probably abstract the table head part into its own function as well while you're at it.

lib/common/transformGroups.js
```diff
@@ -322,4 +322,14 @@ export default {
...
+
+  /**
+   * Transforms:
+   *
+   * [attribute/cti](/reference/hooks/transforms/predefined#attributecti)
+   * [name/human](/reference/hooks/transforms/predefined#namehuman)
+   *
+   * @memberof TransformGroups
+   */
+  markdown: ['attribute/cti', 'name/human'],
```
jorenbroekema, line 334: Same as my other comment, I don't think we need a transformGroup, first one seems redundant and the second one makes more sense to me if it matches the user's transformGroup's name transform (e.g. if they output to CSS they'll likely prefer kebab case for their markdown format so it somewhat matches with the tokens in code, they may even put a custom name transform kebab that also adds "--" prefix to be fully matching)

jorenbroekema (changes requested):
Sorry this took a while, I was pretty swamped with some other work so I only just now was able to really take a close look at this and invest time in properly reviewing.

I think we're definitely on the right track, but I do have quite a few suggestions :)

jorenbroekema:
Another note btw, for color values with color preview, this only works for issues, PRs and discussions, https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#supported-color-models

If we want color value previews that not only work in github markdown elsewhere but also for all markdown renderers, we may have to go with something more agnostic like this: https://stackoverflow.com/questions/11509830/how-to-add-color-to-githubs-readme-md-file (https://placehold.co/)

PR #1354, opened 2024-09-25
fix: filterTokens util to support random metadata props

jorenbroekema:
_Issue #, if available:_
https://github.com/amzn/style-dictionary/pull/1347

_Description of changes:_
Fix assumption that "token" is defined, where it may also be null or undefined or similar, which would throw if you attempt to access a property on it.

@mjmonline I created a PR based on yours and committed with you as co-author. I improved the test a bit more and also slightly refactored the offending code so it makes a bit more sense and is a bit more readable. Not that your PR wasn't valid or anything, it was good so thanks for contributing it :) but this cleans things up a little bit as well in addition to that.

### DIFF
--- a/.changeset/wicked-pears-eat.md
+++ b/.changeset/wicked-pears-eat.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': patch
+---
+
+Fix `filterTokens` utility to deal with random metadata properties throughout token groups, without throwing errors.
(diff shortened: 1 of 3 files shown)


dbanksdesign (approved):

mjmonline:
@jorenbroekema @dbanksdesign Too bad the co-authoring got lost in the merge :(

jorenbroekema:
Wow I've never seen that before, not sure why that happened :( there's another commit that was also merged by PR with co-author and that one works and is still in the commit history :|

mjmonline:
@jorenbroekema could be because of squash merge

PR #1362, opened 2024-10-07
fix(outputReferences): sorting of token without references

### LINKED ISSUE #1350: Alphabetical order and `outputReferences` usage
### Description

I'm not entirely familiar with all the features of Style Dictionary, so I might use it in the wrong way. I've observed some unexpected results around the generated alphabetical order when creating files for raw tokens.

I've tried to create a reproducible example of what I can observe by confronting the treatment to 3 different transform groups, and the usage (or not) of `outputReferences`.

Given the raw color tokens (available in the next section), I'd like to have as an output the following "smart" alphabetical order:

```
orange-10
orange-20
orange-100
orange-110
orange-200
```

The main issue I observe is that the following 3 configurations give different results:
- No `outputReferences`
- `outputReferences: true`
- `outputReferences: false`

I got the feeling that this option ideally shouldn't change the order of the elements in the end.

### Context

<details>
  <summary>Let's consider the following basic configuration files</summary>

`tokens.json`

```json
{
  "colors": {
    "$type": "color",
    "orange": {
      "10": {
        "$value": "#fff5eb"
      },
      "20": {
        "$value": "#fee2e2"
      },
      "100": {
        "$value": "#fffaf0"
      },
      "110": {
        "$value": "#fef3ec"
      },
      "200": {
        "$value": "#feebc8"
      }
    }
  }
}
```

`config.json`

```json
{
  "source": [
    "tests/tokens.json"
  ],
  "platforms": {
    "css": {
      "transformGroup": "css",
      "files": [
        {
          "destination": "tests/build/tokens.css",
          "format": "css/variables"
        }
      ]
    },
    "compose": {
      "transformGroup": "compose",
      "files": [
        {
          "destination": "tests/build/Tokens.kt",
          "format": "compose/object"
        }
      ]
    },
    "ios-swift": {
      "transformGroup": "ios-swift",
      "files": [{
        "destination": "tests/build/TokensAny.swift",
        "format": "ios-swift/any.swift"
      }]
    }
  }
}
```
</details>

### Tests

🟢 Perfect order
🟡 Pure alphabetical order
🟠 Inverted perfect order
🔴 Inverted pure alphabetical order

| `outputReferences`    | `css` + `css/variables` | `compose` + `compose/object` | `ios-swift` + `ios-swift/any.swift` |
| -------- | ------- | -------- | ------- |
| None |  🟢<br>--colors-orange-10<br>--colors-orange-20<br>--colors-orange-100<br>--colors-orange-110<br>--colors-orange-200 | 🟡<br>val colorsOrange10<br>val colorsOrange100<br>val colorsOrange110<br>val colorsOrange20<br>val colorsOrange200 | 🟡<br>public static let colorsOrange10<br>public static let colorsOrange100<br>public static let colorsOrange110<br>public static let colorsOrange20<br>public static let colorsOrange200 |
| `false` | 🟢<br>--colors-orange-10<br>--colors-orange-20<br>--colors-orange-100<br>--colors-orange-110<br>--colors-orange-200 | 🟡<br>val colorsOrange10<br>val colorsOrange100<br>val colorsOrange110<br>val colorsOrange20<br>val colorsOrange200 | 🟡<br>public static let colorsOrange10<br>public static let colorsOrange100<br>public static let colorsOrange110<br>public static let colorsOrange20<br>public static let colorsOrange200 |
| `true` | 🟠<br>--colors-orange-200<br>--colors-orange-110<br>--colors-orange-100<br>--colors-orange-20<br>--colors-orange-10 | 🟠<br>val colorsOrange200<br>val colorsOrange110<br>val colorsOrange100<br>val colorsOrange20<br>val colorsOrange10 | 🟠<br>public static let colorsOrange200<br>public static let colorsOrange110<br>public static let colorsOrange100<br>public static let colorsOrange20<br>public static let colorsOrange10 |

Based on these tests, we can see that:
* By default or with `outputReferences: false`, the output is the pure alphabetical order
* `css` + `css/variables` is great with a smart alphabetical order, but doesn't have the same output as the 2 others (Swift and Compose)
* When `outputReferences: true`, for all combinations:
  * It's an inverted order
  * However, it uses the "smart" alphabetical order (but inverted)

vprothais:
Fixes #1350 (at least for the outputReferences case)

**Issue Summary:**
When using the `outputReferences` option, the tokens output is in reversed order compared to the input order. This is due to the end of the `sortByReference` method which defines that A comes first by default.

Considering the following input :
```
orange-10
orange-20
orange-100
```
We end up with :
```css
--colors-orange-100
--colors-orange-20
--colors-orange-10
```

**Description of changes:**
At the end of the `sortByReference` method I introduced a `return 0` in order to keep original order for token without reference. I introduced a test for the case `bUsesRefs` to still have `A_COMES_FIRST` since it was the previous return value for this case.

I also added a unit test for the two tokens without reference case and had to update some snapshot because of the new order. I checked all generated snapshot and the token using a ref is always declared after the referenced token.

---


jorenbroekema (approved):
Aaaaaah so that's why! Good catch, thanks for contributing!

dbanksdesign (approved):

Issue #1365, opened 2024-10-11
options param is not populated when running preprocessor at platform level


bigglesrocks:
I have a custom preprocessor which I want to run only on a single platform, however I want access to the config via the `options` param that is passed to the `preprocess` function. If I specify this preprocessor at the root level, the `options` param is populated with the config as expected, however, specified on a single platform, the options object is empty. I would expect to get either the full config object or platform configuration, preferably both. 

```js
// config.js
{
  hooks: { 
   preprocessors: { 
    'my-custom-preprocessor': (tokens, options) => {
     console.log(options);
    }
   },
  // when specified at the root level, the console statement in the preprocessor function
  // will log out this config object as expected 👍
    preprocessors: ['my-custom-preprocessor'],  
    platforms: {
      css: { 
        files: [{ // basic css file }],
        transformGroup: 'css',
        // when specified on a platform, the console statement only logs an empty object: {} 😢
        preprocessors: ['my-custom-preprocessor'], 
      }
    }
  }
}
```

I've tested this in a barebones StyleDictionary project using v4.1.3. Also want to note that it makes no difference whether the preprocessor is defined inline like in the example or registered with `registerPreprocessor`.

PR #1370, opened 2024-10-16
fix: platform preprocessors retrieve platform config

tkgroot:
_Issue #, if available:_
closes https://github.com/amzn/style-dictionary/issues/1365

_Description of changes:_

As documented [applying preprocessors in config](https://styledictionary.com/reference/hooks/preprocessors/#applying-it-in-config), it shows that preprocessors can be defined for platforms specifically. However if placed as a platform specific preprocessor it does not include the configuration of the platform, as it does with "global" preprocessors.

- add platform preprocessors options param platform configuration

### DIFF
--- a/.changeset/gentle-lamps-add.md
+++ b/.changeset/gentle-lamps-add.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': patch
+---
+
+Pass PlatformConfig as options param to platform-applied preprocessors.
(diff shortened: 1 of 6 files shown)


jorenbroekema:
Added test, adjusted types and docs accordingly. Thanks for the PR!

jorenbroekema (approved):

dbanksdesign (approved):

Issue #1371, opened 2024-10-22
Types Resolve as `any` for TypeScript<5.x.x


zacowan:
## Description

When importing types from `style-dictionary/types`, TypeScript versions less than `5.x.x` cannot resolve what the types are supposed to be. As a result, the types resolve as `any`.

## To Reproduce

1. Initialize an empty node project
2. Install the previous major version of TypeScript

```
npm add -D typescript@4
```

3. Initialize the `tsconfig.json`

```
npm exec tsc -- --init
```

4. Install `style-dictionary@4.1.3`

```
npm add -D style-dictionary@4.1.3
```

5. Add a file with a type import

```ts
// src/index.ts
import type { TransformedToken } from 'style-dictionary/types';
```

6. Notice that `TransformedToken` resolves as `any`

## Additional Context

Node Version: 18.18.1

If you are using VSCode to inspect the resolved types, you'll need to adjust the settings to use the installed version of TypeScript, rather than the one shipped with VSCode.

```json
// .vscode/settings.json
{
  "typescript.tsdk": "node_modules/typescript/lib"
}
```

jorenbroekema:
You'll need to configure your tsconfig to handle package entrypoints:
```json
{
  "module": "NodeNext",
  "moduleResolution": "NodeNext"
}
```
for moduleResolution, "Bundler" also works in case you have relative extensionless imports and you don't want to migrate yet to using extensions

zacowan:
> You'll need to configure your tsconfig to handle package entrypoints:
> 
> ```json
> {
>   "module": "NodeNext",
>   "moduleResolution": "NodeNext"
> }
> ```
> 
> for moduleResolution, "Bundler" also works in case you have relative extensionless imports and you don't want to migrate yet to using extensions. https://www.typescriptlang.org/tsconfig/#moduleResolution

Thanks for the quick response @jorenbroekema! This does not work - the types still resolve as `any`. To clarify - TypeScript seems to be able to locate the file from the package entrypoints (`style-dictionary/types`), but it cannot then interpret the types that are exported from that file, resulting in all of the types being `any`. Additionally, `Bundler` is an option that is not supported on TS<5.x.x 😅 

This is reproducible with the same set of instructions above, with the addition on the `NodeNext` changes you've suggested.

jorenbroekema:
I'm addressing some of the extensions that style-dictionary is using internally https://github.com/amzn/style-dictionary/pull/1374
[...]

PR #1374, opened 2024-10-28
fix: use js extensions instead of ts in imports

jorenbroekema:
_Issue #, if available:_
closes https://github.com/amzn/style-dictionary/issues/1371

_Description of changes:_

Generally speaking TypeScript expects .js extensions for imports because the compiled output is .js. In the future TS might release a new feature allowing .ts extensions, but currently under NodeNext moduleResolution, .js is the way to go for files that actually compile to JS.

That said, in JSDocs annotated type-only imports we should use .d.ts extension because we only emit declaration files for these, so the compiled output is just .d.ts files.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,8 +3,6 @@
     "target": "ESNext",
     "module": "ESNext",
     "moduleResolution": "Bundler",
-    // Allow importing TypeScript files using their native extension (.ts(x)).
-    "allowImportingTsExtensions": true,
     // Enable JSON imports.
     "resolveJsonModule": true,
     // Enforce the usage of type-only imports when needed, which helps avoiding bundling issues.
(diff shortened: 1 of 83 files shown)


tsconfig.json
```diff
@@ -3,8 +3,6 @@
     "target": "ESNext",
     "module": "ESNext",
     "moduleResolution": "Bundler",
-    // Allow importing TypeScript files using their native extension (.ts(x)).
-    "allowImportingTsExtensions": true,
```
jorenbroekema, line 7: In latest typescript this only works with emitDeclarationOnly, but imo this flag is kinda tech debt anyways because it relies on bundler tool to resolve the import paths, and we actually want to stay bundleless.

zacowan:
This closes #1371 🥂

dbanksdesign (approved):

Issue #1367, opened 2024-10-14
Official Typescript Enums for Formats and Transforms?


DarioSoller:
I think it would be a favourable improvement, if you could centrally provide Typescript Enums for a few things, rather than having everyone writing their own. This would create a smoother DX for anyone using Style-Dictionary in a Typescript project, because a single source of truth (SSOT) for these things also make future update paths easier, cause Typescript errors can guide you for identifying breaking changes in the values for the formats or transfroms.

While I do acknowledge that enums are controversial in some scenarios, I think that they would really shine here for having more explicitly defined format and transform options. As enums can be used as a type but also as a constant, I see some unused benefits for within the Style-Dictionary code base itself and for anyone using Style-Dictionary in a Typescript project.

My proposed two enums that I created myself look like the following:
**sd-predefined-transforms.enum.ts**
```typescript
// extracted from https://github.com/amzn/style-dictionary/blob/main/lib/common/transforms.js
export enum SdPredefinedTransforms {
  attributeCti = 'attribute/cti',
  attributeColor = 'attribute/color',
  nameHuman = 'name/human',
  nameCamel = 'name/camel',
  nameKebab = 'name/kebab',
  nameSnake = 'name/snake',
  nameConstant = 'name/constant',
  namePascal = 'name/pascal',
  colorRgb = 'color/rgb',
  colorHsl = 'color/hsl',
  colorHsl4 = 'color/hsl-4',
  colorHex = 'color/hex',
  colorHex8 = 'color/hex8',
  colorHex8android = 'color/hex8android',
  colorComposeColor = 'color/composeColor',
  colorUIColor = 'color/UIColor',
  colorUIColorSwift = 'color/UIColorSwift',
  colorColorSwiftUI = 'color/ColorSwiftUI',
  colorCss = 'color/css',
  colorSketch = 'color/sketch',
  sizeSp = 'size/sp',
  sizeDp = 'size/dp',
  sizeObject = 'size/object',
  sizeRemToSp = 'size/remToSp',
  sizeRemToDp = 'size/remToDp',
  sizePx = 'size/px',
  sizeRem = 'size/rem',
  sizeRemToPt = 'size/remToPt',
  sizeComposeRemToSp = 'size/compose/remToSp',
  sizeComposeRemToDp = 'size/compose/remToDp',
  sizeComposeEm = 'size/compose/em',
  sizeSwiftRemToCGFloat = 'size/swift/remToCGFloat',
[...]

PR #1375, opened 2024-10-28
Integrating and providing Enum Objects

DarioSoller:
This PR proposes enum like JS objects for most hardcoded string configuration values. As discussed in the related issue #1367 .

A few remarks for the review of this PR:
- Besides the sheer amount of changes, no logic has been changed and it's solely replacing hard coded strings with their respective variable.
- Important consideration is, if and to what extend you want to promote these available enums? Especially regarding the appearance in the JSdoc code snippet examples!? For the sake of keeping these documentation code snippets simple, I can also revert the changes I have done there.
- Import paths for all examples should then be the final import path from the actual style-dictionary npm package. I will update that accordingly, as soon as I know, how you want it to be? Should it directly come from `style-dictionary` or maybe similar like the types `style-dictionary/types` something like `style-dictionary/enums`?
- Where in the project scaffolding are these enums best placed in your opinion? For now all enum JS objects live within one file `lib/buildInEnums.js`, but I can also make a folder with one file per enum object?
- I stumbled over one questionable file format (`'json/nested'`), which seems to be used internally, but should probably be not available as a "public" file format configuration option and therefore not part of the enum!? I am talking about https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L1538
- Surprised about the extend of necessary changes in one of the first commits for the `logging`, `commentStyle`, `commentPosition` have split up all subsequent changes in their separate commits. If wanted, I can interactively rebase and split up changes further, or I can also squash things accordingly, depending on how you want to see the changes in commits for the generated changelog later on?
- When is the best time to merge such a big change, without interfering with other open PRs and code changes, in order to minimize potential merge conflicts?


jorenbroekema:
- Sounds like a good idea to promote the enums
- `style-dictionary/enums` sounds good
- `lib/enums` with separate files `formats` / `transforms`, etc. sounds good
- `json/nested` is public and used by some users for sure so I would add it yeah
- Eventually we'll squash changes yeah, probably into 1 commit, usually 1 commit per changelog/change, sometimes that's many file changes if it means a refactor, that's ok
- Never haha, never a good time, but it's needed nevertheless and there aren't that many active PRs open apart from 2-3, I'll deal with the merge conflicts that happens for other open PRs

Will do a full review in a bit..

README.md
```diff
@@ -95,12 +95,12 @@ const StyleDictionary = require('style-dictionary').extend({
   source: ['tokens/**/*.json'],
   platforms: {
     scss: {
-      transformGroup: 'scss',
+      transformGroup: predefinedTransformGroupNames.scss,
```
jorenbroekema, line 98: I guess the cleanest API for our examples would be:

```js
import { transformGroups } from 'style-dictionary/enums';

const { scss } = transformGroups;

{
  transformGroup: scss
}

File: lib/buildInEnums.js [hunk 2]
```diff
@@ -0,0 +1,167 @@
...
+  inline: 'inline',
+};
+
+export const commentStylesFileHeader = {
+  short: 'short',
+  long: 'long',
+  xml: 'xml',
+};
+
+export const fileFormatNames = {
```
jorenbroekema, line 40:
```suggestion
export const formats = {
```

lib/buildInEnums.js
```diff
@@ -0,0 +1,167 @@
+export const predefinedActions = {
```
jorenbroekema, line 1:
```suggestion
export const actions = {
```

lib/buildInEnums.js
```diff
@@ -0,0 +1,167 @@
...
+  long: 'long',
+  none: 'none',
+};
+
+export const commentPositions = {
+  above: 'above',
+  inline: 'inline',
+};
+
+export const commentStylesFileHeader = {
```
jorenbroekema, line 34:
```suggestion
export const commentStylesFileHeaders = {
```

lib/buildInEnums.js
```diff
@@ -0,0 +1,167 @@
...
+  scssIcons: 'scss/icons',
+  scssMapDeep: 'scss/map-deep',
+  scssMapFlat: 'scss/map-flat',
+  scssVariables: 'scss/variables',
+  stylusVariables: 'stylus/variables',
+  typescriptEs6Declarations: 'typescript/es6-declarations',
+  typescriptModuleDeclarations: 'typescript/module-declarations',
+};
+
+export const predefinedTransformGroupNames = {
```
jorenbroekema, line 80:
```suggestion
export const transformGroups = {
```

lib/cleanFile.js
```diff
@@ -13,6 +13,7 @@
 
 import chalk from 'chalk';
 import { fs } from 'style-dictionary/fs';
+import { logVerbosityLevels } from './buildInEnums.js';
```
jorenbroekema, line 16:
```suggestion
import { logVerbosityLevels } from './buildInEnums.js';

const { silent } = logVerbosityLevels;
```

jorenbroekema (changes requested):
Just did a quick glance and in general I think we can:
- go for brevity for some of these enum names, so they are less wieldy to use/write for people
- destructure/dedupe repeated elements/enum members

And we'll need to add a small barrel file: `lib/enums/index.js` that we can point to in our package.json -> exports -> `"./enums": "./lib/enums/index.js"`

I'll give a more detailed thorough and detailed review eventually when the broader stuff is aligned a bit more

README.md
DarioSoller: I agree this would be very easy to read. Just keep in mind the overlap between `predefinedTransformGroupNames` and `predefinedPropertyFormatNames`. Partly the same string values (`css`, `less`) have been used for different properties.

lib/buildInEnums.js
DarioSoller: My thoughts on the naming have been, that it should not get confusing, if you ever have an array of format objects somewhere in the code? In order to actually refelect, what to expect behind this variable name, I came up with `fileFormatNames`. Downside, it's admittedly longer.

lib/buildInEnums.js
DarioSoller: Same here, what if you ever have an array of actual action objects in your code? But then this should actually also be called `predefinedActionNames` 🙈

lib/buildInEnums.js
DarioSoller: After your suggestion, I thougth about it. Maybe the most precise and best understandable name would be `fileHeaderCommentStyles`, when I think about it!?

lib/buildInEnums.js
DarioSoller: Same here, what if you ever have an array of actual transformGroup objects in your code?

lib/cleanFile.js
DarioSoller: Let's see, for what kind of API we also decide for `const { scss } = transformGroups;` and my concerns of duplicate values between different enums.

lib/buildInEnums.js
jorenbroekema: `import { actions as predefinedActions } from 'style-dictionary/enums';`

Exported stuff should always be somewhat brief if possible, users can quite easily make them more descriptive if they need it but the majority use case will be that brief = better and API design should be targeting majority use cases

lib/buildInEnums.js
jorenbroekema: Yea that's better indeed

README.md
jorenbroekema: I think that's fine, the string values can overlap among multiple enums, even within the same enum's members, but it allows us in the future (> v5, right now the values are public API) to change the values without it being breaking changes if the public API is only designated as the enum keys. Similarly, package entrypoints keys are public APIs whereas its values are private, meaning the locations of files corresponding to the entrypoints can be changed at any time by the maintainer, without that being breaking.

lib/cleanFile.js
jorenbroekema: Addressed similarly to overlapping imports w local variables:
`const { scss: scssGroup } = transformGroups;`

lib/buildInEnums.js
DarioSoller: Alright yeah fine with me. I personally would go with a more precise naming, but most importantly we have talked about it and we have a decision. I will change that also for your other suggested renamings, where we had a similar topic.

lib/buildInEnums.js
DarioSoller: see https://github.com/amzn/style-dictionary/pull/1375#discussion_r1819115164

lib/buildInEnums.js
DarioSoller: see https://github.com/amzn/style-dictionary/pull/1375#discussion_r1819115164

DarioSoller: I have renamed that enum not to `formats` but to `fileFormats` now, because there already have been multiple name conflicts with the important `formats` variable defined in `lib/common/formats.js`.

DarioSoller:
@jorenbroekema I added most of the discussed changes/updates except for the following, which is still a TODO:
- destructure/dedupe repeated elements/enum members
- Updating import paths in the examples to `style-dictionary/enums`

DarioSoller:
@jorenbroekema So I think now would be a good moment to make a more in depth review, if you have time. Following points should get a close inspection:
1. Destructering and depuping were only done, where it makes sense. So only at places with more than one occurrence for any of the enum values. This has been a lot of work, cause each file is kind of individual in regard of which enum values are needed. TBH I am not 100% convinced that this is the big advantage, because it all boils down to consistency vs. readability!? Anyhow it's done now and maybe it really was a one-time effort.
2. Certain `transform` values, that are either a specific example or some old legacy predefined transforms, like [here](https://github.com/DarioSoller/style-dictionary/blob/enums-objects/docs/src/content/docs/reference/api.md?plain=1#L627) or [here](https://github.com/DarioSoller/style-dictionary/blob/enums-objects/examples/advanced/transitive-transforms/sd.config.js#L74)
3. No changes done in `docs/src/content/docs/version-4/migration.md`, because these docs only show the migration to v4.0.0 right?
4. I had issues with a few search and replaces during the refactor, because it seems that backticks are also regarded as singleQuotes by Prettier!? Not sure about it, but there are a few string values in backticks, where actually no template string is needed, which I missed in my initial refactorings. I hope to have them all now!?

jorenbroekema:
I didn't end up getting to this yesterday/today due to other prios but I'll try to get on this Monday!

jorenbroekema (changes requested):
Looks good :)

Two things:
- really wanna call fileFormats formats to be honest, just because that's consistent with the name of the hook "formats"
- I think the enums members currently are loosely typed as just "string", in JSDocs it seems like from my experimentation the only way to strictly type the enum members is to cast every member to the exact string value. I was hoping @readonly or @enum tags in JSDocs would do it, especially @readonly I expected to work as enum but TypeScript doesn't agree with me :\ I pushed a commit to your branch just to show an example of how you would do this so that's three parts:
  - adjust the enums to be strict
  - remove typecasts in the JS source files where they are no longer needed
  - adjust the types TS files where needed to essentially use `(typeof <enum>)[keyof typeof <enum>]` to type the properties as "value of enum"

types/File.ts
```diff
@@ -16,8 +16,8 @@ export interface FormattingOptions extends FormattingOverrides {
 // to customize the output
 // Be careful with indentation if the output syntax is indentation-sensitive (e.g. python, yaml)
 export interface FormattingOverrides {
-  commentStyle?: 'short' | 'long' | 'none';
-  commentPosition?: 'above' | 'inline';
+  commentStyle?: 'short' | 'long' | 'none'; // refers to JS object in lib/enums/commentStyles.js
```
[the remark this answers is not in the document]
DarioSoller: @jorenbroekema Here I am still struggeling with how we could use the `transformTypes` enum for replacing these hardcoded string values for the different Transforms:
```typescript
// types/Transform.ts
export type NameTransform = BaseTransform<'name', string>;
export type AttributeTransform = BaseTransform<'attribute', Record<string, unknown>>;
export type ValueTransform = BaseTransform<'value', unknown | undefined>;
```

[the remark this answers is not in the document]
DarioSoller: Further, do you think there is anything for the `types/Format.ts` that we could an enum for, in order to be more precise with the typings? Not sure here, because custom formats and custom transforms always have to be possible and I am therefore uncertain if it makes any sense to type it something like:
```typescript
import { formats } from '../lib/enums/index.js';
type formats = typeof formats;

export interface Format {
  name: string | formats[keyof formats];
  format: FormatFn;
}
```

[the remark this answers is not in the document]
DarioSoller: And one more thing, which I noticed in [the docs of the format-helpers](https://github.com/amzn/style-dictionary/blob/main/docs/src/content/docs/reference/Utils/format-helpers.md). There are existing different `options.formatting` string values, which we have not covered yet as an enum?

DarioSoller:
Further, do you think there is anything for the `types/Format.ts` that we could use an enum for, in order to be more precise with the typings? Not sure here, because custom formats and custom transforms always have to be possible and I am therefore uncertain if it makes any sense to type it something like:
```typescript
import { formats } from '../lib/enums/index.js';
type formats = typeof formats;

export interface Format {
  name: string | formats[keyof formats];
  format: FormatFn;
}
```

DarioSoller:
And one more thing, which I noticed in [the docs of the format-helpers](https://github.com/amzn/style-dictionary/blob/main/docs/src/content/docs/reference/Utils/format-helpers.md). There are existing different `options.formatting` string values, which we have not covered yet as an enum?

DarioSoller:
I am not sure what causes the CI pipeline to fail atm!? Maybe someone can see more of the AWS Amplify logs than me?

types/File.ts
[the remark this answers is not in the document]
jorenbroekema:
```ts
import { transformTypes as transformTypesEnum } from '../lib/enums/transformTypes.js';

export type NameTransform = BaseTransform<typeof transformTypes.name, string>;
export type AttributeTransform = BaseTransform<typeof transformTypes.attribute, Record<string, unknown>>;
export type ValueTransform = BaseTransform<typeof transformTypes.value, unknown | undefined>;
```

Assuming the transformTypes have been strictly typed with the typecasts I described, if you then check e.g.:
```ts
type foo = NameTransform['type']; // typed as "name"
```

jorenbroekema:
> Further, do you think there is anything for the `types/Format.ts` that we could use an enum for, in order to be more precise with the typings? Not sure here, because custom formats and custom transforms always have to be possible and I am therefore uncertain if it makes any sense to type it something like:
> 
> ```ts
> import { formats } from '../lib/enums/index.js';
> type formats = typeof formats;
> 
> export interface Format {
>   name: string | formats[keyof formats];
>   format: FormatFn;
> }
> ```

Yeah that seems like a good idea for a bit better autocomplete if possible.

`options.formatting` is an object, each property can be any string value, so I think it's okay as is.

I'll take a look at the merge conflicts and the pipeline failure and commit a fix if needed.

DarioSoller:
@jorenbroekema If you like I can also rebase this branch to the current main and resolve the conflicts that way? Most of it should be in the import paths with the `.ts`to `.js` change. Just let me know, what the AWS Amplify logs say about the failing. I think then we could be done with this PR, what do you think?

jorenbroekema:
I'll take another look this weekend/Monday (probably) and get this PR over the finish line, including rebase etc. 👍🏻

jorenbroekema (approved):

jorenbroekema:
Alright I just squashed and rebased and fixed all of the stuff I encountered during thorough review, summary of what I adjusted, which can be viewed separately here https://github.com/amzn/style-dictionary/pull/1375/commits/1e2e088b6e39f03d2325e20efa662b2cbdeb3ac9 (note however that GitHub will eventually garbage collect this orphan commit):
- Some destructuring here and there (formats/transforms file)
- Missing enums that I ran into, e.g. json/asset format
- Add Enums section in docs to the sidebar and adjusted content ever so slightly for accuracy
- Made sure the newly added format uses enums as well (javascript/esm)
- Added a changeset entry for the changelog, minor bump seems appropriate since this is a pretty big feature
- Fixed a couple of test files that failed but didn't cause the runner to exit with an error code, this can happen if it fails to resolve imports in test files for example
- Fixed sd-playground / rollup bundle util to deal with multiple named SD imports, this is a bit complicated so don't worry about it too much, the enums can be used in the playgrounds now :)

dbanksdesign:
There are some broken links in the docs with this PR:

```
2024-11-12T11:06:01.436Z [WARNING]: 11:06:01 [ERROR] ✗ Found 4 invalid links in 2 files.
2024-11-12T11:06:01.436Z [INFO]: 11:06:01 ▶ reference/utils/format-helpers/
11:06:01 └─ /reference/enums#commentStyles - invalid hash
11:06:01 ▶ reference/api/
11:06:01 ├─ /reference/enums#logWarningLevels - invalid hash
11:06:01 ├─ /reference/enums#logVerbosityLevels - invalid hash
11:06:01 └─ /reference/enums#logWarningLevels - invalid hash
2024-11-12T11:06:01.639Z [WARNING]: [AstroUserError] Links validation failed.
```

DarioSoller:
@dbanksdesign thanks for sharing the Amplify error logs. Yeah, I see these page anchors need to be dasherized. I just pushed a fix. 🤞

DarioSoller:
@dbanksdesign as it is still failing, could you share the Amplify error logs one more time with me?

jorenbroekema (dismissed):
Fixed broken link and merge conflicts

FYI you can find broken links by running `npm run docs:build` locally which runs the Starlight Links validator, we may wanna make this part of the verify GH workflow to spot these early rather than via AWS console which not everyone can access.

dbanksdesign (dismissed):

dbanksdesign (approved):

Issue #1384, opened 2024-11-08
Add 'accessControl' field to the Android Compose template


xaviPastHubs:
It seems that the **accessControl** field is not controlled in the **Compose/Object** template (Android), unlike in ios-swift/any.swift for example. It could be added to change the visibility of the created objects, thereby improving the simplicity of development in apps that use it.

Fix [PR link here](https://github.com/amzn/style-dictionary/pull/1385)

PR #1385, opened 2024-11-08
feat: add accessControl field to Android Compose template

xaviPastHubs:
added accessControl to android compose template

It seems that the accessControl field is not controlled in the Compose/Object template (Android), unlike in ios-swift/any.swift for example. It could be added to change the visibility of the created objects, thereby improving the simplicity of development in apps that use it.

Issue [link here](https://github.com/amzn/style-dictionary/issues/1384)


jorenbroekema:
Awesome, can you add a changeset (patch I guess since this is a bugfix imo) by running `npx changeset` locally and committing the results of that?

I would also expect a test to now fail, or if there is no test for this format yet, could you add a test? `npm run test` locally, or `npm run test:watch` to go into debugging mode

xaviPastHubs:
@jorenbroekema I’ve followed the steps you provided and fixed the failing tests. I had to update the compose android snapshot as I also made a couple of formatting adjustments (semicolon after the package, a double space in the constructor, and a new line at the end of the file). Let me know if there’s anything else you need or if you have any further feedback!

jorenbroekema (approved):

jorenbroekema:
I added 2 more tests and some docs, let me know if this looks okay

jorenbroekema (approved):

xaviPastHubs:
@jorenbroekema Yes, the tests you've added look great! Appreciate the hard work you put in

PR #1386, opened 2024-11-09
chore: upgrade to eslint v9 and migrate to flatconfig

aliesenli:
**Description of changes:**

This PR upgrades ESLint to version 9 and [migrates to the new flat-config format](https://eslint.org/docs/latest/extend/plugin-migration-flat-config), which has been the default configuration format since ESLint v9.0.0. 

The `.eslintrc.json` file has been removed in favor of the new `eslint.config.js` file.

Following packages has been adjusted to make this migration possible:
```json
"@typescript-eslint/parser": "^8.13.0",  --> added parser to recognize react (built in)

"eslint": "^8.7.0",                      --> upgrade to "^9.14.0",
"eslint-config-react-app": "^7.0.1",     --> swapped with "eslint-plugin-react": "^7.37.2",
"eslint-plugin-mocha": "^10.2.0",        --> upgrade to "^10.5.0",
```

Patched the globals@11 to make linter happy  
patches/globals+11.12.0.patch


eslint.config.js
```diff
@@ -0,0 +1,45 @@
...
+import eslint from '@eslint/js';
+import typescriptParser from '@typescript-eslint/parser';
+import mochaPlugin from 'eslint-plugin-mocha';
+import reactPlugin from 'eslint-plugin-react';
+
+const GLOBALS_BROWSER_FIX = Object.assign({}, globals.browser, {
+  AudioWorkletGlobalScope: globals.browser['AudioWorkletGlobalScope '],
+});
+
+delete GLOBALS_BROWSER_FIX['AudioWorkletGlobalScope '];
```
aliesenli, line 11: Apparently following little bug has not been fixed and the `browser globals` is throwing an error without removing the extra trailing space: https://github.com/sindresorhus/globals/pull/184

eslint.config.js
```diff
@@ -0,0 +1,45 @@
...
+  { ignores: ['dist', 'docs'] },
+  {
+    languageOptions: {
+      ecmaVersion: 'latest',
+      sourceType: 'module',
+      parser: typescriptParser,
+      globals: {
+        ...globals.node,
+        ...globals.mocha,
+        ...GLOBALS_BROWSER_FIX,
```
aliesenli, line 24: Our dependency `@babel/plugin-transform-private-property-in-object` relies on a version of `globals` where the `AudioWorkletGlobalScope` property in `globals.browser` contains an extra trailing space, causing a linting error.

As a temporary fix, I addressed this by removing the extra space (GLOBALS_BROWSER_FIX).

Alternatively, we could consider overriding the globals package in `package.json`:

```json
"overrides": {
  "globals": "^15.12.0"
}
```

jorenbroekema (changes requested):
Thanks for the PR, I migrated a bunch of repos to ESLint v9 recently and contemplated doing it for Style Dictionary but couldn't find the time for it, so this is much appreciated.

Just 1 suggestion for patching globals@11 version

aliesenli:
> Thanks for the PR, I migrated a bunch of repos to ESLint v9 recently and contemplated doing it for Style Dictionary but couldn't find the time for it, so this is much appreciated.
> 
> Just 1 suggestion for patching globals@11 version

Thanks for the suggestion! I've just patched the globals@11 and rebased onto main.

jorenbroekema (approved):

jorenbroekema:
I think the package-lock version gets out of sync quite a bit because during the version bump the package-lock would have to be synced by doing an npm install after the bump in package.json and committing that with the version bump, I guess our release orchestration (changesets) isn't doing this. It's not a problem imo but could be solved by doing npm install after rebasing on latest main and committing the pkg lock version bump. Not a blocker for me though

dbanksdesign (approved):

PR #1392, opened 2024-11-15
feat(typescript/combine-json): add .ts file processing if runtime sup…

### LINKED ISSUE #1391: Feature Request: Typescript token files support.
When using typescript with

`deno`
`bun` 
or `node` with the `--experimental-strip-types` (https://github.com/nodejs/node/pull/53725)

It is not required to transpile typescript into javascript anymore.

With `style-dictionary` it is possible to use javascript to generate style rules. such as I do.

I use typescript to proof that all my themes are consistent and for stability reasons. 

`source: ["build/*.*js"]`
```
// src/dark.ts
import { light } from "../light/color.token.js";

// Let Typescript check that dark theme has the same values as light theme
type Dark = typeof light;

const dark: Dark = {...}
```
 
Unfortunately style dictionary cannot load my `.ts` files directly by using:

```
`source: ["src/*.*ts"]`
```

This would be much more convenient especially since all major runtimes do now support running typescript natively and i can run all my stuff without the intermediate step of transpiling everything.

hsjobeki:
As described in https://github.com/amzn/style-dictionary/issues/1391 many runtimes implement support for executing `.ts` files without any need for transpilation.

It seems the only problem is that `style-dictionary` tries to parse `.ts` files with `JSON5.parse` rather than just importing the file like the regular `.js` files.

I've tested the changes with:
- `deno` `v2.0.0`
- `node --experimental-strip-types`  `v22.9.0`
- `bun` `v1.1.29`

closes #1391 


lib/utils/combineJSON.js
```diff
@@ -99,19 +99,40 @@ export default async function combineJSON(
...
+          case '.js':
+          case '.mjs':
+          case '.ts': {
+            let resolvedPath = resolve(filePath, vol?.__custom_fs__);
+            // eslint-disable-next-line no-undef
+            if (isNode && process?.platform === 'win32') {
+              // Windows FS compatibility. If in browser, we use an FS shim which doesn't require this Windows workaround
+              resolvedPath = new URL(`file:///${resolvedPath}`).href;
+            }
+            file_content = (await import(/* @vite-ignore */ /* webpackIgnore: true */ resolvedPath))
```
jorenbroekema, line 112: Just a soft suggestion, perhaps it makes sense to warn the user that they need the experimental flag for NodeJS (or use Bun/Deno) if the import of .ts file fails.

lib/utils/combineJSON.js
```diff
@@ -99,19 +99,40 @@ export default async function combineJSON(
...
+            break;
+          }
+          default: {
+            // Use json parser fallback by default
+            // Other file types like .hjson, .topojson, or .ndjson, might be handled with dedicated parsers in the future.
+            // This warning is here to hint users that their file is parsed with the standard JSON5 parser and not any dedicated parser.
+            // Users should open an issue and provide the use case that caused the issue so we can investigate if we should allow users to inject custom parsers or add dedicated support.
+            console.warn(
+              `Unrecognized file extension: ${extname(filePath)}. Using JSON5 parser. If you see this open an issue on the Style Dictionary GitHub: https://github.com/amzn/style-dictionary/issues`,
+            );
```
jorenbroekema, line 131:
```suggestion
            console.warn(
              `Unrecognized file extension: ${extname(filePath)}. Using JSON5 parser as a default. Alternatively, create a custom parser to handle this filetype https://styledictionary.com/reference/hooks/parsers/`,
            );
```
I don't think it makes sense for them to create an issue unless there is a specific reason why they feel like Style Dictionary should add default support for it. Majority of cases I think will be that creating a custom parser is the way to go.

jorenbroekema (changes requested):
Few notes, but overall I like adding support for "native" TS token files.
- can you add a changeset with `npx changeset` locally, probably a minor bump for this
- I wonder if it would be possible to create a test for this feature that we only run in the CI workflow for node version which supports the strip types experimental flag, and run Node in the CI with that flag. The test would have to run conditionally based on whether the flag is active, maybe with some kind of environment variable

lib/utils/combineJSON.js
hsjobeki: Yeah had similar thoughts. I'll add that.

hsjobeki:
For reference here is the example log output for running `node`  without --experimental-strip-types

```bash
Could not import TypeScript file: ./tokens/token.ts

Executing typescript files during runtime is only possible via
- 'node >= 22.6.0' with '--experimental-strip-types'

Alternatively by using 'deno' or 'bun'

If you are not able to satisfy the above requirements, consider transpiling the TypeScript file to plain JavaScript before running the Style Dictionary build process.

Could not import TypeScript file: /home/johannes/git/style-dictionary/tokens/token.ts

Executing typescript files during runtime is only possible via
- 'node >= 22.6.0' with '--experimental-strip-types'

Alternatively by using 'deno' or 'bun'

If you are not able to satisfy the above requirements, consider transpiling the TypeScript file to plain JavaScript before running the Style Dictionary build process.

node:internal/modules/esm/get_format:218
  throw new ERR_UNKNOWN_FILE_EXTENSION(ext, filepath);
        ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Failed to load or parse JSON or JS Object: Failed to load or parse JSON or JS Object: Unknown file extension ".ts" for ./tokens/token.ts
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:218:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:244:36)
    at defaultLoad (node:internal/modules/esm/load:122:22)
    at async ModuleLoader.load (node:internal/modules/esm/loader:570:7)
    at async ModuleLoader.moduleProvider (node:internal/modules/esm/loader:443:45)
    at async ModuleJob._link (node:internal/modules/esm/module_job:106:19) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
Node.js v22.9.0
```

jorenbroekema (dismissed):

jorenbroekema (dismissed):
I started adding some more tests and realised we would probably want to support TS configs as well, we had some duplicate code going on. So I extracted the file loading stuff into its own util, with its own tests. It's much cleaner now I think.

I also made it a minor bump since this new feature is pretty significant.

@dbanksdesign needs your review since I now co-authored this PR instead of only reviewing.

hsjobeki:
Any updates on this?
I am using my own fork currently. Would be nice if this makes it into the next release.

dbanksdesign (approved):

PR #1399, opened 2024-11-28
Fix: fonts.css template doesn't correctly parse the `style` or `weight` options

LewisDaleUK:
Modify the fonts.css.template logic, so that it correctly coerces the null values when `font.style` and `font.weight` are undefined.

The current logic is using null coersion (the `??` operator), which means that the value is only written when `font.style` is null - which of course then throws an error because it tries to read `value` from an undefined object.

Then, if `font.style` and `font.weight` are defined, the output is `[object Object][object Object]`, because the actual statement is never reached.

_Issue #, if available:_

_Description of changes:_

Update the logic to use a ternary statement with empty strings as the fallbacks.

### DIFF
--- a/.changeset/two-islands-dance.md
+++ b/.changeset/two-islands-dance.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': patch
+---
+
+Fix font-style and font-weight logic for fonts.css.template.js
--- a/__tests__/formats/__snapshots__/fontsCSS.test.snap.js
+++ b/__tests__/formats/__snapshots__/fontsCSS.test.snap.js
@@ -1,0 +1,35 @@
+/* @web/test-runner snapshot v1 */
+export const snapshots = {};
+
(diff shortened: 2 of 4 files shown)


jorenbroekema:
Hey, this looks fine by me, could you maybe add a test and run `npx changeset` locally to create a human-readable changeset (patch bump is fine)?

LewisDaleUK:
> Hey, this looks fine by me, could you maybe add a test and run `npx changeset` locally to create a human-readable changeset (patch bump is fine)?

Added both now, let me know if they're okay/you need any more detail in them

jorenbroekema (approved):
I moved test file (we're a bit inconsistent about folder paths in lib versus tests) and used snapshots so it's easier to update in the future e.g. when we migrate to using tabs in v5 in all our formats, approved!

dbanksdesign (approved):

Issue #1409, opened 2024-12-10
`build` command with `--platform` option causes CLI to crash


juzerzarif:
If you run the `style-dictionary build` CLI command with the `--platform` option, the CLI crashes with the error:
```
file:///Users/my-project/node_modules/style-dictionary/bin/style-dictionary.js:129
    return Promise.all(options.platforms.map((platform) => sd.buildPlatform(platform)));
                                         ^

TypeError: Cannot read properties of undefined (reading 'map')
```

It looks like maybe a typo here - https://github.com/amzn/style-dictionary/blob/main/bin/style-dictionary.js#L129 ? `options.platforms` instead of `options.platform`?

jorenbroekema:
Fixing it here https://github.com/amzn/style-dictionary/pull/1416

PR #1416, opened 2024-12-17
fix: cli --platform flag

jorenbroekema:
_Issue #, if available:_
fixes https://github.com/amzn/style-dictionary/issues/1409

_Description of changes:_
Property name typo.
Above issue only occurred because we didn't have strict type checking enabled for the CLI, so I went ahead and added type checking to it, fixing a couple of minor issues in the process.


bin/style-dictionary.js
```diff
@@ -120,27 +142,33 @@ function getSD(configPath, options) {
...
+ * @param {BuildOptions} [options]
+ */
 async function styleDictionaryBuild(options) {
   options = options || {};
   const configPath = getConfigPath(options);
   const sd = getSD(configPath, options);
 
   if (options.platform && options.platform.length > 0) {
-    return Promise.all(options.platforms.map((platform) => sd.buildPlatform(platform)));
+    Promise.all(options.platform.map((platform) => sd.buildPlatform(platform)));
```
juzerzarif, line 154: Doesn't the promise returned by `Promise.all` here also need to be `await`ed? (same for `styleDictionaryClean`)

jorenbroekema: Yeah I don't think it matters coz this function is called by the CLI and then nothing else happens, the node process will end as soon as the stack is empty, but we should put an `await` just so it's less confusing to read

dbanksdesign (approved):

Issue #1420, opened 2024-12-19
javascript/es6 format doesn't handle newlines in comments


Zkrgu:
Comments are inserted as a line comment here: https://github.com/amzn/style-dictionary/blob/737a02b638b9450eb4f989fe9718357f5cba4239/lib/common/formats.js#L643

This has previously been fixed for some formats in #953 

Example of failing token

```json
{
    "token": {
        "value": 0,
        "comment": "This\nis not valid"
    },
    "another": {
        "value": 1,
    }
}
```

This fails to build with
```
SyntaxError: Missing semicolon. (6:3)
  4 |
  5 | export const token = 0; // This
> 6 | is not valid
  7 | export const another = 0;
  8 |
```

jorenbroekema:
I'm a bit conflicted about adding the enhancement flag, it's a missing feature imo but it could also be seen as a bug I suppose.

Anyways, the `createPropertyFormatter` format helper utility has a fix for this I introduced in v3 for the CSS / SCSS / LESS / Stylus formats, see the addComment helper. We can probably make that reusable in some way, and reuse it for our JS formats since they use a similar syntax for comments (`//` and `/* */`). Actually, I'm not sure `//` is valid in CSS, but at least the latter one they do share

curious-companion:
My approach will sanitize the comments by:
1. Escaping any occurrences of `*/` to prevent premature block closure using `.replace(/\*\//g, '*\\/')`.
2. Replacing newline characters (`\n`) with spaces to ensure the comment remains on a single line using `.replace(/\n/g, ' ')`.
Please tell me if there is issue with this approach!

PR #1431, opened 2025-01-18
fix the format which does not handle newline in the comment.

curious-companion:
_Issue #1420 

_Description of changes:_ My approach will sanitize the comments by:
1. Escaping any occurrences of `*/` to prevent premature block closure using `.replace(/\*\//g, '*\\/')`.
2. Replacing newline characters (`\n`) with spaces to ensure the comment remains on a single line using `.replace(/\n/g, ' ')`.
This ensures generated code is valid and comments are readable without formatting issues.

I'm open to suggestions if there's a preferred way to handle this. Looking forward to your guidance!

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -640,7 +640,12 @@ const formats = {
         const comment = options.usesDtcg ? token.$description : token.comment;
         const to_ret = `export const ${token.name} = ${value};`;
 
-        return comment ? to_ret.concat(`// ${comment}`) : to_ret;
+        if(comment){
+          //Sanitizing comment by escaping problematic comments.
+          const sanitizedComment = comment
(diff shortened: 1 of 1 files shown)


jorenbroekema:
I think we can reuse this utility here https://github.com/amzn/style-dictionary/blob/main/lib/common/formatHelpers/createPropertyFormatter.js#L48

It adds a comment to a token's output value (string), and takes care of multi-line comments and formatting preferences (e.g. `//` vs `/* */` preference). 

We should also add one smoke test for this format to see it working with multi-line comments imo: https://github.com/amzn/style-dictionary/blob/main/__tests__/formats/javascriptEsm.test.js  in this file probably

jorenbroekema:
Closing in favor of https://github.com/amzn/style-dictionary/pull/1439

Issue #362, opened 2019-12-05
Inconsistent support for alpha colors


blackfalcon:
In my JSON I have the following

```
      "ui-opacity-20": {
        "value": "0074C833",
        "comment": "base UI color @ 20%"
      },
```

The purpose of this is to apply alpha transparency using an 8 character hex. 

The CSS output looks like this

```
--color-ui-disabled-on-light: rgba(0, 116, 200, 0.2);
```

This is great. But, now I need to generate data to build documentation. When I output a JavaScript file, I get this

```
      "disabled": {
        "onLight": {
          "value": "#0074c8",
```

Here, instead of getting back the 8 character hex or the RGBA, I get a standard hex. 

To make matters worse, if I try to use an RGBA value ...

```
      "ui-opacity-20": {
        "value": "rgba(0, 116, 200, 0.2)",
        "comment": "base UI color @ 20%"
      },
```

The data still comes back with a standard hex

```
      "disabled": {
        "onLight": {
          "value": "#0074c8",
```

dbanksdesign:
I believe this has to do with different color transforms being applied, it looks like the javascript file is using the [color/hex](https://amzn.github.io/style-dictionary/#/transforms?id=colorhex) transform and the css file is using the [color/css](https://amzn.github.io/style-dictionary/#/transforms?id=colorcss) one. The color/hex one uses tinycolor's `toHexString()` method which looks to always return a 6-digit hex regardless of if the input color has transparency. The color/css does a bit of logic to see if the color has an transparency and uses different methods accordingly:

```
      var color = Color(prop.value);
      if (color.getAlpha() === 1) {
        return color.toHexString();
      } else {
        return color.toRgbString();
      }
```

There could be a few ways to fix this,

1. Update the 'js' transformGroup (and other transform groups that use color/hex) to use the color/hex8 transform instead of color/hex, then it will always return an 8-digit hex value. We could add a note that color/hex transform will always output a 6-digit hex regardless of transparency. The color/hex transform might still be useful for languages that don't understand 8-digit hex values
1. Update the color/hex transform to use the `toHex8String()` method and deprecate color/hex8 transform

Also, should we deprecate the color/css transform and just use color/hex8 if CSS understands 8-digit hex codes? That might make it easier all around.

blackfalcon:
I went down the route of updating the transformGroup, but really wanted to point out that there is an issue with expected default behavior.

chazzmoney:
@dbanksdesign lets clean this too

PR #1441, opened 2025-02-03
Fix: Fix colorHex to handle alpha values

laxa88:
_Issue #, if available:_
- https://github.com/amzn/style-dictionary/issues/362#issuecomment-562206735

_Description of changes:_

By default, when using the `js` transformGroup, all tokens are converted into hex values (for example, `#000000`) before additional transforms are applied. For `rgba` tokens, this causes the color values to lose the transparency value, resulting in all tokens to default to `#xxxxxxFF` (full opacity).

This PR fixes the default transform to use `hex8` to account for RGBA transforms.

I'm not entirely sure if there is a historical reason for using `colorHex` instead of `colorHex8` for `js` transforms, but if this is not an oversight, I would appreciate if someone can share some details. Otherwise, I hope this PR helps. 🙇 


jorenbroekema:
Hi, I think this is mostly an oversight to be honest, I actually would prefer if we didn't have 2 transforms in the first place and just have 1 that handles both hex8 but also uses hex6 when the alpha channel is set 1.
So if possible, can we adjust the regular hex transform to handle hex8 when necessary and deprecate the hex8 specific transform, so we can remove it in the next breaking change but start telling users to use the regular hex one instead?

And can we add some tests to the PR?

laxa88:
@jorenbroekema Thank you for the info and feedback.

- Reverted the change to `colorHex8`
- Updated the `colorHex` logic to handle hex6 and hex8, similar to `colorCss`
- Added and updated `colorHex` tests

jorenbroekema (approved):
Nice! thanks :)

I added a commit with some documentation / release note stuff to wrap this one up

dbanksdesign (approved):

PR #1454, opened 2025-02-19
Update filters.md

vricop:
Fix typo when inlining filter hooks, the type signature is not the same as the filter register config parameter.

_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/docs/src/content/docs/reference/Hooks/filters.md
+++ b/docs/src/content/docs/reference/Hooks/filters.md
@@ -51,7 +51,7 @@ StyleDictionary.registerFilter(myFilter);
 export default {
   hooks: {
     filters: {
-      'my-filter': myFilter,
+      'my-filter': myFilter.filter,
     },
   },
   // ... the rest of the configuration


DarioSoller:
I also stumbled over this just yesterday.

DarioSoller:
Relating to my linting hints, I don't think that this is caused by the name omission, but rather by the nested filter function. 
Bildschirmfoto 2025-03-12 um 01 11 54
Bildschirmfoto 2025-03-12 um 01 12 22
Bildschirmfoto 2025-03-12 um 01 12 38

jorenbroekema (approved):

jorenbroekema:
The inconsistency has to do with the fact that a filter only has 2 properties, a name and a filter function. Given that the name is the key in the inline hooks, we're only left with 1 property, so then we might as well only pass that prop instead of the whole object.

With transforms, there's other props like type, transitive, etc. 

I'm open to discussing to align this across all hooks to always have the full object even if it only has 1 property, for consistency reasons and to avoid confusion for consumers. This would be a breaking change though, for v6 milestone imo (we're trying to make v5 a drop-in replacement for 95% of users with only no/super-low -impact breaking changes)

DarioSoller:
#1469

dbanksdesign (approved):

PR #1460, opened 2025-02-28
Transform errors

jorenbroekema:
_Description of changes:_

Built-in transforms as well as custom transforms, generally make assumptions about the token value and what its type is.
For example, color tokens are assumed to be strings. Dimensions are assumed to be numbers or strings.
These assumptions aren't always correct because token authors can make mistake, as well as transform hook authors.

With this PR, we better handle errors as a result of transforming tokens. They are no longer fatal, and more information gets logged so it is easier to debug why and where a certain token transformation failed.


lib/transform/config.js
```diff
@@ -70,10 +70,11 @@ Unknown transformGroup "${to_ret.transformGroup}" found in platform "${platformN
   // the StyleDictionary module. We need to map the strings to
   // the actual functions.
   to_ret.transforms = transforms.map(function (name) {
```
jorenbroekema, line 72: why this change?

PR #1465, opened 2025-03-11
Refactor expand

jorenbroekema:
~~Definitely not ready yet, still fixing tests and expanding nested composite tokens.~~

This makes the expand util a LOT more performant, as it will create a clone of the Token Map and resolve references for any tokens that are pure refs and resolve to objects or have subproperties with pure refs that resolve to objects. Then, we can run the expand on the original token map but using the resolved map as a reference point to determine whether a token should be expanded or not and with what value.


jorenbroekema:
Okay I think I fixed expand by refactoring it to use the Map, and also added some small optimizations for resolving references for the purpose of token expansion, to avoid doing too much work (e.g. resolving refs that we know won't end up resolving to objects because they must resolve to strings)

lib/StyleDictionary.js
```diff
@@ -586,10 +593,10 @@ export default class StyleDictionary extends Register {
     }
 
     // TODO: When the transform / resolve have applied to the tokenMap instead of tokens, use the map and convert back to array / tokens obj
-    allTokens = convertTokenData(tokens, { output: 'array', usesDtcg: this.usesDtcg });
-    tokenMap = convertTokenData(allTokens, { output: 'map', usesDtcg: this.usesDtcg });
+    allTokens = convertTokenData(tokens, { output: 'array', usesDtcg });
+    tokenMap = convertTokenData(allTokens, { output: 'map', usesDtcg });
     // TODO: re-enable below when the Map is the source of truth while transforming/resolving
-    // tokens = convertTokenData(allTokens, { output: 'object', usesDtcg: this.usesDtcg });
```
jorenbroekema, line 592: Might need to remove these TODOs

Issue #1466, opened 2025-03-14
Update `addComment` to properly place comments when using `typescript/es6-declarations` + `ts/descriptionToComment` (`@tokens-studio/sd-transforms`)


jcfranco:
Moved from https://github.com/tokens-studio/sd-transforms/issues/349

---

> ### What happened?
> The comment placement looks out of place when generating output using the `typescript/es6-declarations` format with the `ts/descriptionToComment` transform.
> 
> Given the following token input:
> 
> ```json5
> {
>   "dimension": {
>     "scale": {
>       "value": "2",
>       "type": "sizing"
>     },
>     "xs": {
>       "value": "4px",
>       "type": "sizing",
>       "description": "xs comment"
>     },
>     "sm": {
>       "value": "{dimension.xs} * {dimension.scale}",
>       "type": "sizing",
>       "description": "sm comment"
>     },
>     // ...
>   }
> }
> ```
> the generated output:
> 
> ```ts
> export const dimensionScale: string; /** xs comment */
> export const dimensionXs: string; /** sm comment */
> // ...
> ```
> causes comments to shift to the wrong tokens, making it hard to scan when dealing with multiple comments.
> 
> ### Reproduction
> [SD configurator repro case](https://github.com/user-attachments/files/19243375/sd-transforms--incorrect-comment-placement-in-dts-output.zip)
> 
> ### Expected output
> ```ts
> export const dimensionScale: string; 
> export const dimensionXs: string; /** xs comment */
> export const dimensionSm: string; /** sm comment */
> // ...
> ```
> ### Version
> 1.2.9

From @jorenbroekema:

> Can you create an issue in the style-dictionary repo? The bug is because of the way the comment is added here [amzn/style-dictionary@`main`/lib/common/formats.js#L751](https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L751)
> 
> It should be using the `addComment` helper [amzn/style-dictionary@`main`/lib/common/formats.js#L654](https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L654) This one is a bit smarter about creating multi-line comments (always uses long & above) when needed and letting the user configure the commentStyle (short vs long) and commentPosition (above vs inline) for single-line comments
> 
> PR to Style Dictionary very welcome :) should be a pretty easy one

jcfranco:
[...]

PR #950, opened 2023-03-15
feat(formats) Use TSDoc comment format for typescript/es6-declarations

christianvuerings:
*Issue #, if available:*

*Description of changes:* Use [TSDoc](https://tsdoc.org/) comment format for `typescript/es6-declarations`

*Why:* So developers can see the comment on hover in their editor (e.g. VSCode)

*Before:*
Screenshot 2023-03-15 at 3 49 30 PM

*After:*
image

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1473, opened 2025-03-18
fix(formats): fix misaligned comments in `typescript/es6-declarations`

jcfranco:
**Related Issue:** #1466

## Summary

Ensures `addComment` places TS declaration file comments in JSDoc format above their respective targets for proper annotation, aligning with TS-emitted declaration files.

### Notes

* This updates all long comments to start with `/**`
* enforces `long` + `above` formatting options in `typescript/es6-declarations` format to ensure JSDoc syntax

### DIFF
--- a/__tests__/formats/__snapshots__/typeScriptEs6Declarations.test.snap.js
+++ b/__tests__/formats/__snapshots__/typeScriptEs6Declarations.test.snap.js
@@ -6,6 +6,7 @@ snapshots["formats typescript/es6-declarations with outputStringLiterals should 
  * Do not edit directly, this file was auto-generated.
  */
 
+export const colorBlue: "#0000FF";
 /** Used for errors */
 export const colorRed: "#FF0000";
 export const fontFamily: '"Source Sans Pro", Arial, sans-serif';
@@ -17,6 +18,7 @@ snapshots["formats typescript/es6-declarations without outputStringLiterals shou
  * Do not edit directly, this file was auto-generated.
  */
 
+export const colorBlue: string;
 /** Used for errors */
 export const colorRed: string;
 export const fontFamily: string;
@@ -28,6 +30,7 @@ snapshots["formats typescript/es6-declarations with DTCG tokens and outputString
  * Do not edit directly, this file was auto-generated.
  */
 
+export const colorBlue: "#0000FF";
 /** Used for errors */
 export const colorRed: "#FF0000";
 export const fontFamily: '"Source Sans Pro", Arial, sans-serif';
@@ -39,9 +42,21 @@ snapshots["formats typescript/es6-declarations with DTCG tokens and without outp
  * Do not edit directly, this file was auto-generated.
  */
 
+export const colorBlue: string;
 /** Used for errors */
 export const colorRed: string;
 export const fontFamily: string;
 `;
 /* end snapshot formats typescript/es6-declarations with DTCG tokens and without outputStringLiterals should match snapshot */
+
+snapshots["formats typescript/es6-declarations with formatted comment should match snapshot"] = 
+`/**
+ * Do not edit directly, this file was auto-generated.
+ */
+
+export const colorBlue: string; /** Used for errors */
+export const colorRed: string;
+export const fontFamily: string;
+`;
+/* end snapshot formats typescript/es6-declarations with formatted comment should match snapshot */
 
--- a/__tests__/formats/__snapshots__/typeScriptEs6Declarations.test.snap.js
+++ b/__tests__/formats/__snapshots__/typeScriptEs6Declarations.test.snap.js
@@ -49,14 +49,3 @@ export const fontFamily: string;
 `;
 /* end snapshot formats typescript/es6-declarations with DTCG tokens and without outputStringLiterals should match snapshot */
 
-snapshots["formats typescript/es6-declarations with formatted comment should match snapshot"] = 
-`/**
- * Do not edit directly, this file was auto-generated.
- */
-
-export const colorBlue: string; /** Used for errors */
-export const colorRed: string;
-export const fontFamily: string;
-`;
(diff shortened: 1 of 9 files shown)


__tests__/formats/__snapshots__/typeScriptEs6Declarations.test.snap.js
```diff
@@ -39,9 +42,21 @@ snapshots["formats typescript/es6-declarations with DTCG tokens and without outp
...
 `;
 /* end snapshot formats typescript/es6-declarations with DTCG tokens and without outputStringLiterals should match snapshot */
 
+snapshots["formats typescript/es6-declarations with formatted comment should match snapshot"] = 
+`/**
+ * Do not edit directly, this file was auto-generated.
+ */
+
+export const colorBlue: string; /** Used for errors */
+export const colorRed: string;
```
jorenbroekema, line 58:
```suggestion
export const colorBlue: string;
/** Used for errors */
export const colorRed: string;
```

this would be my expectation, the comment seems to be attached to the red token, not the blue, and in the format it says long / above are enforced to match typescript declaration emission behavior.

The code looks okay, maybe this snapshot just needs to be updated with `npm run test:update-snapshots`

__tests__/formats/typeScriptEs6Declarations.test.js
```diff
@@ -13,7 +13,7 @@
 import { expect } from 'chai';
 import formats from '../../lib/common/formats.js';
 import createFormatArgs from '../../lib/utils/createFormatArgs.js';
-import { convertTokenData } from '../../lib/utils/convertTokenData.js';
+import { convertTokenData } from '../../lib/utils/index.js';
```
jorenbroekema, line 16: Revert, better to import from the file directly and avoid the barrel file.

lib/common/formats.js
```diff
@@ -746,15 +748,21 @@ const formats = {
...
+          indentation: '',
+          ...formatting,
 
-        return [
-          `${comment ? `/** ${comment} */` : ''}`,
-          `export const ${token.name} : ${typescriptType};`,
-        ].join('\n');
+          // position and style are enforced to match TypeScript declaration file emitting
+          commentPosition: above,
+          commentStyle: long,
```
jorenbroekema, line 758:
```js
{
  indentation: '',
  commentPosition: 'above',
  commentStyle: '',
  ...formatting,
}
```
I prefer the above, above/long are good defaults because it matches declaration file emitting, but user should still be able to override that, since it's not incorrect/invalid syntax to do so. "above" and "long" is already enforced for multi-line comments

jorenbroekema (changes requested):

lib/common/formats.js
jcfranco: Would it be acceptable to treat that as a follow-up? IMO, this is slightly beyond the scope of fixing the comment alignment, especially when the current behavior doesn’t allow overriding formatting options either.

FWIW, I do have a concern about users unintentionally opting out of JSDoc behavior if they don’t use `above` + `long`. 

https://github.com/amzn/style-dictionary/pull/950 enforced a specific comment format to ensure JSDoc comments are picked up correctly. From what I understand, JSDoc requires comments to be placed directly above their target. I can’t find a spec, but every example on the [JSDoc site](https://jsdoc.app/) follows that pattern. So when a user switches to `inline` and/or `short`, the behavior of the comment will change.

I think part of the confusion might come from treating JSDoc and multiline comments as synonymous when they serve different purposes. Maybe this distinction could be clarified by:

1. Disallowing customizing formatting options for now – this preserves current behavior and still resolves the misalignment issue  
2. Creating a follow-up issue to enhance `typescript/es6-declarations` to honor formatting options, where it will:  
   1. Stop using JSDoc-style comments by default to align with other formats  
   2. Introduce a separate `jsdoc` (or similar) JS-only format that explicitly generates JSDoc comments, with enforced options to ensure the output works as expected (this could help other JS formats too)  
   3. Use the new format as the default for `typescript/es6-declarations`, which can be overridden

LMK how you'd like me to proceed.

jorenbroekema: > I think part of the confusion might come from treating JSDoc and multiline comments as synonymous when they serve different purposes

While I agree with this, the design philosophy of Style Dictionary is to be agnostic where possible. This is a trade-off where we prefer flexibility over constraints, even if that flexibility means users may make mistakes.

I think it's not up to us to enforce correct JSDocs syntax but rather provide defaults that are correct, but allow users to override when needed. It's up to the user to decide if they want to break JSDocs syntax, e.g. because they're not using JSDocs and they prefer a different commentStyle for whatever reason.

I think I will insist on including this in this PR's scope, if we override the user-provided formatting settings now and then later we give the user-provided formatting settings precedence over our defaults, that change would be slightly breaking ("slightly" meaning: it's breaking in the sense that the token output changes after a Style Dictionary non-breaking version bump, which is okay in some cases but we avoid it where possible to better respect SemVer)

jorenbroekema:
Just the one comment that needs resolving about user-preferred settings getting precedence over the JSDocs-based default values. While I would never recommend overriding the defaults, it's aligned with Style Dictionary's design philosophy to allow users that flexibility.

Looks good to me other than that, so I'll approve when the above the addressed.

lib/common/formats.js
jcfranco: Fair enough. Thanks again for reviewing and the reasoning breakdown. 💪 I'll make the suggested changes!

jcfranco:
@jorenbroekema Let me know if anything else is needed on this. Thanks!

jorenbroekema (approved):

jorenbroekema:
Approved, just have to check with @dbanksdesign and then we can merge/release it. I'll port this to v5 branch when that's ready for 5.0.0 release.

dbanksdesign (approved):

PR #1474, opened 2025-03-19
feat(formats/javascriptEsm): add `flat` option

### LINKED ISSUE #1451: [Format Request]: Javascript esm flattern format request
Hi, 

I know there is [javascriptModuleFlat](https://github.com/amzn/style-dictionary/blob/main/lib/common/formats.js#L483) format, wondering if there is a similar one but with esm export? basically, it's replacing `module.exports` with `export default`. I tried to create my custom formatter, however the `formatJS` is not exported as part of the util, recreating it might be reinventing the wheel. If having esm flat format is not possible would you guys consider exporting the `formatJS()` function?

I think this would be a great addition.
Thanks

tiloyi:
_Issue #, if available:_
Resolves #1451

_Description of changes:_
This PR adds the `flat` option to the `javascriptEsm` format, which gives flat output, as discussed in the associated issue.

_Additional comment:_
@jorenbroekema in the issue you also mention adding an `export` option, but if that's OK with you, I'd prefer to do it in a separate PR, and not mix these features to keep the commits/changelog clearer. What do you think?

---


docs/src/content/docs/reference/Hooks/Formats/predefined.md
```diff
@@ -312,6 +312,7 @@ Creates an ES6 module object of the style dictionary.
...
 | `options`                 | `Object`            |                                                                                                                                                                                                                                                                                                                               |
 | `options.minify`          | `boolean`           | Whether or not to minify the output. Defaults to `false`.                                                                                                                                                                                                                                                                     |
+| `options.flat`            | `boolean`           | Whether or not to flatten the output. Defaults to `false`. _(Cannot be combined with ‘minify' option)._                                                                                                                                                                                                                       |
```
jorenbroekema, line 315:
```suggestion
| `options.minify`          | `boolean`           | Whether or not to minify the output. Defaults to `false`. Has no effect when `options.flat` is `true`.                                                                                                                                                                                                                                                                    |
| `options.flat`            | `boolean`           | Whether or not to flatten the output. Defaults to `false`. If `true`, renders only the token value, which means it is minified.                                                                                                                                                                                                                   |
```

This is a bit clearer, you can combine them but it just won't have an affect if flat is true. Saying you cannot combine them implies there would be some kind of error/issue if you did.

lib/common/formats.js
```diff
@@ -689,11 +689,25 @@ const formats = {
...
+      dictionaryString =
+        '{\n' +
+        dictionary.allTokens
+          .map(function (token) {
+            return `  "${token.name}": ${JSON.stringify(
+              options.usesDtcg ? token.$value : token.value,
+            )}`;
+          })
+          .join(',\n') +
+        '\n}';
```
jorenbroekema, line 703:
```suggestion
      dictionaryString = JSON.stringify(
        Object.fromEntries(
          dictionary.allTokens.map(token => [
            token.name,
            options.usesDtcg ? token.$value : token.value
          ]),
        ),
        null,
        2
      );
```

Cleaner and more optimized to create the object from the array and stringify that in its entirety rather than stringifying each token value separately and string-building the JSON ourselves. You'll have to format this suggestion though, linting will complain if you commit the suggestion as is.

jorenbroekema (changes requested):
Looks good for the most part, only 2 comments but the 2nd one is a blocker because it can have a significant negative effect on performance for very large tokensets.

docs/src/content/docs/reference/Hooks/Formats/predefined.md
tiloyi: Fixed.

lib/common/formats.js
tiloyi: Fixed.

jorenbroekema (dismissed):

jorenbroekema:
@tiloyi could you add a changeset with `npx changeset` ? Then pick "minor" and add a human readable description of this new feature. Looks good otherwise

tiloyi:
@jorenbroekema it's done ✅

jorenbroekema (approved):

dbanksdesign (approved):

Issue #1485, opened 2025-04-07
[Proposal] support selector option as string array in css/variables formatter


YumiChen:
## Use case
We're using style-dictionary to generate multiple CSS stylesheets containing CSS variables. We'd like the variable values to be overridden at different device breakpoints and in light/ dark mode, which would require the variables to be nested in a media query. Currently this is not achievable with `css/variables` formatter `selector` option:
```css
/* selector as @media (max-width: 768px) */
/* Invalid CSS, a selector should be nested in the media query */
@media (max-width: 768px) {
    --color-primary: white; 
}  

/* selector as @media (max-width: 768px) :root { */
/* Invalid CSS, a closing curly bracking is missing at the end */
@media (max-width: 768px) :root { {
    --color-primary: white; 
}  
```

Expected output
```css
@media (max-width: 768px) {
  :root { 
    --color-primary: white; 
  }
}
```

This may also be useful for other queries like `@supports` and `@container`

## Proposal
Having the selector option support both `string` and `string[]`, so that the selectors could be nested
```js
{
  "platforms": {
    "css": {
      "files": [{
        "format": "css/variables",
        "selector": ["@media (max-width: 768px)", ":root"] // supports array
      }]
    }
  }
}
```

I'll create a PR draft for better reference

jorenbroekema:
I like it, concept has my approval, will check your draft PR now

PR #1486, opened 2025-04-07
feat: support selector option as string or string array in css/variables formatter

YumiChen:
_Issue #, if available:_
Resolves https://github.com/amzn/style-dictionary/issues/1485

_Description of changes:_
Support `selector` option as s string or string array. When a string array is provided, the variables would be nested in the selectors, with the first selector in the array being the outermost


lib/common/formats.js
```diff
@@ -198,25 +198,47 @@ const formats = {
...
-        formatting,
-        usesDtcg,
-      }) +
-      `\n}\n`
+      selector.reverse().reduce(
+        (
+          /** @type {string} */ result,
+          /** @type {string} */ _selector,
+          /** @type {number} */ index,
+        ) => wrapVariables(_selector, result, indentation.repeat(selector.length - 1 - index)),
```
jorenbroekema, line 228: This is hard to read because we're swapping the order of the reduce "accumulated" and "current" args in the wrapVariables function parameters. I don't mind using Array.reduce() here but let's see if this can be written a bit better so it's easier to understand what it does. "variables" param in the wrapVariables is also a bit misleading because it could also be the variables + already a selector around it, if the selector length > 1 and we're calling it for the 2nd (or more) time.

jorenbroekema (changes requested):
Looks good overall, just a few comments

lib/common/formats.js
```diff
@@ -198,25 +198,47 @@ const formats = {
...
     const { outputReferences, outputReferenceFallbacks, usesDtcg, formatting } = options;
     const header = await fileHeader({
       file,
       formatting: getFormattingCloneWithoutPrefix(formatting),
       options,
     });
+
+    const indentation = formatting?.indentation || '  ';
+    const wrapVariables = (
+      /** @type {string | string[]} */ selector,
```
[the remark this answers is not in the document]
YumiChen: Thanks for catching this. Updated.

lib/common/formats.js
YumiChen: @jorenbroekema Thank you for the feedback. I've renamed variables/ functions and changed the order of parameters at https://github.com/amzn/style-dictionary/pull/1486/commits/6a0bd898ba188ab802326923860d2090ba34eb6e

YumiChen: Given the typically short length of `option.selector`, I think an alternative solution would be recursion, with which it would be like as below:
```javascript
...
const nestInSelector = (
      /** @type {string[]} */ selector,
      /** @type {string} */ content,
    ) => {
      if (!selector.length) return content; 
  
      const [current, ...remaining] = selector;
      const indent = indentation.repeat(selector.length - 1);
      const wrapped = `${indent}${current} {\n${content}\n${indent}}`;
  
      return remaining.length 
        ? nestInSelector(remaining, wrapped, indentation)
        : wrapped;
    };
...
    return (
      header +
      nestInSelector(selector.reverse(), variables) +
      '\n'
    );
```

Both solutions seem readable to me. Please let me know if this solution is more desired

jorenbroekema (approved):

YumiChen:
> Could you add a changeset (minor bump) using `npx changeset` locally in the root project folder? To add a human readable changelog entry (aka patch note) for our CHANGELOG.md

Sure. Thanks for the review. Just committed the changeset file. Please let me know if there are any questions

jorenbroekema (approved):

dbanksdesign (approved):

jorenbroekema:
@YumiChen everything's approved but the performance test is failing. This could be CI flakiness (just being slower than usual), but given the nature of the changes, I do kinda wanna double check the performance metrics locally to see if there's any significant differences before/after the changes and verify that the CI failure is a false-positive.

YumiChen:
> @YumiChen everything's approved but the performance test is failing. This could be CI flakiness (just being slower than usual), but given the nature of the changes, I do kinda wanna double check the performance metrics locally to see if there's any significant differences before/after the changes and verify that the CI failure is a false-positive.

No problem. I'll try running the performance tests locally and compare the average execution time before and after the changes in this MR to check for any performance differences.

PR #1489, opened 2025-04-10
fix: guaranteed order of preprocessor execution

DarioSoller:
_Description of changes:_
Having multiple preprocessors defined in the SD config, did not guarantee the execution of the preprocessors in the exact order, how they have been configured in the SD config.

Technical reason behind that is, that it has been looped over the properties of a JS object, which had the respective preprocessor functions connected to the preprocessor names as the object properties. As we all know Javascript has a flaw, when it comes to ordering in JS objects.

Therefore we now rely on the right order, by looping over the preprocessor names in the `appliedPreprocessors` array parameter, which already has been available in the `preprocess` function.

### DIFF
--- a/.changeset/swift-buses-march.md
+++ b/.changeset/swift-buses-march.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': patch
+---
+
+If several preprocessors are defined in the SD configuration, the execution of the preprocessors is now guaranteed in the exact order in which they were configured in the SD configuration.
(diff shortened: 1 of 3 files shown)


jorenbroekema:
Can you add a changeset? `npx changeset`, patch is fine. And maybe a regression test?

jorenbroekema (approved):

dbanksdesign (approved):

PR #1491, opened 2025-04-16
fix: Handling tokens with `.value` suffix in the name

not-in-stock:
### Description:
This PR fixes handling of tokens which name contains the suffix `.value`.

The issue occurs because the token processing logic uses the presence of a `.value` property to determine whether to traverse down the token tree. When a token's name itself contains `.value` (e.g., `colors.semantic.value`), this can cause incorrect behavior.

### Changes:
- Added proper detection of nested value properties to prevent incorrect token traversal
- Introduced `hasNestedValue` checks to distinguish between:
  - A token's name containing `.value` as part of its path
  - A token having a `.value` property that indicates it's a leaf node

### The fix ensures that:
1. Tokens with `.value` in their name are processed correctly
2. Nested value properties are properly handled
3. The token tree traversal behavior remains correct for actual value properties

### DIFF
--- a/lib/filterTokens.js
+++ b/lib/filterTokens.js
@@ -53,8 +53,11 @@ async function filterTokenObject(tokens, filter, options) {
     if (!isPlainObject(token)) {
       return acc;
     } else {
-      const tokenValue = options.usesDtcg ? token.$value : token.value;
-      if (typeof tokenValue === 'undefined') {
+      const valueKey = options.usesDtcg ? '$value' : 'value';
+      const tokenValue = token?.[valueKey];
+      const hasNestedValue = tokenValue && Object.hasOwn(tokenValue, valueKey);
+
+      if (typeof tokenValue === 'undefined' || hasNestedValue) {
         // If we got here we have an object that is not a token. We'll assume
         // it's token group containing multiple tokens and recursively filter it
         // using the `filterTokenObject` function.
--- a/__tests__/__json_files/nested_value.json
+++ b/__tests__/__json_files/nested_value.json
@@ -1,0 +1,9 @@
+{
+  "colors": {
+    "semantic": {
+      "value": {
+        "value": "#FF0000"
+      }
+    }
+  }
+}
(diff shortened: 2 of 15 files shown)


lib/filterTokens.js
```diff
@@ -53,8 +53,11 @@ async function filterTokenObject(tokens, filter, options) {
     if (!isPlainObject(token)) {
       return acc;
     } else {
-      const tokenValue = options.usesDtcg ? token.$value : token.value;
-      if (typeof tokenValue === 'undefined') {
+      const valueKey = options.usesDtcg ? '$value' : 'value';
+      const tokenValue = token?.[valueKey];
+      const hasNestedValue = tokenValue && Object.hasOwn(tokenValue, valueKey);
```
jorenbroekema, line 58: Given how often this particular code is repeated, I feel like we should abstract it to a utility function `isToken(token, usesDtcg)`

jorenbroekema (changes requested):
Nice contribution! I think this is now possible now that we no longer (since v5) automatically strip `.value` suffixes from token references. Now, `.value` in a token reference has semantic meaning, it means a token hierarchy path that ends with "value".

That said, there's a problem: a token type that has a value that is of type "object", and inside that object there may be a value property. For example:

```json
{
  "foo": {
    "value": {
      "unit": "px",
      "value": 15
    },
    "type": "dimension"
  }
}
```

See: https://tr.designtokens.org/format/#dimension

> small note: we dont yet support this new dimension format in SD, there's an open issue for it

The heuristics for determining:

- Token group level
- Token level
- Token value (sub/inner) level

is getting rather confusing when the token value property is "value" (which is valid for all 3 levels) and not "$value" (only valid for 1 level).
So in the new DTCG format, which uses special $ prefix for special properties, this is much less of a problem than in the "older" SD format.

Maybe you can think of a way around this problem and it's fixable by expanding the heuristics a little bit that determines when a property with the key "value" is a `DesignToken` and not a `DesignTokens` (group) or a `DesignToken["value"]` (inner).
E.g. we could look at the "type" property. in the old SD format this is a mandatory field for a `DesignToken` (even though for DTCG format, this is optional if there's a `$type` on its parents somewhere)

Besides the above issue and my suggestion to abstract the utility that checks whether a prop is of type `DesignToken`, I'm on board with the changes.

PR #1510, opened 2025-05-14
feat(docs): Automatically save playground when switching tabs

floscr:
_Description of changes:_

https://github.com/user-attachments/assets/c0ab069a-df8c-4f8d-b726-dae75d367ebb

Automatically saves files when switching tabs.
I was pretty confused why my changes to test style dictionary on the docs site were being reset when I switched tabs, only after hitting `Ctrl+s` by a hunch I was able to test out the library.

Other Changes:

- Fixes broken hot reload of docs


docs/astro.config.mjs
```diff
@@ -15,6 +15,9 @@ export default defineConfig({
     server: {
       force: true,
     },
+    ssr: {
+      noExternal: ['starlight-links-validator'],
+    },
```
jorenbroekema, line 20: Why was this change needed? It's only a dev dependency used as a binary script when building the docs to check for broken links, I don't see why Vite would ever resolve an import to this module.

docs/src/components/sd-playground.ts
```diff
@@ -204,7 +204,15 @@ class SdPlayground extends LitElement {
...
-          this.currentFile = (ev.target as HTMLInputElement).value as Files;
+          const nextCurrentFile = (ev.target as HTMLInputElement).value as Files;
+          // Auto save when switching to output tab
+          if (nextCurrentFile === 'output' && this.currentFile !== 'output') {
+            this.saveFile().then(() => {
+              this.currentFile = nextCurrentFile;
+            });
+          } else {
+            this.currentFile = nextCurrentFile;
+          }
```
jorenbroekema, line 215: You can make the sl-change listener an async function and then simplify this:

```suggestion
          if (nextCurrentFile === 'output' && this.currentFile !== 'output') {
            await this.saveFile();
          }
          this.currentFile = nextCurrentFile;
```

jorenbroekema (changes requested):

jorenbroekema (approved):

PR #1511, opened 2025-05-15
fix: Update `loadFile` to deep clone ES module exports to avoid unintended mutations

chuganzy:
_Description of changes:_

This PR aims to fix an issue where the `isSource` property of tokens provided by ES Modules is not updated correctly when running Style Dictionary multiple times with different configurations within the same Node.js process.

This is because Style Dictionary directly mutates the resolved contents during the first run and reuses the value from the second time.

Alternative solution is to just update [`combineJSON`](https://github.com/amzn/style-dictionary/blob/86c2c30ba289121f7dc9c28f573d1996dbc4a3a8/lib/utils/combineJSON.js#L110) to always update the `isSource`, however I thought it makes more sense to clone the original content to make the behavior consistent with other file types. The downside of this approach is that non-serializable data types cannot be used, however I think it is not an issue because Style Dictionary uses `structuredClone` in other places too and does not expect non-serializable data to be put as a token value to begin with.


jorenbroekema (dismissed):
Oh interesting, I didn't know ES modules could be mutated like that, given that they are read-only.
I know in CommonJS module tampering is a thing, where one module could mutate another (which makes mocking easy), but I'm a little surprised this also works in ES modules as long as it's a nested prop of the exports.

Could you create a changeset for this? just a patch will do

PR #1516, opened 2025-05-21
De-Amazon the repository

jorenbroekema:
_Issue #, if available:_
fixes https://github.com/style-dictionary/style-dictionary/issues/1514


hyandell (changes requested):
For the LICENSE file, it's not intended that the Copyright be changed from the example. My recommendation there is to replace the LICENSE file with the text from https://www.apache.org/licenses/LICENSE-2.0.txt

Also noting that the package.json has amzn urls in it. 

Otherwise looks good to me :)

jorenbroekema:
> For the LICENSE file, it's not intended that the Copyright be changed from the example. My recommendation there is to replace the LICENSE file with the text from https://www.apache.org/licenses/LICENSE-2.0.txt

This example text does have a copyright placeholder, so does it make sense still to put Style Dictionary Contributors there?
Does it require a year stamp? As far as I know it's not needed, but has some advantages in US law context, but I could be mistaken.

hyandell:
> > For the LICENSE file, it's not intended that the Copyright be changed from the example. My recommendation there is to replace the LICENSE file with the text from https://www.apache.org/licenses/LICENSE-2.0.txt
> 
> This example text does have a copyright placeholder, so does it make sense still to put Style Dictionary Contributors there? Does it require a year stamp? As far as I know it's not needed, but has some advantages in US law context, but I could be mistaken.

It's an annoying confusion from how Apache-2.0 was written back in 2000 or so. That last piece is an example for folk on how to do a source header. Ideally it would have been part of the website rather than in the actual license. By changing it it makes it harder for tooling to notice that the file is a straight-up default Apache license.

My recommendation is to not have the year stamp. Source control, and even things like zip archive metadata, provide the timestamps with much better accuracy than a human edited year that is often forgotten about and sometimes updated too aggressively when there hasn't been a copyrightable change. One of those things that source control has obsoleted :)

jorenbroekema:
> That last piece is an example for folk on how to do a source header. Ideally it would have been part of the website rather than in the actual license.

Ah I see, okay I think it should be okay now then.

hyandell (dismissed):
Looks good.

I'd advise dropping year from the NOTICE file, but it's all up to you as a group how you handle that. Keeping just the initial year in there is also goodness.

jorenbroekema:
@hyandell needs a re-approve after I changed the NOTICE to remove the date

PR #1529, opened 2025-07-04
fix: no throw for cleaning non-existent files in silent mode

jorenbroekema:
User reported that they got errors in silent mode when cleaning non-existent files (which commonly happens in CI env where there are no output files yet).


Mwalthery:
Seems good!

Mwalthery (approved):

six7 (approved):

PR #1546, opened 2025-08-17
fix: resolve TypeScript issues with token type validation in transforms

Romkin7:
_Issue #, if available:_

Token structure has changed in V5 and there is not type key available directly under token, now it's inside attributes object.
```

{
    value: '4',
    filePath: 'styleDictionary/tokens/sizes/spacings.json',
    isSource: true,
    key: '{size.spacing.xxxl}',
    original: { value: '4', key: '{size.spacing.xxxl}' },
    name: 'fb-size-spacing-xxxl',
    attributes: { category: 'size', type: 'spacing', item: 'xxxl' },
    path: ['size', 'spacing', 'xxxl'],
};
```

_Description of changes:_

Commit Summary
This commit fixes TypeScript issues related to token type validation in the Style Dictionary transforms system. Here's what was changed:

Main Changes:

Fixed token property access pattern: Changed from [token.type] to [token.attributes?.type] throughout the codebase to match the proper token structure

Updated type filter functions: Modified several filter functions ([isDimension], [isFontSize], [isAsset], [isContent], [isColor], [processFontFamily], [transformCubicBezierCSS] to use the correct property path

Added type safety checks: For the [fontFamilyCss] and [cubicBezierCss] transforms, added [typeof type === 'string'] checks before calling [includes()] to prevent TypeScript errors when the type might be an object

Updated documentation comments: Fixed all JSDoc comments to reflect the correct property path ([token.attributes?.type]instead of [token.type].

Why this was needed:

The original code was accessing [token.type] directly, but the actual token structure stores type information in [token.attributes.type]
TypeScript was throwing errors because [includes()] method expected a string but could receive an object
This ensures compatibility with both legacy and new token formats while maintaining type safety
Files changed:

[transforms.js] - Updated all transform functions and their documentation
This fix ensures that the transform functions work correctly with the proper token structure and prevents TypeScript compilation errors.

### DIFF
--- a/lib/common/transforms.js
+++ b/lib/common/transforms.js
@@ -45,7 +45,7 @@ const camelOpts = {
  */
 export function isColor(token, options) {
   const val = options.usesDtcg ? token.$value : token.value;
-  const type = options.usesDtcg ? token.$type : token.type;
+  const type = options.usesDtcg ? token.$type : token.attributes?.type;
   return (
     type === 'color' &&
     Color(val).isValid() &&
@@ -62,7 +62,7 @@ export function isColor(token, options) {
  * @returns {boolean}
  */
 function isDimension(token, options) {
-  return (options.usesDtcg ? token.$type : token.type) === 'dimension';
+  return (options.usesDtcg ? token.$type : token.attributes?.type) === 'dimension';
 }
 
 /**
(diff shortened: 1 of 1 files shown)


jorenbroekema:
I appreciate the effort to contribute a PR, but I cannot accept this change.

The token's type property should be there on the root level of the token, can you create an issue describing how you got to a token with a missing type property? Then maybe I can advise you on how to get it back :)

PR #1557, opened 2025-08-26
fix: move patch-package to dev deps and prepare script

jorenbroekema:
_Issue #, if available:_
fixes https://github.com/style-dictionary/style-dictionary/issues/1548

_Description of changes:_

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
   },
   "dependencies": {
     "@bundled-es-modules/deepmerge": "^4.3.1",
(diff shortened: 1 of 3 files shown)


package.json
cylewaitforit, line 1: Will also need to depend on a version of @bundled-es-modules/glob that doesn't depend on patch-package. See https://github.com/bundled-es-modules/glob/pull/3

dbanksdesign (dismissed):

dbanksdesign (approved):

PR #1569, opened 2025-09-19
feat: add size/compose/{dp, sp} to builtin transforms

Pythonen:
_Description of changes:_

Hello!

I’m currently building a cross-platform design system at work, and style-dictionary has been an absolute breeze to work with — thanks for that!

While working on this, I noticed there isn’t a built-in transform for `size/compose/{dp,sp}`. Of course, it’s easy to define a custom transform in your own Style Dictionary config, and I understand that [SD aims to stay minimal](https://github.com/style-dictionary/style-dictionary/blob/main/CONTRIBUTING.md#what-should-be-included:~:text=we%20don%27t%20want%20to%20be%20a%20swiss%20army%20knife%20that%20does%20everything%20out%20of%20the%20box).

That said, there are already built-in transforms for [size/compose/remToSp](https://styledictionary.com/reference/hooks/transforms/predefined/#sizecomposeremtosp) and [size/compose/remToDp](https://styledictionary.com/reference/hooks/transforms/predefined/#sizecomposeremtodp).
Since [size/sp](https://styledictionary.com/reference/hooks/transforms/predefined/#sizesp) and [size/dp](https://styledictionary.com/reference/hooks/transforms/predefined/#sizedp) don’t work directly on compose objects (especially when the object contains both dimensions and colors) as these transforms do not add the dot before the unit (i.e., `10.00dp` vs `10.00.dp`), I thought it would make sense to include size/compose/{dp,sp} as built-in transforms for consistency.

If these transforms are deemed unnecessary or if I've missed something relevant to the matter, feel free to close this PR. 

### DIFF
--- a/docs/src/content/docs/reference/enums.md
+++ b/docs/src/content/docs/reference/enums.md
@@ -256,8 +256,8 @@ export const transforms = {
   sizeComposeRemToSp: 'size/compose/remToSp',
   sizeComposeRemToDp: 'size/compose/remToDp',
   sizeComposeEm: 'size/compose/em',
-  sizeComposeDp: 'size/compose/dp'
-  sizeComposeSp: 'size/compose/sp'
+  sizeComposeDp: 'size/compose/dp',
+  sizeComposeSp: 'size/compose/sp',
   sizeSwiftRemToCGFloat: 'size/swift/remToCGFloat',
   sizeRemToPx: 'size/remToPx',
   sizePxToRem: 'size/pxToRem',
--- a/docs/src/content/docs/reference/enums.md
+++ b/docs/src/content/docs/reference/enums.md
@@ -256,6 +256,8 @@ export const transforms = {
   sizeComposeRemToSp: 'size/compose/remToSp',
   sizeComposeRemToDp: 'size/compose/remToDp',
   sizeComposeEm: 'size/compose/em',
+  sizeComposeDp: 'size/compose/dp'
+  sizeComposeSp: 'size/compose/sp'
   sizeSwiftRemToCGFloat: 'size/swift/remToCGFloat',
   sizeRemToPx: 'size/remToPx',
   sizePxToRem: 'size/pxToRem',
--- a/.changeset/strong-impalas-enjoy.md
+++ b/.changeset/strong-impalas-enjoy.md
@@ -1,5 +1,5 @@
 ---
-'style-dictionary': patch
+'style-dictionary': minor
 ---
 
 Add new size/compose/{sp,dp} transforms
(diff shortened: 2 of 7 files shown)


lib/common/transformGroups.js
```diff
@@ -244,6 +248,8 @@ export default {
     sizeComposeEm,
     sizeComposeRemToSp,
     sizeComposeRemToDp,
+    sizeComposeSp,
+    sizeComposeDp,
```
jorenbroekema, line 252: This would be a breaking change I believe, adding them to the compose transformGroup means they are applied to tokens by default if this transformGroup is used. Due to the order, the last one wins and `.dp` will be added to all matching tokens. I don't think that's intended?

jorenbroekema (changes requested):
Great PR, and thanks for reading the contributing guidelines.

The only thing that's concerning is adding the transforms to the compose transformGroup, it means all dimension tokens will be .dp and all fontsize tokens .sp.

This may be a good opt-in for folks, but I am not sure we want this to be the default for compose? Can you maybe explain if (and why) this is your intention, to make it the default? Then maybe we can consider making it the default for the next major version of SD, but make it opt-in for v5.

Pythonen:
> Great PR, and thanks for reading the contributing guidelines.
> 
> The only thing that's concerning is adding the transforms to the compose transformGroup, it means all dimension tokens will be .dp and all fontsize tokens .sp.
> 
> This may be a good opt-in for folks, but I am not sure we want this to be the default for compose? Can you maybe explain if (and why) this is your intention, to make it the default? Then maybe we can consider making it the default for the next major version of SD, but make it opt-in for v5.

My initial thinking was that it would be kind of nice to have it on by default, as the default format wouldn't apply the units to the compose file if there were colors mixed in the file.

However, I think you're right in that it would be better to have it as opt-in the same way e.g.,`color/composeColor` is.

lib/common/transformGroups.js
Pythonen: Good catch! Yes as discussed, we are probably better of with just exposing them as opt-in options.

docs/src/content/docs/reference/enums.md
```diff
@@ -256,6 +256,8 @@ export const transforms = {
   sizeComposeRemToSp: 'size/compose/remToSp',
   sizeComposeRemToDp: 'size/compose/remToDp',
   sizeComposeEm: 'size/compose/em',
+  sizeComposeDp: 'size/compose/dp'
+  sizeComposeSp: 'size/compose/sp'
```
jorenbroekema, line 260:
```suggestion
  sizeComposeDp: 'size/compose/dp',
  sizeComposeSp: 'size/compose/sp',
```
only just noticed this, missing commas

jorenbroekema (dismissed):

PR #1575, opened 2025-09-29
fix: handle double quoted fontNames for fontFamily/css

### LINKED ISSUE #1476: fontFamily/css does not handle double quotes
Is there a specific reason why the `fontFamily/css` transform only handles single quotes and ignores double quotes?

The current behavior unfortunately leads to problems when a font family string uses double quotes which is a valid alternative to single quotes in CSS as far as I know.

For example, `fontFamily/css` transforms the string
```
Some Font, "With Double Quotes", 'And Single Quotes'
```
to
```
'Some Font', '"With Double Quotes"', 'And Single Quotes'
```
which breaks the second entry because the browser would then look for a font named `"With Double Quotes"` including the quotes.

I am using style-dictionary 4.3.3.

Pythonen:
Closes #1476 

_Description of changes:_

Check if fontName is wrapped in single quotes OR double quotes.


jorenbroekema:
@Pythonen released v5.1.0 just now, thanks for contributing! Feel free to reach out if you need anything else

Pythonen:
> @Pythonen released v5.1.0 just now, thanks for contributing! Feel free to reach out if you need anything else

Will do, thanks!

Issue #1563, opened 2025-08-28
Error "filtered out token references were found" and references not resolved when token name starts with "value"


julrich-bee360:
When using the `outputReferences` option, references to tokens whose name starts with `value` are not resolved.

The culprit is [line 54 in `getReferences.js`](https://github.com/style-dictionary/style-dictionary/blob/0dba4ccde6ced7bcd8677e6d74052740c6aae05e/lib/utils/references/getReferences.js#L54) which removes all occurrences of `.value` from the reference.

In my use case, the referenced token is named `value_chain` inside a group named `object_type`. Therefore, the code tries to find a token named `object_type_chain` which does not exist because the token is actually named `object_type.value_chain`.

Is it (still) supported to reference a token using `{<token path>}.value` resp. `{<token path>}.$value`?
If not, I guess this behavior should simply be removed.
If so, then I guess the code should be changed to only remove the `.value` at the end?! So maybe something like
```js
const trailingValueReferenceRegEx = usesDtcg ? /\.\$value$/ : /\.value$/;
variable = variable.trim().replace(trailingValueReferenceRegEx, '');
```

And a related problem: in [line 199 of createPropertyFormatter.js](https://github.com/style-dictionary/style-dictionary/blob/0dba4ccde6ced7bcd8677e6d74052740c6aae05e/lib/common/formatHelpers/createPropertyFormatter.js#L199) and [in lines 63-70 of sortByReference.js](https://github.com/style-dictionary/style-dictionary/blob/main/lib/common/formatHelpers/sortByReference.js#L63-L70), the code misses to pass the `usesDtcg` flag. Therefore, the bug also appears if the tokens use DTCG format.
So those lines should be changed to
```js
        { unfilteredTokens, usesDtcg, warnImmediately: false },
```

jorenbroekema:
Ah yea makes sense to me, PR to fix this is very welcome :)!

julrich-bee360:
Fixed by #1577 and released with version 5.1.1.

PR #1577, opened 2025-10-06
fix: resolve references to tokens with 'value' in their name

MarcelOlsen:
_Issue #, if available:_ #1563 

  _Description of changes:_

  Fixed bug in outputReferences when token name contains word 'value'.

  ## Problem
  Function `getReferences()` was using `.replace()` with string parameter.
  This removes first occurrence of `.value`, not the last one. Breaking tokens
   like:
  - Token name: `object_type.value_chain`
  - Reference: `{object_type.value_chain.value}`
  - Current behavior: resolves to `object_type_chain.value` (wrong)
  - Expected behavior: resolves to `object_type.value_chain` (correct)

  ## Solution
  Changed string replace to regex that matches only end of string:
  - **getReferences.js**: Use regex `/\.value$/` or `/\.\$value$/` to remove
  only trailing suffix
  - **createPropertyFormatter.js**: Pass `usesDtcg` flag to `getReferences()`
  (was missing)
  - **sortByReference.js**: Pass `usesDtcg` flag to `getReferences()` (was
  missing)
  - **Tests**: Added 8 test cases for tokens with 'value' in name

  ## Testing
  All tests passing. Added tests for both traditional and DTCG formats.

  Works correctly now for both token formats.

  By submitting this pull request, I confirm that my contribution is made
  under the terms of the Apache 2.0 license.

### DIFF
--- a/.changeset/fix-value-in-token-names.md
+++ b/.changeset/fix-value-in-token-names.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': patch
+---
+
+Fix outputReferences for tokens with 'value' in their name. Previously, references to tokens like `object_type.value_chain` were incorrectly resolved because the code removed the first occurrence of `.value` instead of only the trailing suffix.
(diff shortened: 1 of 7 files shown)


jorenbroekema (approved):

jorenbroekema:
Thanks a ton for the fix :)! I'll release this asap

jorenbroekema:
released v5.1.1

Issue #1584, opened 2025-10-27
design tokens spec dimension tokens


tkgroot:
I revisited the design token specification and saw that the spec has changed for dimensions tokens.
I was already hacking together some changes for `shadow/css/shorthand` transformer and maybe more, but before I go and get loose on this, maybe good to know what your take on this is.

do you actually want to implement these spec changes in the first place?
since the changes are mostly breaking with existing implementations I was wondering if I should make them backwards-compatible?

[https://www.designtokens.org/tr/drafts/format/#dimension](https://www.designtokens.org/tr/drafts/format/#dimension)

```json
"spacing-stack-0": {
    "$value": {
      "value": 0,
      "unit": "px"
    },
    "$type": "dimension"
}
```

i'm happy and willing to contribute and help implement these wherever necessary. just would like to align first before creating unnecessary work.

tkgroot:
similar mentioning and discussions around this topic within this issue #1398

PR #1585, opened 2025-10-29
feat: update functionality for size transformers to match new DTCG spec

tkgroot:
_Issue #1584 

_Description of changes:_

- transformers can now handle the changed/updated design tokens specification for dimension tokens
- build in a backwards-compatible way so that the older design token specification can still be used

I began adding a `changeset` for every transformer. However, I think that this is not necessary and can be changed. But for now I want to leave it this way and wait for comments and review. 
I've created multiple atomic commits for every transformer, so that it is easier to re/view commit by commit, isolate them and test them independently (like all the benefits of using atomic commits basically 😅).

### DIFF
--- a/__integration__/__snapshots__/async.test.snap.js
+++ b/__integration__/__snapshots__/async.test.snap.js
@@ -161,7 +161,7 @@ snapshots["integration async hooks should match snapshot"] =
   --color-font-warning: #601700;
   --color-font-success: #08422f;
   --size-border-radius-large: 30rem;
-  --size-border-radius-medium: 24rem;
+  --size-border-radius-medium: 1.5rem;
   --size-border-radius-small: 16rem;
   --size-font-small: 0.75rem;
   --size-font-medium: 1rem;
--- a/__integration__/__snapshots__/css.test.snap.js
+++ b/__integration__/__snapshots__/css.test.snap.js
@@ -164,7 +164,7 @@ snapshots["integration css css/variables should match snapshot"] =
     --color-font-warning: #601700;
     --color-font-success: #08422f;
     --size-border-radius-large: 30rem;
-    --size-border-radius-medium: 24rem;
+    --size-border-radius-medium: 1.5rem;
     --size-border-radius-small: 16rem;
     --size-font-small: 0.75rem;
     --size-font-medium: 1rem;
@@ -319,7 +319,7 @@ snapshots["integration css css/variables with references should match snapshot"]
   --color-core-yellow-1000: #542a00;
   --color-core-yellow-1100: #2d1a05;
   --size-border-radius-large: 30rem;
-  --size-border-radius-medium: 24rem;
+  --size-border-radius-medium: 1.5rem;
   --size-border-radius-small: 16rem;
   --size-font-small: 0.75rem;
   --size-font-medium: 1rem;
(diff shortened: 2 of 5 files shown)


tkgroot:
Even with this transformer adaption I wonder if the return value should also reflect and keep the nested object structure in its value.

in the following it does not keep the nested object structure and just returns a string as value
```
const myVal = transforms[sizeComposeDp].transform({ value: { value: 12, unit: 'px' }, {}, {}); // myVal = 12dp
```

below the return value and the token value would still be an object. However, the `value` and `unit` would not be inline with the DTCG specification. `An integer or floating-point value representing the numeric value` and `Unit of distance. Supported values: "px", "rem".`, as it is clearly stated in the [spec](https://www.designtokens.org/tr/2025.10/format/#dimension). I assume that every other value would be prohibited or at least not allowed when it comes to "unhandled" design tokens, but processed design tokens or `TransformedTokens` within style-dictionary could follow an internal representation of the tokens which does not need to follow the spec.

```
const myVal = transforms[sizeComposeDp].transform({ value: { value: 12, unit: 'px' }, {}, {}); // myVal = { value: 12, unit: 'dp' }
```

This makes me wonder in how style-dictionary want to handle these value objects internally. Does it want to transform the value object of a dimension token to provide the correct `unit` after transformation or does it want to transform the value into a string as most transformers do?

```json
{
 "spacing-stack-0": {
    "original": { "value" : { "value": 1, "unit": "px" } },
    "$type": "dimension",
    "$value": { "value": "16.00", "unit": "dp" }
   },
  "spacing-stack-1": {
    "original": { "value" : { "value": 4, "unit": "px" } },
    "$type": "dimension",
    "$value": "64.00.dp"
  }
}
```

Changing the transformers return value will impact transformation or formatting of tokens down the line. Biggest impact I could find is the `createPropertyFormatter` which will fail processing the token values accordingly and subsequently a lot of formatters will fail to output the expected format

tkgroot:
During implementation I came across a couple of transformers who assume that all token values are basically `rem`. Due to the change in the DTCG spec we can now explicitly provide the unit, therefore some transformers actually should, in my opinion, output a different transformed value.

There would be an option to provide formatting of value's accordingly by introducing a property for the `PlatformConfig` or `LocalOptions` , ex. `useUnits`. The benefit is that this would not be a breaking-change and will keep the existing functionality. One might consider the way the transformers now handle units provided within the token value as a bug of the transformer and than it would become a fix instead of a breaking-change.

If `useUnits` is set one could assume the consumer explicitly wants to use the unit provided with the token value and we could use transformers which do/do not scale token values to take the unit into account and act accordingly. Sometimes it really doesn't make sense to scale values, or to consider all values are being `rem` values, ex. `sizeRemToSp`, `sizePx`, `sizeRem`, `sizeRemToPx` and `sizePxToRem` are prominent examples where the unit actually matters.

tkgroot:
One thing I realized when applying the new DTCG spec/format to my current setup with the patched-version (all changes in this PR) of style-dictionary is that my spacing tokens generate a different output. Thus even though we can consider this change non-breaking it might be breaking on the consumer side.

```json
{
  "spacing-stack-0": {
    "$value": "14px"
    "$type": "dimension"
  },
  "spacing-stack-1": {
  "$value": {
      "value": 14,
      "unit": "px"
    },
    "$type": "dimension"
  }
}
```

one would expect when applying style-dictionary's `transformGroup` `web` or `css` the output would be the same, but that isn't the case.

```css
:root {
  --spacing-stack-0: 14px; 
  --spacing-stack-1: 14rem;
}
```

this is due to the `size/rem` transformer which transforms the `spacing-stack-0` differently than `spacing-stack-1`. Since `stack-0` already has a unit it will not transform the value to `rem` and instead returns the value without transforming it. `stack-1` does not have a unit and therefore will be transformed. 

Should I update the `size/rem` transformer to reflect and handle this appropriately and be non-breaking on the consumer side and adapt the tests, meaning considering that a unit has been provided, or should it just stay "dumb"?

jorenbroekema:
done by https://github.com/style-dictionary/style-dictionary/pull/1658

PR #1586, opened 2025-10-30
docs: fix layout issue

morinokami:
_Issue #, if available:_

_Description of changes:_ Setting `a` elements to `inline-block` was causing a layout issue in the header and link buttons. I've removed this style to restore proper display. I've reviewed each page and confirmed that no other issues have been introduced. (Apologies if there was a specific requirement for `inline-block`!)

| Before | After |
| --- | --- |
| Screenshot 2025-10-31 at 0 52 57 | Screenshot 2025-10-31 at 0 54 15 |

### DIFF
--- a/docs/src/styles.css
+++ b/docs/src/styles.css
@@ -39,10 +39,6 @@ table code {
   max-width: 900px;
 }
 
-a {
-  display: inline-block;
-}
-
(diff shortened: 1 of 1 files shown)


jorenbroekema:
Thanks for that one :) !

PR #1587, opened 2025-10-30
fix: size/rem transformer NaN handling

tkgroot:
_Issue #, if available:_

_Description of changes:_

- fix: handling of NaN property value of token in size/rem transformer
- chore: fix test-case to test size/rem transformer instead of size/dp transformer


jorenbroekema:
Thanks! good find

PR #1592, opened 2025-11-06
Create SECURITY.md for security policy

meetpansuriya04:
Added a security policy document outlining supported versions and vulnerability reporting.

_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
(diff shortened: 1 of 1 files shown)


meetpansuriya04:
good to go?

jorenbroekema:
Not good to go, placeholder content isn't helpful as a security.md document.

I'm open to adding one, but we should tell people how to contact us about security issues. Honestly, at this point, disclosing vulnerabilities is best done by emailing me or Danny directly, until we have an official email address. So, thanks for starting the initiative, but not yet ready for merging.

As for which versions we support for vulnerability fixes, I'd say v3.x.x, v4.x.x, v5.x.x, but I'd like to get @dbanksdesign input on that, maybe at this point it's sensible to drop v3.x.x given v4 has been out for quite a while. But we also have to look at the amount of users on majors and npm downloads suggest (at least to me) that v3 should get some amount of support for vulnerability fixes.

dbanksdesign:
Apologies for the late reply. Just checked the downloads per major version and 3.x is still at 46.8%, but 4.x has been out since June 28, 2024 so it has been over a year. Typically I've seen organizations give a warning before fully deprecating a version and we probably need to think about the deprecation strategy (marking the 3.x major version as deprecated in npm). I would probably say we should give 3.x users 6 months to a year to upgrade so we would support security fixes on 3.x for lets say a year and then mark that version as deprecated. 

For this PR thought, I agree with @jorenbroekema that I don't want to have placeholder content for a security document. I think we should get some examples together of what a good security document looks like and think through the security policy.

PR #1605, opened 2025-12-03
Fix fontFamily token reference quoting in CSS output

### LINKED ISSUE #1600: composite typography fontFamily token reference gets interpreted as string
From my understanding in the spec it says the `fontFamily` can be a value or token, however SD always transforms it as a string. 

spec: [fontFamily](https://www.designtokens.org/tr/drafts/format/#typography)

Input:

```json
{
  "heading-xxl": {
    "$type": "typography",
    "$value": {
      "fontFamily": "{Typography.Heading.Family Sans}",
      "fontSize": "{Typography.Heading.Size XXL}",
      "fontWeight": "{Typography.Heading.Weight Bold}",
      "fontStyle": "normal",
      "lineHeight": "{Typography.Line Height.LH 9}"
    }
  }
}
```

Output:

```css
:root {
  --heading-xxl: normal var(--typography-heading-weight-bold)
    var(--typography-heading-size-xxl) /
    var(--typography-line-height-lh-9)
    'var(--typography-heading-family-sans)';
}
```

Expected output:

```css
:root {
  --heading-xxl: normal var(--typography-heading-weight-bold)
    var(--typography-heading-size-xxl) /
    var(--typography-line-height-lh-9)
    var(--typography-heading-family-sans);
}
```

cnaples79:
## Summary
Fixes the bug where composite typography tokens with fontFamily references output incorrectly quoted CSS variables.

## Problem
When a typography token uses a reference for fontFamily, the output was:
```css
:root {
  --heading-xxl: normal var(--typography-heading-weight-bold)
    var(--typography-heading-size-xxl) /
    var(--typography-line-height-lh-9)
    'var(--typography-heading-family-sans)';
}
```

The fontFamily reference gets wrapped in quotes, resulting in invalid CSS.

## Expected Behavior
```css
:root {
  --heading-xxl: normal var(--typography-heading-weight-bold)
    var(--typography-heading-size-xxl) /
    var(--typography-line-height-lh-9)
    var(--typography-heading-family-sans);
}
```

## Root Cause
In `createPropertyFormatter.js`, when processing object values with `outputReferences: true`, the code replaced the reference value directly. For fontFamily, the value is often quoted (e.g., `'Open Sans'`), and when replaced with `var(--token-name)`, the quotes remained around it.

## Solution
Added logic to detect and handle quoted reference values (both single and double quotes) when replacing object values. When a quoted version is found, it's replaced with the CSS variable without preserving the surrounding quotes.

## Changes Made
- Modified `lib/common/formatHelpers/createPropertyFormatter.js` (lines 227-246)
- Added check for quoted versions of reference values before standard replacement
- Handles both single-quoted and double-quoted values
- Maintains backward compatibility with unquoted values

## Testing
This fix resolves the issue described in the reproduction case:
- Input: Typography token with fontFamily reference
- Output: CSS variable without surrounding quotes
- Works with both `$type: "fontFamily"` (explicit) and inherited types

Fixes #1600

### DIFF
--- a/lib/common/formatHelpers/createPropertyFormatter.js
+++ b/lib/common/formatHelpers/createPropertyFormatter.js
@@ -226,10 +226,24 @@ export default function createPropertyFormatter({
           };
           // TODO: add test
           // technically speaking a reference can be made to a number or boolean token, in this case we stringify it first
-          value = `${value}`.replace(
-            originalIsObject ? refVal : new RegExp(`{${ref.path.join('\\.')}(\\.\\$?value)?}`, 'g'),
-            replaceFunc,
-          );
+
+          // When replacing object values, also check for quoted versions of the value
+          // This fixes typography tokens where fontFamily values are quoted (e.g., 'Open Sans')
+          // and need to be replaced with CSS variables without the surrounding quotes
+          const stringValue = `${value}`;
+          const quotedRefVal = `'${refVal}'`;
+          const doubleQuotedRefVal = `"${refVal}"`;
+
+          if (originalIsObject && stringValue.includes(quotedRefVal)) {
+            value = stringValue.replace(quotedRefVal, replaceFunc());
+          } else if (originalIsObject && stringValue.includes(doubleQuotedRefVal)) {
+            value = stringValue.replace(doubleQuotedRefVal, replaceFunc());
+          } else {
+            value = stringValue.replace(
+              originalIsObject ? refVal : new RegExp(`{${ref.path.join('\\.')}(\\.\\$?value)?}`, 'g'),
+              replaceFunc,
+            );
+          }
         }
       });
     }


jorenbroekema:
So, this took me a while to be able to reproduce in a failing test but I managed to do it eventually in `createPropertyFormatter.test.js`:

```js
        it.only('should handle quoted values e.g. fontFamily properly', () => {
          const tokens = {
            typo: {
              original: {
                value: {
                  fontSize: '16px',
                  fontFamily: '{family}',
                },
                type: 'typography',
              },
              name: 'typo',
              path: ['typo'],
              value: "16px 'Open Sans, sans-serif'",
              type: 'typography',
            },
            family: {
              original: {
                value: 'Open Sans, sans-serif',
                type: 'fontFamily',
              },
              name: 'family',
              path: ['family'],
              value: 'Open Sans, sans-serif',
              type: 'fontFamily',
            },
          };
          
          const propFormatter = createPropertyFormatter({
            outputReferences: true,
            dictionary: {
              tokens,
            },
            format: css,
          });
          expect(propFormatter(tokens.typo)).to.equal('  --typo: 16px var(--family);');
        });
```

This ends up failing because the output is `  --typo: 16px 'var(--family)';`

The core reason why this happens is because in the fontFamily transform, the fontFamily is wrapped in quotes to ensure a valid CSS font-family value inside the stringified typography token (CSS font value).
When Style-Dictionary adds references back into the output (CSS in this case), it does a regex find & replace, but because the value has changed from its original value (no wrapping quotes), it ends up creating a weird output.

Generally speaking, when the value itself is changed significantly (like wrapping it in quotes), outputting references is unsafe. 
This is why I created [this utility](https://styledictionary.com/reference/utils/references/#outputreferencestransformed) for now to prevent this from happening, by not outputting a reference if it is unsafe to do so.

Unfortunately this cannot be easily fixed without serious technical debt; the part of the code that outputs reference should not be coupled in any way to how specific transforms work. These should remain decoupled. Therefore I cannot accept this workaround fix.

That said, there's multiple open issues about this and it's probably the one I want to fix the most out of any particular issues in the next major version of Style Dictionary (v6), but it requires a pretty big rewrite of how transforms work, how outputting references works, and how they ought to be coupled within the scope of transforms themselves.

Relevant issues for you to track:
https://github.com/style-dictionary/style-dictionary/issues/1602
https://github.com/style-dictionary/style-dictionary/issues/1415
https://github.com/style-dictionary/style-dictionary/issues/1063

I hope to be able to tackle all of those in the rewrite, the rough gist of it is that:
- Transforms should be able to transform values with references in them, some transformations don't require the resolving of refs e.g. stringifying an object value like for CSS fonts shorthand for typography tokens.
- Transforms are responsible for resolving references (via easy to use utilities or through accessing some kind of reference/dependency map) inside their own scope, if they need to do that, e.g. because they are transitive by nature like color modifiers or math compiler transforms
- Formats will output refs only for values that have strings matching reference syntax `{...}` rather than this error-prone method of find & replacing the value (which could be transformed by then 😓) within the string


Hopefully this makes sense, I appreciate the PR and I didn't realize this issue was actually a duplicate/symptom of the issues above, which is why I said "PR welcome" in the issue. I thought this would be a simpler bug, apologies!

PR #1627, opened 2026-01-22
feat: use a logger instance to track issues in SD

jorenbroekema:
New logger which is extremely flexible in controlling exactly which logs are logged at what level.
Also allows error throws to be milder if non-fatal, like broken references in token values.

**This PR will be merged to the "logger" branch** which is an intermediate branch based on current main just to collect logger-related work/PRs in pieces.  @dbanksdesign I renamed your very old logger branch to logger-backup-danny in case we still want to look at it as a reference point for discussion or something. We could delete it once this work is completed.

WIP, still needs:
- docs -> as a last PR
- all logs/error catches to be using the logger, we started with the transform errors but almost everything else still needs to be done -> separate PRs to this branch

For this PR TODOs -> see comments +:
- [x] changeset
- [x] coverage check


lib/utils/logging/logger.js
```diff
@@ -0,0 +1,197 @@
...
+  silent: 999,
+};
+
+export class Logger {
+  /**
+   * @private
+   * @returns {Map<keyof logKeys, Map<string, Log[]>>}
+   */
+  get logMap() {
+    return new Map([
```
jorenbroekema: Probably can be shortened so it scales and we don't need to maintain this list here.

lib/utils/logging/logger.js
```diff
@@ -0,0 +1,197 @@
...
+    this.warn = {
+      collisions: new Map(),
+      nameCollisions: new Map(),
+    };
+
+    this.success = {};
+
+    this.info = {};
+
+    this.debug = {};
```
jorenbroekema: maybe can be done dynamically as well based on the enums. So we don't maintain the same list here

PR #1628, opened 2026-01-22
feat(utils): add stable applyReferenceSafeOrdering for reference-safe token ordering

Yeom-JinHo:
Follow-up work for https://github.com/style-dictionary/style-dictionary/pull/1611 🙇

## Description

This PR improves `formattedVariables` ordering behavior when `outputReferences: true` by ensuring **define-before-use** semantics without unnecessarily reshuffling tokens.

Instead of relying on a comparator-style reference sorter, we introduce a dedicated final-pass utility, `applyReferenceSafeOrdering(sortedTokens, tokens, opts)`, which **starts from the user-sorted array** and performs a **stable, reference-safe ordering pass**.

This approach preserves the user's sort order as much as possible while guaranteeing referenced tokens are defined before the tokens that use them.

## Changes

- **New utility: `applyReferenceSafeOrdering`**
  - Enforces reference-safe ordering (define-before-use) as a final pass.
  - Uses a stable Kahn-style topological ordering strategy.
  - Keeps ordering deterministic across runs.
  - Handles cycles gracefully via a deterministic fallback (does not crash).

- **`formattedVariables` integration**
  - First applies user sorters to produce `sortedTokens`.
  - When `outputReferences: true`, runs `applyReferenceSafeOrdering` as the final pass.
  - No additional sorters are applied after the reference-safe adjustment.

- **Stability fix**
  - Ensures the initial `available` queue is sorted by `originalIndex`, so that when multiple nodes are available, we always pick the earliest token from the user-sorted order.

- **Tests**
  - Added unit tests covering:
    - no references (order preserved)
    - direct dependencies
    - transitive dependencies
    - minimal movement from a user order like `(c, d, a, b)`
    - ignoring references to tokens not present in `sortedTokens`
    - cycles (deterministic, non-crashing fallback)
    - DTCG (`usesDtcg: true`)

## Motivation

When emitting CSS/SCSS/LESS/Stylus variables with `outputReferences: true`, we must guarantee that referenced tokens are defined before use. However, users also expect their chosen sorting strategy (e.g. by name or custom comparator) to be preserved whenever possible.

A dedicated final-pass reordering step makes the responsibility explicit:
- **User sorting** happens first.
- **Reference safety** is applied last, with minimal movement.

This yields more predictable output and avoids unnecessary reshuffling.

### DIFF
--- a/__tests__/common/formatHelpers/applyReferenceSafeOrdering.test.js
+++ b/__tests__/common/formatHelpers/applyReferenceSafeOrdering.test.js
@@ -1,0 +1,240 @@
+import { expect } from 'chai';
+import applyReferenceSafeOrdering from '../../../lib/common/formatHelpers/applyReferenceSafeOrdering.js';
+
+const createToken = (name, value, originalValue, usesDtcg = false) => {
+  const valueKey = usesDtcg ? '$value' : 'value';
+  return {
+    name,
+    [valueKey]: value,
+    original: { [valueKey]: originalValue },
+    path: name.split('-'),
+  };
+};
+
+describe('common', () => {
+  describe('formatHelpers', () => {
+    describe('applyReferenceSafeOrdering', () => {
+      it('should preserve order when no references exist', () => {
+        const tokenA = createToken('token-a', '#000', '#000');
+        const tokenB = createToken('token-b', '#111', '#111');
+        const tokenC = createToken('token-c', '#222', '#222');
+        const tokenD = createToken('token-d', '#333', '#333');
+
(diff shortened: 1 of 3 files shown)


Yeom-JinHo:
@jorenbroekema 
I added the utility we discussed while implementing user-defined sorting for formattedVariables!!

When you have time, I’d really appreciate any feedback you can share😃

Yeom-JinHo:
The existing unit tests are failing because switching from sortByReference to applyReferenceSafeOrdering changes the output ordering.

From my perspective, this change is expected and the new ordering is correct. Could you take a look and, if everything looks good, update the snapshots accordingly?

Issue #1263, opened 2024-07-03
Use @style-dictionary monorepo


jorenbroekema:
Just throwing out the idea because @dbanksdesign has brought it up a few times and I also think it makes sense.

We could use a monorepo and workspaces, which is handy for some of the packages/utilities/examples that we have.

Potentially:

- `@style-dictionary/core` -> Node API
- `@style-dictionary/cli` -> CLI, depends on core of course
- `@style-dictionary/example-react` -> there would be many of these, can be used in combination with CLI so user can initialize using `style-dictionary init`, will throw error if an example is picked that isn't installed.

It should be noted that we could potentially also split up some of the helpers / utils:
- `@style-dictionary/utils`

PR #1630, opened 2026-01-30
fix(examples): resolve directory path in matching-build-files tokens for ESM

### LINKED ISSUE #1509: [BUG] Matching Build Files fails to build with ENOENT
Attempting to run `build` on the `matching-build-files` example errors out with an "ENOENT" message stemming from the `./tokens/index.js` file.

Steps to reproduce:
1. Clone style-dictionary 
2. switch to examples/advanced/matching-build-files dir
3. Install - run `npm install`
4. Build - run `npm run build`
5. node errors out:

> Error: ENOENT: no such file or directory, scandir 'file:///Users/[myuser]/Code/playground/temp/style-dictionary/examples/advanced/matching-build-files/tokens/index.js'

I can verify that the `file:///` path in the error _does_ exist and correctly resolves to the `index.js`. 

Image

Yeom-JinHo:
closed #1509 

## Description

This PR fixes an `ENOENT` error in the **`examples/advanced/matching-build-files`** example when running `npm run build`.

The issue was caused by passing `import.meta.url` (a **file URL string**, e.g. `file:///.../tokens/index.js`) directly into Node’s `fs.readdirSync`, which expects a **filesystem path**. As a result, Node attempted to scan a non-existent path that literally starts with `file:///...`, producing `ENOENT`.

This change converts the module URL to a real filesystem path and ensures we pass the **tokens directory** (not the `index.js` file) to `readdirSync`.

## Changes

- **Fix path resolution for ESM**: Convert `import.meta.url` → filesystem path via `fileURLToPath`, then derive the directory using `dirname`.
- **Correct scan target**: Scan the `tokens/` directory instead of mistakenly scanning the `tokens/index.js` file URL.
- **(Optional) Improve performance**: Use `readdirSync(..., { withFileTypes: true })` + `Dirent.isDirectory()` to avoid per-entry `statSync`.

## Motivation

- The example currently fails with an error like:
  - `Error: ENOENT: no such file or directory, scandir 'file:///.../tokens/index.js'`
- In ESM, `import.meta.url` returns an absolute **`file:` URL**, not a plain path string, so it must be converted before using it with path-based `fs` APIs.
- Fixing this improves the reliability of the example and avoids confusion for users following the repo examples.

### DIFF
--- a/examples/advanced/matching-build-files/tokens/index.js
+++ b/examples/advanced/matching-build-files/tokens/index.js
@@ -1,12 +1,8 @@
 import { readdirSync } from 'node:fs';
-import { dirname } from 'node:path';
-import { fileURLToPath } from 'node:url';
-
-const tokensDir = dirname(fileURLToPath(import.meta.url));
 
 const dirs = (p) =>
   readdirSync(p, { withFileTypes: true })
     .filter((d) => d.isDirectory())
     .map((d) => d.name);
 
-export default dirs(tokensDir);
+export default dirs(import.meta.dirname);
--- a/examples/advanced/matching-build-files/README.md
+++ b/examples/advanced/matching-build-files/README.md
@@ -18,25 +18,25 @@ At this point, you can run `npm run build`. This command will generate the outpu
 
 The "build" command processes the JSON files in the `tokens` folder. The `index.js` file adds each folder, allowing you to map through them in `config.js`. The script goes through each folder and generates a file for each folder and populates it with tokens that match the filter.
 
-```sh
+```json
 # tokens/color/base.json
 {
-   "color": {
-       "red": {
-            "value": "#FF0000"
-        }
-   }
+  "color": {
+    "red": {
+      "value": "#FF0000"
+    }
+  }
 }
 ```
 
-```sh
+```json
 # tokens/size/base.json
 {
-   "size": {
-       "small": {
-            "value": "2px"
-        }
-   }
+  "size": {
+    "small": {
+      "value": "2px"
+    }
+  }
 }
 ```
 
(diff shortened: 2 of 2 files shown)


Yeom-JinHo:
@dbanksdesign  @jorenbroekema 
While updating the examples, I noticed they’re all written for v4. Are there plans to upgrade the examples to v5?

examples/advanced/matching-build-files/tokens/index.js
```diff
@@ -1,4 +1,12 @@
...
+import { fileURLToPath } from 'node:url';
+
+const tokensDir = dirname(fileURLToPath(import.meta.url));
+
+const dirs = (p) =>
+  readdirSync(p, { withFileTypes: true })
+    .filter((d) => d.isDirectory())
+    .map((d) => d.name);
+
+export default dirs(tokensDir);
```
jorenbroekema:
```suggestion
import { readdirSync } from 'node:fs';

const dirs = (p) =>
  readdirSync(p, { withFileTypes: true })
    .filter((d) => d.isDirectory())
    .map((d) => d.name);

export default dirs(import.meta.dirname);
```
This is simpler I think. And supported in Node LTS version atm I think (v20.11+, v21.2+)

Let me know what you think @Yeom-JinHo , if it works on your end I'll merge this :)

Yeom-JinHo:
@jorenbroekema 
No worries at all — I know you’ve been busy, and I also wouldn’t want my sorting work to be merged in an unstable state. Please take your time reviewing it.

When I first started using this library, I relied a lot on the v4 examples. Right now it looks like all the examples are still written only for v4. Do you have any plans to upgrade the examples to v5?

If there is a plan, I’d be happy to contribute gradually over the longer term.

jorenbroekema:
> When I first started using this library, I relied a lot on the v4 examples. Right now it looks like all the examples are still written only for v4. Do you have any plans to upgrade the examples to v5?

Yes we have an issue planned for the v6 milestone that would put the example in its own monorepo workspace, possibly a separate workspace for each example if it has a separate package.json + dependencies.

See https://github.com/style-dictionary/style-dictionary/issues/1263

I am open to updating the examples for v5 though, I agree they were a bit left behind. Shouldn't be too many changes, since the v5 update was pretty subtle and non-breaking for most use cases.

Yeom-JinHo:
> > When I first started using this library, I relied a lot on the v4 examples. Right now it looks like all the examples are still written only for v4. Do you have any plans to upgrade the examples to v5?
> 
> Yes we have an issue planned for the v6 milestone that would put the example in its own monorepo workspace, possibly a separate workspace for each example if it has a separate package.json + dependencies.
> 
> See #1263
> 
> I am open to updating the examples for v5 though, I agree they were a bit left behind. Shouldn't be too many changes, since the v5 update was pretty subtle and non-breaking for most use cases.

Totally makes sense — thanks for sharing the direction.

I’d like to get a deeper understanding of the library, so I’ll take things slowly and, if the examples need updates along the way, I’ll work on them.

Would you be okay with me updating each example workspace’s package.json to target the latest stable version (v5)?
(e.g. bumping style-dictionary to ^5.x and making any small adjustments if needed.)

jorenbroekema:
Yep that sounds like a good start!

PR #1632, opened 2026-02-04
feat(dtcg): color module support

bennypowers:
_Issue #, if available:_

https://github.com/style-dictionary/style-dictionary/issues/1590#issuecomment-3715360239

_Description of changes:_

Adds support for DTCG v2025.10 color formats in color transformers

**NOTE**: in order to support the full gamut (:drum:) of color spaces, this PR introduces colorjs.io as a dependency.

### DIFF
--- a/lib/common/transforms.js
+++ b/lib/common/transforms.js
@@ -87,7 +87,7 @@ function dtcgToColorJS(dtcgColor) {
   const { colorSpace, components, alpha } = dtcgColor;
 
   // Map DTCG color space to colorjs.io color space ID
-  const colorjsSpace = DTCG_TO_COLORJS_SPACE[colorSpace] || colorSpace;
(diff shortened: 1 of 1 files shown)


lib/common/transforms.js
```diff
@@ -8,13 +9,176 @@ import { transforms, transformTypes } from '../enums/index.js';
...
+ * @returns {{ r: number, g: number, b: number, a: number }}
+ */
+export function getColorRgb(token, options) {
+  const val = options.usesDtcg ? token.$value : token.value;
+
+  if (isDTCGColorObject(val)) {
+    return dtcgColorToRgb(val);
+  }
+
+  // Legacy format - use tinycolor2
```
jorenbroekema: There should be an open issue for getting rid of tinycolor2 and using ColorJS.io for everything. Let's not do this in this PR and cause scope creep, but I just wanted to let you know that if you have time and you're feeling like it, we're open for a PR to migrate away from tinycolor2 entirely.

PR #1635, opened 2026-02-09
Fix/file header comment style

DarioSoller:
https://github.com/style-dictionary/style-dictionary/issues/1312

Introduces `formatting.fileHeader` options, in a backwards compatible way. Some old confusing options got deprecated and should get decommissioned and removed with SD v6. Detailed description WIP


DarioSoller:
@jorenbroekema if you have some time left, may you want to give it a quick check, if this goes in the right direction. I feel, we need some more tests to be more confident on all the possible option combinations, especially in regards of the backwards compatibility. Just let me know, what you think?

PR #1694, opened 2026-05-29
fix: regression 5.4.0 sizeRem transform, keep 0 type intact

### LINKED ISSUE #1693: Potentially Breaking Regression in v5.4.1
The `size/rem` transform change from **v5.4.1** also caused all design token values, which already have been `"0"` to now be `0` (of type number) in the outputs for the built-in JS and JSON formats. This type change within a JS or JSON consumer can be potentially breaking, because e.g. conditional `$value` null checks do not work the same anymore.

I boiled it down to [this code line in the changed transform](https://github.com/style-dictionary/style-dictionary/blob/46ec860e070076bc278bfc827238b87b639ee023/lib/common/transforms.js#L1210), which to my understanding could be changed from:
```javascript
if (parsedVal === 0) return 0;
```
to
```javascript
if (parsedVal === 0) return "0";
```

Open for helping with a PR and fix, if this is acknowledged to be a regression and bug?

jorenbroekema:
_Issue #, if available:_
fixes https://github.com/style-dictionary/style-dictionary/issues/1693

_Description of changes:_

### DIFF
--- a/.changeset/eight-ties-wait.md
+++ b/.changeset/eight-ties-wait.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': patch
+---
+
+Fix sizeRem and sizePxToRem transform to keep the 0-valued tokens type number/string intact in the result (regression 5.4.0)
--- a/__tests__/common/transforms.test.js
+++ b/__tests__/common/transforms.test.js
@@ -1271,8 +1271,17 @@ describe('common', () => {
         });
 
         ['0', 0].forEach((value) => {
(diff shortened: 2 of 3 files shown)


DarioSoller (approved):
I can confirm, that the fix works for me as intended:

- JS and JSON zero token value changed type from 0 to "0", which restores the primitive type that these tokens had before the regression introduced by v5.4.0.
- CSS/SCSS and custom format zero values are unchanged

PR #1700, opened 2026-06-08
feat: add emitEmptyFiles option to write files even when no tokens match

### LINKED ISSUE #1689: Format or platform option to create file whether tokens exist or not
There are a few places where I need a file emitted even when all tokens have been filtered out. Right now I have a custom skip extension that I can use to trick SD into emitting the tokenless file, but it's a hack.

It could be a format option, a platform option , or both.

Thanks for working on Style Dictionary. Over the past four years it has become essential infrastructure for me.

mvanhorn:
## Summary
A new `emitEmptyFiles` file option writes the output file even when no tokens match the filter, instead of skipping it.

## Why this matters
When a file's `filter` matches no tokens, Style Dictionary skips writing the file entirely. For build pipelines that expect a file to always exist (and commit it, or consume it downstream), a missing file is harder to handle than an empty one. `emitEmptyFiles: true` opts into writing the file with empty output; the default behavior is unchanged.

## Changes
- Add the `emitEmptyFiles` option to the `File` and `Config` types.
- Honor it in `StyleDictionary` when deciding whether to skip an empty file.

## Testing
Added tests for both the default skip behavior and the `emitEmptyFiles: true` path.

Fixes #1689

AI was used for assistance.

### DIFF
--- a/__tests__/__helpers.js
+++ b/__tests__/__helpers.js
@@ -37,8 +37,8 @@ export const fileToJSON = (_path, _fs = fs) => {
 
 export const clearOutput = (outputFolder = '__tests__/__output', _fs = fs) => {
   try {
-    _fs.rmdirSync(outputFolder);
-  } catch {
+    _fs.rmSync(outputFolder, { force: true, recursive: true });
+  } catch (_e) {
     //
   }
 };
--- a/.changeset/emit-empty-files.md
+++ b/.changeset/emit-empty-files.md
@@ -1,0 +1,5 @@
+---
+"style-dictionary": minor
+---
+
+Add `emitEmptyFiles` option to write output files even when no tokens match the file's filter, instead of skipping them.
--- a/__integration__/android.test.js
+++ b/__integration__/android.test.js
@@ -11,6 +11,7 @@ const { android } = transformGroups;
 
 describe('integration', async () => {
   before(async () => {
+    clearOutput(buildPath);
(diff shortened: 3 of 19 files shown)


jorenbroekema (commented):
Looks good! Can you run `npx changeset` and create a minor patch-note for this change/feature addition?

mvanhorn:
Done, added a minor changeset in `0eb119a` (`.changeset/emit-empty-files.md`) for the `emitEmptyFiles` option. The changeset-bot warning should clear now.

jorenbroekema (approved):

__tests__/__helpers.js
```diff
@@ -37,8 +37,8 @@ export const fileToJSON = (_path, _fs = fs) => {
 
 export const clearOutput = (outputFolder = '__tests__/__output', _fs = fs) => {
   try {
-    _fs.rmdirSync(outputFolder);
-  } catch {
+    _fs.rmSync(outputFolder, { force: true, recursive: true });
```
jorenbroekema, line 40: I can't quite believe this is the first time I encounter this difference being significant in an actual test 😓 took some debugging to figure it out, these methods have been historically quite messy in NodeJS

mvanhorn:
Thank you @jorenbroekema. The emitEmptyFiles option is a handy escape hatch when you want the output file regardless of token matches.

Issue #1633, opened 2026-02-05
[Docs] Auto-generate types and enums docs from file contents


jorenbroekema:
It's not great to maintain the same list of enums and types in both the source code as well as the MD docs files.

We should ideally just autogenerate. For enums this should be pretty easy via a remark plugin.

For types, I'd ideally also have anchors for types that are imported from other files, so you can click it and the docs redirect you to the right page.
Maybe also a remark plugin, or perhaps there is something existing already out there for TS files.

PRs welcome if someone wants to dive into this a bit. If you want an example of a working remark plugin you could check out:
- https://github.com/style-dictionary/style-dictionary/blob/main/docs/astro.config.mjs#L10
- https://github.com/style-dictionary/style-dictionary/blob/main/docs/src/remark-playground.ts

tomatotomata:
I was thinking about taking a first pass at the docs generation here. My initial idea is to start with the enum list through the existing Astro remark pipeline, then check whether the same approach can produce stable links for imported TypeScript types without making the docs build brittle. Since this is a broader docs-tooling change, do you think that direction fits the project, and is there a preferred source/docs pair to prototype against?

jorenbroekema:
I'll check @Guflly  PR for the enums. For the types, probably the "Configuration" section is a good first example to try out!

jorenbroekema:
Partially done -> enums

Geendersom:
Hi! I'd like to work on this issue as my first contribution to Style Dictionary. I saw the recent discussion about using an Astro remark plugin. Is this issue still available for contribution?

jorenbroekema:
@Geendersom  It's currently being worked on by @tomatotomata  https://github.com/style-dictionary/style-dictionary/pull/1726

I haven't had the time to thoroughly review this one yet though, but if you're keen to help out then it would be nice if you could check out that pull request and give it a test run and review yourself as well.

PR #1724, opened 2026-07-27
docs: generate enum reference from sources

Guflly:
_Issue #, if available:_ #1633

_Description of changes:_

Generates the enum reference from the modules exported by `lib/enums/index.js` during the documentation build. This keeps the reference synchronized with the source and adds focused coverage for the remark transform.


jorenbroekema (dismissed):
Awesome work! I actually don't really have comments on this one, looks great in the deploy preview! and we dont even need any new dev dependencies

jorenbroekema:
Ah it seems the test coverage went down from 98% to slightly below. There's an NPM script to open the coverage analysis to see what could be improved for the remark plugin to get the coverage up, I'm happy to help with it as well if you want.

docs/src/remark-enums.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+async function enumNodes(root) {
+  const enumDirectory = join(root, 'lib', 'enums');
+  const indexSource = await readFile(join(enumDirectory, 'index.js'), 'utf8');
+  const sourcePaths = [...indexSource.matchAll(exportPattern)].map((match) => match[1]);
+  const uniquePaths = [...new Set(sourcePaths)];
+
+  if (uniquePaths.length === 0) {
+    throw new Error('No enum modules were exported from lib/enums/index.js');
```
jorenbroekema, line 22: let's make the enums folder a variable, and make it a param for remarkEnums plugin function. That should also help you create a test for when a folder doesn't have any enum modules, adding test coverage for this error throw

Guflly: Done. The enum directory is now configurable, and I added the empty-directory test. I also moved these tests to the Node suite because the plugin reads from the filesystem. Coverage checks pass locally.

jorenbroekema (approved):

jorenbroekema:
Thanks a lot for this one! I'll try to find some time to look into the draft that was made for types

PR #1734, opened 2026-08-11
fix(formats): suppress comments when commentStyle is none

### LINKED ISSUE #1727: javascript/es6 format outputs undefined when commentStyle: "none" is set on tokens with descriptions
## Description of the bug
When using the javascript/es6 format with formatting: { commentStyle: "none" } in the file options, tokens that have a $description (or comment) field produce undefined as a literal string in the output instead of suppressing the comment.

## Environment detals
**Node version:** 22.15.1
**SD version:** 5.5.0

## Steps to reproduce
1. Have a token with a $description field in your token JSON.
2. Configure the javascript/es6 format with commentStyle: "none":

```
{
  destination: "color.js",
  format: "javascript/es6",
  options: {
    formatting: {
      commentStyle: "none",
    },
  },
}
```

3. Build. The output contains undefined after each token that has a description:

```
export const colorStateBackgroundDisabled = "#f4f5f2";
undefined;
export const colorStateBackgroundReadonly = "#fafbf9";
undefined;
```

## Expected behavior
Comments should be suppressed — no undefined, just the export const line.

## Root cause:
addComment in createPropertyFormatter.js has a switch with cases only for short and long. When commentStyle is none, no case matches, so processedComment remains undefined. It then gets concatenated into the output string:

```
// createPropertyFormatter.js
switch (commentStyle) {
  case short: ...
  case long:  ...
  // no case for `none` → processedComment stays undefined
}
to_ret_token = `${to_ret_token} ${processedComment}`; // → "export const x = y; undefined"
```

By contrast, createPropertyFormatter (used by CSS/SCSS/LESS formats) has a guard that correctly skips addComment when commentStyle is none:

```
// createPropertyFormatter.js line 247
if (comment && commentStyle !== none) {
  to_ret_token = addComment(to_ret_token, comment, mergedOptions);
}
```

## Suggested fix

Either add the same guard in the javascript/es6 format before calling addComment:

```
return comment && format.commentStyle !== 'none'
  ? addComment(to_ret, comment, format)
  : to_ret;
```

Or add case none: break; inside addComment's switch so processedComment stays empty string instead of undefined.

If you agree that this is an issue, I can tackle it as a contribution.
Thanks.

lara-ferrer:
## Issue
Fixes #1727. javascript/es6 and typescript/es6-declarations no longer output literal "undefined" for tokens with descriptions when commentStyle is set to none.

## Description of changes

When using the `javascript/es6` or `typescript/es6-declarations` formats with `commentStyle: "none"`, tokens with a `$description` or `comment` field were producing literal `undefined` in the output instead of suppressing comments.

**Root cause:** `addComment` only handled `short` and `long` comment styles. With `none`, `processedComment` stayed `undefined` and was concatenated into the output string. CSS/SCSS formats already guarded against this in `createPropertyFormatter`, but the JS/TS formats did not.

**Changes:**
1. **`addComment`** — early return when `commentStyle` is `none`, so the token string is returned unchanged.
2. **`javascript/es6` and `typescript/es6-declarations`** — added the same guard used elsewhere: only call `addComment` when a comment exists and `commentStyle !== none`.
3. **Tests** — added coverage in `es6Constants.test.js` and `createPropertyFormatter.test.js`.

**Before:**
```js
export const colorStateBackgroundDisabled = "#f4f5f2";
undefined;
```

**After:**
```js
export const colorStateBackgroundDisabled = "#f4f5f2";
```

### DIFF
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -674,9 +674,7 @@ const formats = {
           ...formatting,
         };
 
-        return comment && format.commentStyle !== none
-          ? addComment(to_ret, comment, format)
-          : to_ret;
+        return addComment(to_ret, comment, format);
       }),
     ]
       .flat()
@@ -793,9 +791,7 @@ const formats = {
           ...formatting,
         };
 
-        return comment && format.commentStyle !== none
-          ? addComment(to_ret, comment, format)
-          : to_ret;
+        return addComment(to_ret, comment, format);
       }),
     ]
       .flat()
--- a/lib/common/formats.js
+++ b/lib/common/formats.js
@@ -674,7 +674,9 @@ const formats = {
           ...formatting,
         };
 
-        return comment ? addComment(to_ret, comment, format) : to_ret;
+        return comment && format.commentStyle !== none
+          ? addComment(to_ret, comment, format)
+          : to_ret;
       }),
     ]
       .flat()
@@ -791,7 +793,9 @@ const formats = {
           ...formatting,
         };
 
-        return comment ? addComment(to_ret, comment, format) : to_ret;
+        return comment && format.commentStyle !== none
+          ? addComment(to_ret, comment, format)
+          : to_ret;
       }),
     ]
       .flat()
--- a/.changeset/shiny-grapes-join.md
+++ b/.changeset/shiny-grapes-join.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': patch
+---
+
+Disallow output literal "undefined" for tokens with descriptions when commentStyle is set to none.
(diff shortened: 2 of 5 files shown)


lib/common/formats.js
```diff
@@ -674,7 +674,9 @@ const formats = {
           ...formatting,
         };
 
-        return comment ? addComment(to_ret, comment, format) : to_ret;
+        return comment && format.commentStyle !== none
+          ? addComment(to_ret, comment, format)
+          : to_ret;
```
jorenbroekema: I think I would adjust the addComment function to just return the token as is, rather than having a conditional here. I think the original code is also not ideal, if comment is falsy then addComment can just take care of also just returning the token as is. This way, we can also put a test for the addComment function itself to handle nullish comments or commentStyle none. Integration tests with the formats are still welcome, but optional then.

jorenbroekema (changes requested):
Looks good for the most part but I think we can move some logic to the addComment function to handle nullish comments or comment style none.

Also, can you create a changeset by using `npx changeset` in the root of the project? Then select patch and add a patch note for this bugfix :)

lib/common/formats.js
lara-ferrer: @jorenbroekema fixed, thanks for pointing it out :)

jorenbroekema (approved):

Issue #1713, opened 2026-06-27
Unexpected alpha component rounding for rgba() values when using color/css transform


miqh:
Hello,

I ran into some unexpected behaviour with the `color/css` transform for some design token values I was given to work with. Specifically, `rgba()` values (e.g. `rgba(0, 0, 0, 0.0625)`) where the alpha component has more than two decimal places. The CSS file output appears to have rounded the alpha component to two decimal places.

The Style Dictionary configuration I'm using doesn't include anything remarkable either, just `transformGroup: "css"` and `format: "css/variables"` under the definition for `platforms.css`.

After a bit of digging, I noticed that the `color/css` transform is using `toRgbString()` from `tinycolor2`.

https://github.com/style-dictionary/style-dictionary/blob/9a9cca0413c51be65030067c611124194babdf54/lib/common/transforms.js#L861

And the implementation of `toRgbString()` appears to hard-code a two decimal place rounding on the alpha component.

https://github.com/bgrins/TinyColor/blob/a93228a6782bfe9a6154da14f5a7ece3ad9fb6c4/tinycolor.js#L85

https://github.com/bgrins/TinyColor/blob/a93228a6782bfe9a6154da14f5a7ece3ad9fb6c4/tinycolor.js#L140-L142

I've got a local workaround for now, but I thought I'd just raise an issue in case the intent of `color/css` applied to `rgba()` values should preserve whatever alpha component is given.

One possible fix I can see would be to instead use `toRgb()` to grab the non-alpha components, round those, then separately grab the alpha component and return it as-is.

jatintakhtani2612:
Hi @miqh, I'd like to work on this issue as my first contribution to Style Dictionary.

I understand the issue as a precision loss in the color/css transform when converting RGBA values with alpha components beyond two decimal places.

I’d like to investigate the current transform implementation, add a regression test for values such as rgba(0, 0, 0, 0.0625), and implement a minimal fix that preserves the alpha precision without changing the existing behavior for other color formats.

If no one is currently working on this, I'd be happy to take it up.

miqh:
@jatintakhtani2612, I don't know why you tagged me to enquire about working on the issue I raised here. I'm not a maintainer of this repository. It's not for me to decide whether the status quo is intended behaviour.

Even so, I literally pointed out what I believe is exacerbating the issue and also proposed a possible fix. Assume from the lack of attention that there's probably no intention in addressing this issue, for which I've already said I'm working around anyway.

[...]

PR #1740, opened 2026-09-03
fix: preserve CSS alpha precision

LeonMAG:
## Issue

Related to #1713.

## Description of changes

Thanks for the guidance, @jorenbroekema.

* Preserve the parsed alpha value in `color/css` instead of using tinycolor2's two-decimal string serializer.
* Retain hexadecimal output for opaque colors and the existing output for `0.6` alpha values.
* Add regression coverage for legacy CSS and DTCG sRGB values with alpha `0.0625`.
* Add a patch changeset.

This intentionally does not introduce a configurable precision policy or migrate to ColorJS.io. It preserves the numeric alpha value already parsed by the current transform; it does not preserve lexical spelling such as trailing zeroes.

## Verification

* `npm run lint`
* Node tests: 991 passing, 4 pending.
* Browser tests: 985 passing, 4 skipped.
* Performance tests: 6 passing.
* Node 22.6 strip-types test.

### DIFF
--- a/lib/common/transforms.js
+++ b/lib/common/transforms.js
@@ -855,11 +855,11 @@ export default {
     filter: isColor,
     transform: function (token, _, options) {
       const color = getColor(token, options);
-      if (color.getAlpha() === 1) {
+      const { r, g, b, a } = color.toRgb();
+      if (a === 1) {
         return color.toHexString();
-      } else {
-        return color.toRgbString();
       }
+      return `rgba(${r}, ${g}, ${b}, ${a})`;
     },
   },
 
--- a/.changeset/preserve-css-alpha-precision.md
+++ b/.changeset/preserve-css-alpha-precision.md
@@ -1,0 +1,5 @@
+---
+'style-dictionary': patch
+---
+
+Preserve alpha precision in the `color/css` transform.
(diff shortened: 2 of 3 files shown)


lib/common/transforms.js
```diff
@@ -855,11 +855,11 @@ export default {
     filter: isColor,
     transform: function (token, _, options) {
       const color = getColor(token, options);
-      if (color.getAlpha() === 1) {
+      const { r, g, b, a } = color.toRgb();
```
jorenbroekema, line 858: this change is good but we need to keep the feature that colors with an alpha of 1 get transformed to hex instead of rgb format.

I don't like that it works this way, but changing this would be a breaking change. I'd be surprised if we don't have a test that verifies the behavior of alpha === 1 -> hex output and that test breaks now. Then again, I also know our coverage isn't 100%

jorenbroekema (changes requested):

lib/common/transforms.js
LeonMAG: Thanks for flagging this. I checked the explicit `alpha === 1` case: both `main` and this branch transform `rgba(170, 170, 170, 1)` to `#aaaaaa`. The branch is still preserved because `toRgb().a` remains `1` and returns `toHexString()`.

You're right that the current opaque-color test uses `rgb(...)` rather than declaring alpha `1` explicitly. I've added a focused regression test for `rgba(..., 1) -> hex` so that compatibility contract is unambiguous.

jorenbroekema (approved):

